PR #29, opened 2020-11-01
Configure Sonar report in PR

akphi:


akphi:
@gs-bracej is there a way to separate the deploy step from the build step?

PR #41, opened 2021-02-06
chore: force resolutions to fix CVEs

akphi:


MauricioUyaguari (approved):

akphi:
@maoo do you know why my `whitesource.config` does not seem to have any effect here? I used `yarn` resolutions to force using version suggested by Whitesource, but looks like my `yarn.config` is not getting picked up even when I already specified `npm.yarnProject` according to Whitesource Unified Agent [doc](https://whitesource.atlassian.net/wiki/spaces/WD/pages/1544880156/Unified+Agent+Configuration+Parameters).

I also took an extra step to disable `npm.runPreStep` and add `package-lock.json` just in case to see if I can speed up Whitesource check but it still takes a very long time (> ~13mins). Is there any way we can check the log for whitesource here to see what is going on? Thanks @maoo!

akphi:
@pierredebelen looks like Whitesource doc mentions that changes to whitesource config can only take effect when it's on the `default branch` (a.k.a. `master`). I will go ahead and merge these, anticipating a revert.

akphi:
@maoo Ah, looks like it's really like I said, we need to merge the changes to the default branch (a.k.a `master`) for it to take effect. So I don't think I need help here anymore 😭 

@pierredebelen Please go ahead and create docker image `0.1.1`. We should be good now. CVEs report for Whitesource are now proper and should come only from dependencies we actually use. Now, since the scanner is instructed to use `yarn.lock` manifest file, it does not need to run `npm install` anymore. This means the check time is reduced to ~1 min, and Whitesource no longer gives false positive from random artifacts in `./node_modules` - by product of `npm install`.

PR #50, opened 2021-03-03
fix(service test state): minify url creation for JSON model connection

aziemchawdhary-gs:
URL creation for services with JSON model connections should minify the JSON data so as not to break string size limits in Legend Java generation


akphi (approved):
Thanks!

PR #58, opened 2021-03-04
Small fixes

MauricioUyaguari:
Fixes config to start dev server

### DIFF
--- a/packages/legend-studio-app/scripts/setup.js
+++ b/packages/legend-studio-app/scripts/setup.js
@@ -43,7 +43,7 @@ fs.writeFileSync(
       appName: 'studio',
       env: 'local',
       sdlc: {
-        url: 'https://legend.finos.org/sdlc',
+        url: 'https://legend.finos.org/sdlc/api',
       },
       engine: {
         url: 'https://legend.finos.org/exec',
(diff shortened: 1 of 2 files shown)


packages/legend-studio-app/scripts/setup.js
```diff
@@ -43,7 +43,7 @@ fs.writeFileSync(
       appName: 'studio',
       env: 'local',
       sdlc: {
-        url: 'https://legend.finos.org/sdlc',
+        url: 'https://legend.finos.org/sdlc/api',
       },
       engine: {
         url: 'https://legend.finos.org/exec',
```
akphi, line 49: if that is the case for SDLC, why don't we have to do the same for EXEC?

MauricioUyaguari: always been the case :) we should change our clients to handle the /api in building the url. Right now all calls to legend sdlc required the /api prefix.

akphi: ok, let's not change the config setup here, let's change the client instead. Looks super weird from config perspective.

akphi:
Closing in favor of https://github.com/finos/legend-studio/pull/59

PR #134, opened 2021-03-24
fix: ensure ObjectInputData stores unformatted test data

aziemchawdhary-gs:
Fixes issue where test input data is not necessarily stored in unformatted state: flow is when a user creates a test for M2M, and the test data is filled by createMockDataForClass from MockDataUtils. This function returns formatted JSON which is then stored in the ObjectInputData class. 

If the user does not modify the test data, then the callback to unformat the test data is not called and hence the data is kept with formatting and breaks when switching to grammar mode. 

This change ensures that data stored in ObjectInputData is stored without formatting.

### DIFF
--- a/.changeset/fresh-shirts-tie.md
+++ b/.changeset/fresh-shirts-tie.md
@@ -1,0 +1,5 @@
+---
+'@finos/legend-studio': patch
+---
+
+Ensure that test data is stored with no formatting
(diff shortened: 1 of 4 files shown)


akphi:
@aziemchawdhary-gs @MauricioUyaguari this bug https://github.com/finos/legend-studio/issues/135 appears recently (unfortunately, we don't have a regression test to cover this) but change-detection is thrown off and it might have something to do with this

MauricioUyaguari (approved):

PR #153, opened 2021-04-12
fix: correct studio change detection for self-join

hardikmaheshwari:


packages/legend-studio/src/models/protocols/pure/v1/transformation/pureGraph/from/V1_DatabaseTransformer.ts
```diff
@@ -163,7 +166,9 @@ export const V1_transformTableAliasToTablePointer = (
 ): V1_TablePtr => {
   const tablePtr = new V1_TablePtr();
   tablePtr.database = tableAlias.relation.ownerReference.valueForSerialization;
-  tablePtr.schema = tableAlias.relation.value.schema.name;
+  tablePtr.schema = tableAlias.isSelfJoinTarget
+    ? SELF_JOIN_SCHEMA_NAME
+    : tableAlias.relation.value.schema.name;
```
MauricioUyaguari, line 171: what if you have a self join such as 
`schemaA.tableA.colA = {target}.colB` 
wouldn't this change the self join schema to default pointing to an incorrect schema?
This fix may not be as simple and may require the use of inferable.

MauricioUyaguari (approved):

PR #158, opened 2021-04-16
Fix setting source for PurePropertyMapping

emilia-sokol-gs:

### DIFF
--- a/packages/legend-studio/src/models/protocols/pure/v1/transformation/pureGraph/to/V1_ProtocolToMetaModelPropertyMappingVisitor.ts
+++ b/packages/legend-studio/src/models/protocols/pure/v1/transformation/pureGraph/to/V1_ProtocolToMetaModelPropertyMappingVisitor.ts
@@ -20,6 +20,7 @@ import {
   assertNonEmptyString,
   guaranteeType,
   guaranteeNonNullable,
+  returnUndefOnError,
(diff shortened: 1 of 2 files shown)


packages/legend-studio/src/models/protocols/pure/v1/transformation/pureGraph/to/V1_ProtocolToMetaModelPropertyMappingVisitor.ts
```diff
@@ -187,9 +187,16 @@ export class V1_ProtocolToMetaModelPropertyMappingVisitor
...
+    let sourceSetImplementation: SetImplementation = topParent;
+    if (protocol.source) {
+      try {
+        sourceSetImplementation = topParent.parent.getClassMapping(
+          protocol.source,
+        );
+      } catch (e: unknown) {
+        // ignore
+      }
+    }
```
MauricioUyaguari, line 199: can we use `returnUndefOnError`, which does the try catch and returns undefined on error.
so if it returns undefined we use top parent

emilia-sokol-gs: good point!

MauricioUyaguari (approved):

PR #173, opened 2021-04-26
add lambda resolver

MauricioUyaguari:

This adds a RawLambdaResolver as well as small fixes to the schema and model definition of V1_AppliedProperty. The main points of the PR are 

- [X] A RawLambda resolver is added to resolve element paths inside lambdas that leverage imports. The resolver was added since the SDLC server does not currently support the use of sections (element containing imports). Currently, we resolve element paths everywhere in the app except in the lambdas since lambdas are kept in the format they are saved in the SDLC server. This resolver allows the graph to be compiled without the need to save the section elements.
- [X]  The cardinality of V1_AppliedProperty was fixed to not require a class and match the protocol in pure.
- [X] The ordering of the schema of V1_AppliedProperty was fixed to be alphabetically ordered.

### DIFF
--- a/.changeset/good-years-bake.md
+++ b/.changeset/good-years-bake.md
@@ -1,0 +1,5 @@
+---
+'@finos/legend-studio': minor
+---
+
+A RawLambda resolver is added to resolve element paths inside lambdas that leverage imports. This allow successful compilation of graph that contain lambdas using imports.
--- a/packages/legend-studio/src/models/protocols/pure/v1/model/valueSpecification/application/V1_AppliedProperty.ts
+++ b/packages/legend-studio/src/models/protocols/pure/v1/model/valueSpecification/application/V1_AppliedProperty.ts
@@ -18,7 +18,7 @@ import type { V1_ValueSpecificationVisitor } from '../../../model/valueSpecifica
 import { V1_ValueSpecification } from '../../../model/valueSpecification/V1_ValueSpecification';
 
 export class V1_AppliedProperty extends V1_ValueSpecification {
-  class!: string;
+  class?: string;
   property!: string;
   parameters: V1_ValueSpecification[] = [];
 
--- a/packages/legend-studio/src/models/protocols/pure/v1/transformation/pureGraph/to/helpers/V1_RawLambdaResolver.ts
+++ b/packages/legend-studio/src/models/protocols/pure/v1/transformation/pureGraph/to/helpers/V1_RawLambdaResolver.ts
@@ -1,0 +1,384 @@
+/**
+ * Copyright (c) 2020-present, Goldman Sachs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {
+  assertErrorThrown,
+  returnUndefOnError,
+} from '@finos/legend-studio-shared';
+import { CORE_LOG_EVENT } from '../../../../../../../../utils/Logger';
+import type { PackageableElement } from '../../../../../../../metamodels/pure/model/packageableElements/PackageableElement';
+import type { PackageableElementImplicitReference } from '../../../../../../../metamodels/pure/model/packageableElements/PackageableElementReference';
+import { RawLambda } from '../../../../../../../metamodels/pure/model/rawValueSpecification/RawLambda';
+import { isValidFullPath } from '../../../../../../../MetaModelUtility';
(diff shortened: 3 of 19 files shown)


.changeset/good-years-bake.md
```diff
@@ -0,0 +1,5 @@
+---
+'@finos/legend-studio': patch
+---
+
+A RawLambda resolver is added to resolve element paths inside lambdas that leverage imports. The resolver was added since the SDLC server does not currently support the use of sections (element containing imports). Currently, we resolve element paths everywhere in the app except in the lambdas since lambdas are kept in the format they are saved in the SDLC server. This resolver allows the graph to be compiled without the need to save the section elements.
```
akphi, line 5: You're actually going the opposite direction this time 😭 But I love the details. I think we keep this as the PR description and simplify this down. This is a changelog so people want faster read.

packages/legend-studio/src/models/protocols/pure/v1/model/valueSpecification/application/V1_AppliedProperty.ts
```diff
@@ -18,7 +18,7 @@ import type { V1_ValueSpecificationVisitor } from '../../../model/valueSpecifica
 import { V1_ValueSpecification } from '../../../model/valueSpecification/V1_ValueSpecification';
 
 export class V1_AppliedProperty extends V1_ValueSpecification {
-  class!: string;
+  class?: string;
```
akphi, line 21: Why? Could you at least mention in the PR summary why we do this? Maybe break the PR summary down into multiple points and explain what you do to keep a record.

.changeset/good-years-bake.md
```diff
@@ -0,0 +1,5 @@
+---
+'@finos/legend-studio': patch
```
akphi, line 2: Also, I think we can safely put this one as a `minor` version bump. This is not a trivial patch right?

packages/legend-studio/src/models/protocols/pure/v1/transformation/pureProtocol/serializationHelpers/V1_ValueSpecificationSerializer.ts
```diff
@@ -207,13 +207,13 @@ const V1_appliedFunctionModelSchema = createModelSchema(V1_AppliedFunction, {
...
   _type: usingConstantValueSchema(V1_ValueSpecificationType.APPLIED_PROPERTY),
   class: optional(primitive()),
-  property: primitive(),
   parameters: list(
     custom(
       (val) => V1_serializeValueSpecification(val),
       (val) => V1_deserializeValueSpecification(val),
     ),
   ),
+  property: primitive(),
```
akphi, line 216: Let's separate this into another commit (also because changes like this often means we don't have test covering it). We will rebase this PR and not squash it. If this proves to be too painful, we can pass. But let's do that next time!

packages/legend-studio/src/stores/ApplicationConfig.ts
```diff
@@ -98,7 +98,13 @@ class ApplicationCoreOptions {
    * to figure out how we want to store this element in SDLC.
    */
   EXPERIMENTAL__enableFullGrammarImportSupport = false;
-
+  /**
+   * Allows disabling of resolving element paths inside a RawLambda
+   *
+   * NOTE: when we move to save imports as part of the user's project, this feature
+   * will no longer be needed and can be removed
+   */
```
akphi, line 106: Maybe add one more line here saying we should use it with the section index flag, as they are in the same group. Though I agree, for unforeseeable reasons, it makes sense to split them like this.

packages/legend-studio/src/stores/__tests__/buildGraph/RawLambaResolver.test.ts
```diff
@@ -0,0 +1,114 @@
+/**
```
akphi, line 1: Love the minimal test case 🎉

akphi (changes requested):

packages/legend-studio/src/models/protocols/pure/v1/transformation/pureGraph/to/helpers/V1_RawLambdaResolver.ts
```diff
@@ -0,0 +1,385 @@
...
+ * Method resolves element paths inside V1_RawLambda
+ * To do this we take V1_RawLamba (body: object, parameters: obect) convert it to V1_Lambda and
+ * resolve where possible (where an element path is defined as part of the value specification flow.
+ * When completed we convert it back to V1_RawLambda.
+ * This method ignores any function matching therefore if any time we encounter an error resolving an element path
+ * inside the value specification, we continue on the graph unchanging the value spec.
+ * We use this new resolved lamba if we have changed any element path (noted by the hasModifiedLambda flag)
+ * The whole flow is wrapper around a try/catch and we use the orginal lambda if any errors arise from this flow.
+ */
+
```
akphi, line 345: Thanks for this! Though I think when you make comment like this, don't leave an empty line like this as IDE will fail to make it the comment of the function. It would be nice when you can hover on the function and see its doc.

.changeset/good-years-bake.md
MauricioUyaguari: yeah agree this should be a minor increase

akphi (approved):

Issue #165, opened 2021-04-22
RFC: Serialization rework


akphi:
## 💬 Request for Comments

Serialization right now leaks out some inference we do during graph building which makes it not 1-1 with the grammar and the input JSON. We need to `dumbify` the serializer, this will help with change detection as especially if in the future SDLC starts supporting entities store as grammar text.

### Context and Motivation 

The basic flow of the app data is like this: `protocol JSON` -- [deserialize]--> `protocol` -- [graph build] --> `metamodel` -- [transform] --> `protocol` -- [serialize] --> `protocol JSON`

In `[graph build]`, we have been trying to do some level of inference to fill the gaps in the protocol JSON. This is needed since sometimes if the user input is grammar text, the parser itself cannot derive enough contextual information to fill the protocol JSON, note that this is also the [intended behaviour of the parser to keep the grammar and the protocol 1-1](https://github.com/finos/legend-engine/tree/master/legend-engine-language-pure-grammar#philosophy). The problem with this is during serialization, we don't want to `leak` these graph-building inference to the protocol JSON. Otherwise, we will see problems with change detection, especially when we start to get entities from SDLC which are stored as grammar text.

### Implementation Plan

Since there is no easy way to keep track of all the issues we might face, the best way to manage this is to file separate issues for each problem found and refer them back to this mega-thread for better context. Each of the issues will need to include a roundtrip test case which:
- Takes grammar text as input
- Converts the grammar text into JSON by calling engine backend. Store this JSON as `A`
- Build the graph with `A` and serialize the graph to obtain JSON `B`
- Make sure `A` matches `B`
- Make sure we can compute hash using `A` because when `SDLC` server starts storing the graph models in grammar text, when they send us back the `entities`, those will be used for hash computation

### Test runner
- [x] Automate the test in the build pipeline: use `maven` to build and run the latest version of `finos/legend-engine` (https://github.com/finos/legend-studio/pull/293)

### Sample Issue tracker
- [x] Basic M2M (#188) 
- [x] Embedded relational mapping (#189)

PR #188, opened 2021-05-04
fix: make sure grammar roundtrip work for basic M2M test case

akphi:
Related to #165

The one problem remains with this is the fact that grammar parser adds a parameter for `src` to the property mapping lambda, which Studio excludes intentionally, @MauricioUyaguari check if we can remove this addition from the grammar parser side and so we should be good with this case


packages/legend-studio-manual-tests/src/__tests__/roundtrip-grammar/RoundtripGrammar.test.ts
```diff
@@ -15,6 +15,32 @@
  */
 
 /// <reference types="jest-extended" />
+
+// NOTE: mock these methods to make sure we rule out false positive. The grammar parser for any List type field,
```
MauricioUyaguari, line 19: perfect

packages/legend-studio/src/models/protocols/pure/v1/transformation/pureGraph/from/V1_ConnectionTransformer.ts
```diff
@@ -239,7 +240,8 @@ const transformJsonModelConnection = (
 ): V1_JsonModelConnection => {
   const connection = new V1_JsonModelConnection();
   connection.class = V1_transformElementReference(element.class);
-  connection.store = V1_transformElementReference(element.store);
+  // to make this consistent with grammar parser, ModelStore is a pseudo-store and will be skipped
+  connection.store = '';
```
MauricioUyaguari, line 244: would making store optional be problematic here?

MauricioUyaguari: Can we leave this as undefined?

packages/legend-studio-manual-tests/src/__tests__/roundtrip-grammar/RoundtripGrammar.test.ts
akphi: 🙇‍♂️

MauricioUyaguari (approved):

PR #189, opened 2021-05-05
Resolve some issues around grammar roundtrip for embedded relational mapping

akphi:
See https://github.com/finos/legend-studio/issues/165

> NOTE: this PR attempts to resolve a subset of problem regarding the grammar roundtrip for embedded relational mapping. There are other issues like `mainTable` and `primaryKey` which we will need to address in another fix. Those require some logic (or more like reverse logic to graph building) in the transformer.

### DIFF
--- a/packages/legend-studio-manual-tests/src/__tests__/roundtrip-grammar/RoundtripGrammar.test.ts
+++ b/packages/legend-studio-manual-tests/src/__tests__/roundtrip-grammar/RoundtripGrammar.test.ts
@@ -52,7 +52,8 @@ const ENGINE_SERVER_URL = 'http://localhost:6060/api';
 const TEST_CASE_DIR = path.resolve(__dirname, 'cases');
 const EXCLUDED_CASE_FILES: string[] = [
   'basic-M2M.pure', // until we fix the grammar parser to not include the variable for the property mapping lambda
-  'embedded-relational-mapping-with-imports.pure', // TODO?
+  // 'embedded-relational-mapping.pure',
+  'nested-embedded-relational-mapping.pure',
 ];
 
 const checkGrammarRoundtrip = async (
--- a/packages/legend-studio/src/models/protocols/pure/v1/transformation/pureGraph/from/V1_ConnectionTransformer.ts
+++ b/packages/legend-studio/src/models/protocols/pure/v1/transformation/pureGraph/from/V1_ConnectionTransformer.ts
@@ -229,8 +229,7 @@ const transformModelChainConnection = (
   element: ModelChainConnection,
 ): V1_ModelChainConnection => {
   const connection = new V1_ModelChainConnection();
-  // to make this consistent with grammar parser, ModelStore is a pseudo-store and will be skipped
-  connection.store = undefined;
+  connection.store = undefined; // @MARKER: GRAMMAR ROUNDTRIP --- omit this information during protocol transformation as it can be interpreted while building the graph
   connection.mappings = element.mappings.map(V1_transformElementReference);
   return connection;
 };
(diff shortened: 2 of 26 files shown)


packages/legend-studio-manual-tests/src/__tests__/roundtrip-grammar/RoundtripGrammar.test.ts
```diff
@@ -52,7 +52,8 @@ const ENGINE_SERVER_URL = 'http://localhost:6060/api';
 const TEST_CASE_DIR = path.resolve(__dirname, 'cases');
 const EXCLUDED_CASE_FILES: string[] = [
   'basic-M2M.pure', // until we fix the grammar parser to not include the variable for the property mapping lambda
-  'embedded-relational-mapping-with-imports.pure', // TODO?
+  // 'embedded-relational-mapping.pure',
+  'nested-embedded-relational-mapping.pure',
```
MauricioUyaguari, line 56: so these are now passing. nested-embedded and basic M2M?

packages/legend-studio/src/models/protocols/pure/v1/transformation/pureGraph/from/V1_ConnectionTransformer.ts
```diff
@@ -229,8 +229,7 @@ const transformModelChainConnection = (
   element: ModelChainConnection,
 ): V1_ModelChainConnection => {
   const connection = new V1_ModelChainConnection();
-  // to make this consistent with grammar parser, ModelStore is a pseudo-store and will be skipped
-  connection.store = undefined;
+  connection.store = undefined; // @MARKER: GRAMMAR ROUNDTRIP --- omit this information during protocol transformation as it can be interpreted while building the graph
```
MauricioUyaguari, line 232: Did we remove the ModelStore check in our own compiler

packages/legend-studio/src/models/protocols/pure/v1/transformation/pureGraph/to/V1_ProtocolToMetaModelPropertyMappingVisitor.ts
```diff
@@ -442,7 +448,17 @@ export class V1_ProtocolToMetaModelPropertyMappingVisitor
     );
     const relationalPropertyMapping = new RelationalPropertyMapping(
       this.topParent ?? this.immediateParent,
-      PropertyExplicitReference.create(property),
+      propertyOwner instanceof Class // TODO: we also probably need to handle this for association mapping
```
MauricioUyaguari, line 451: Confused as to why the owner determines impl vs expli property type

MauricioUyaguari: Doesn't it make sense to check if the immediate parent is RootRelational -> expl if not use impl?

packages/legend-studio/src/models/protocols/pure/v1/transformation/pureGraph/from/V1_ConnectionTransformer.ts
akphi: no, this is just for transformer, the compiler check is still there

packages/legend-studio-manual-tests/src/__tests__/roundtrip-grammar/RoundtripGrammar.test.ts
akphi: good point, it's working now, I will remove this exclusion in next PR

MauricioUyaguari (approved):

PR #214, opened 2021-05-25
Rework mapping editor execution panel

akphi:
Rework the mapping editor execution panel to accommodate relational mapping (i.e. we need to make its behaviour more consistent across different types of mapping). This includes:
- [x] The `target` panel is changed from supporting `graph fetch` to supporting generic query and should be edited using the query builder
- [x] The `source` panel is changed so that test data can be generated manually
- [x] Instead of choosing the source/target, now the user will choose the class mapping to generate content for these 2 panels. This might be a little bit less conceptually friendly, but will prove to be more conceptually accurate
- [x] Remove the flag `TEMPORARY__enableGraphFetch` that disables graph fetch support in query builder: this is needed so we can edit M2M query right away when we boot up query builder


packages/legend-studio/src/stores/editor-state/element-editor-state/mapping/MappingExecutionState.ts
```diff
@@ -217,77 +254,114 @@ export class MappingExecutionJsonModelConnectionRuntimeState extends MappingExec
...
+  }
+
+  get isValid(): boolean {
+    return true;
+  }
+
+  get runtime(): Runtime {
+    const datasourceSpecification = new LocalH2DatasourceSpecification();
+    datasourceSpecification.setTestDataSetupSqls(
+      // TODO: we need to properly review this gross simplification of formatting SQL strings
```
MauricioUyaguari, line 347: gross simplification  :) love it

packages/legend-studio-plugin-query-builder/src/components/QueryBuilderFetchStructurePanel.tsx
```diff
@@ -78,27 +78,25 @@ export const QueryBuilderFetchStructurePanel = observer(
             <div className="panel__header__title__label">fetch structure</div>
           </div>
           <div className="panel__header__actions">
-            {queryBuilderState.TEMPORARY__enableGraphFetch && (
```
MauricioUyaguari, line 81: One thing about removing this. are we losing any features by switching to the graph fetch in query builder? No right as both graph fetch trees should be the same. wouldnt want to lose key features like "select all mapped properties".

packages/legend-studio-plugin-query-builder/src/QueryBuilderPlugin.tsx
```diff
@@ -139,6 +130,21 @@ export class QueryBuilderPlugin extends EditorPlugin {
     ];
   }
 
+  getExtraMappingExecutionQueryEditorRendererConfigurations(): MappingExecutionQueryEditorRendererConfiguration[] {
```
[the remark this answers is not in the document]
akphi: because this is for `Service` plugin, we will move it out of this later

MauricioUyaguari (approved):

Issue #249, opened 2021-06-11
RFC: (mega-thread) Execution plan viewer


MauricioUyaguari:
### Motivation

The execution plan holds all the information required for the engine server (legend-engine) to execute a query. With this viewer, we plan to shed some light on the context of the execution plan and showcase to the user the steps taken to execute their queries to provide the data in the shape they want it.

Since the execution plan holds various types of nodes for different types of queries and data sources, it is not feasible to inaugurate a user interface covering all use cases. For this reason, this feature will be implemented in steps. We will cover basic relational, model-to-model cases in the beginning and lead up to more advanced use cases. For the nodes not supported we will default to a basic JSON string viewer.

### Implementation

As mentioned above we will tackle the viewer through phases. When creating the viewer for each phase, the steps will be similar to adding support for a new packageable element except there is no need for transformation as the execution plan viewer is a **readonly** UI for now. Therefore to add support please follow the below steps. 

- [x] Add `V1_ExecutionPlan` models (#248)
- [x] Add V1 serializer/de-serializer support to convert from/to raw JSON and `V1_ExecutionPlan` Models (#248)
- [x] Add metamodel `ExecutionPlan` models (#248)
- [x] Add Execution Plan builder converting `V1_ExecutionPlan` protocol to `ExecutionPlan` metamodel (and potentially vice versa so we can do more testing) (#248)
- [x] Use the metamodel `ExecutionPlan` to create a viewer. Using the metamodel execution plan will provide you with the compiled metamodel graph to be able to quickly find referenced elements in the graph such as classes, mappings, and runtimes. 
- [x] MVP execution plan viewer interface (See https://github.com/finos/legend-studio/issues/289)
- [ ] In plan viewer, classify the nodes in different buckets, and render them differently:
  - [ ] Flow control (sequence/conditional/loop/...)
  - [ ] Templating node (free-marker)

#### Execution Plan
See Attached Execution Plan Json
[ExecutionPlan_Phase_1.txt](https://github.com/finos/legend-studio/files/6635388/message.2.txt)

#### Pure Grammar
 ```
###Relational
Database model::Test
(
  Table FirmTable
  (
    id INTEGER PRIMARY KEY,
    Legal_name VARCHAR(200)
  )
  Table PersonTable
  (
    id INTEGER PRIMARY KEY,
    firm_id INTEGER,
    firstName VARCHAR(200),
    lastName VARCHAR(200)
  )

[...]

PR #252, opened 2021-06-14
Support execution plan processing and more query builder stabilization

akphi:
This PR supports roundtrip processing of execution plan (facilitates https://github.com/finos/legend-studio/issues/249) and miscellaneous fixes around stabilization of relational flow.


packages/legend-studio-plugin-query-builder/src/stores/QueryResultSetModifierState.ts
```diff
@@ -179,7 +179,10 @@ export class QueryResultSetModifierState {
       val.parametersValues[1] = collection;
       currentExpression = val;
     }
-    if (this.limit || options?.overridingLimit) {
+    if (
+      this.canModifyLambdaInProjectionMode(lambda) &&
+      (this.limit || options?.overridingLimit)
+    ) {
```
MauricioUyaguari, line 185: Is the best approach. 
What is the cost to do
graphFetch()->take()->serialize()
On large data sets, this would become an issue if we remove the take altogether

akphi: `take()` simply does not work with `graphFetch` as I understood. The cost, therefore is tremendous - i.e. 💥

packages/legend-studio/src/models/protocols/pure/PureProtocolProcessorPlugin.ts
```diff
@@ -23,11 +23,17 @@ import type { V1_PureModelContextData } from './v1/model/context/V1_PureModelCon
...
 import type { Runtime } from '../../metamodels/pure/model/packageableElements/runtime/Runtime';
+import type { V1_GraphTransformerContext } from './v1/transformation/pureGraph/from/V1_GraphTransformerContext';
 
 export type V1_ElementProtocolClassifierPathGetter = (
   protocol: V1_PackageableElement,
 ) => string | undefined;
 
+export type V1_ElementTransformer = (
+  metamodel: PackageableElement,
+  context: V1_GraphTransformerContext,
```
MauricioUyaguari, line 34: Might be that I'm missing something, but when do you actually use the context in the transformer?

MauricioUyaguari (approved):

packages/legend-studio/src/models/protocols/pure/PureProtocolProcessorPlugin.ts
akphi: This is more to make sure plugins/presets can access the `context` as well.

PR #286, opened 2021-06-24
include mapping/runtime in executeInput

MauricioUyaguari:
Include mapping and runtime in execute input. This is to support executing queries with connection that references other mappings such as ModelChainConnections. 

As part of the commit, I removed the light approach of walking the graph and choosing the relevant elements of the graph as this would get more complicated adding these type of connections with not much gain as we expect to implement this relevant walker in the future taking into account these type of connections as well as the elements in the query lambda. 

### DIFF
--- a/.changeset/kind-gifts-obey.md
+++ b/.changeset/kind-gifts-obey.md
@@ -1,0 +1,5 @@
+---
+'@finos/legend-studio': patch
+---
+
+Include all runtime and mappings in executeInput to support connections referencing other mappings such as ModelChainConnections.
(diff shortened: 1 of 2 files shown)


.changeset/kind-gifts-obey.md
```diff
@@ -0,0 +1,5 @@
+---
+'@finos/legend-studio': patch
+---
+
+Include all runtime and mappings in executeInput to support connections referencing other mappings such as ModelChainConnections.
```
akphi, line 5:
```suggestion
Include all runtimes and mappings in execution input to support connections referencing other mappings such as ModelChainConnections.
```

akphi (approved):

Issue #285, opened 2021-06-24
Bug: Embedded relational mapping get erased when opening the mapping editor


MauricioUyaguari:
## 🐛 Bug Report
If you have an existing mapping with an embedded relational class mapping and open the mapping editor to the specified class mapping, the embedded class mapping gets deleted. 
### Step to Reproduce
```
###Relational
Database mapping::db
(
  Table employeeFirmDenormTable
  (
    id INTEGER PRIMARY KEY,
    name VARCHAR(200),
    firmId INTEGER,
    legalName VARCHAR(200),
    address1 VARCHAR(200),
    postcode VARCHAR(10)
  )
  Table FirmInfoTable
  (
    id INTEGER PRIMARY KEY,
    name VARCHAR(200),
    other VARCHAR(200)
  )

  Join PersonFirmJoin(employeeFirmDenormTable.firmId = FirmInfoTable.id)
)


###Pure
Class other::Person
{
  name: String[1];
  firm: other::Firm[1];
}

Class other::Firm
{
  legalName: String[1];
  otherInformation: String[1];
}


###Mapping
Mapping mappingPackage::myMapping
(
  other::Firm[firm1]: Relational
  {
    ~primaryKey
    (
      [mapping::db]FirmInfoTable.id
    )
    ~mainTable [mapping::db]FirmInfoTable
    legalName: [mapping::db]FirmInfoTable.name,
    otherInformation: [mapping::db]FirmInfoTable.other
  }
  other::Person: Relational
  {
    ~primaryKey
    (
      [mapping::db]employeeFirmDenormTable.id
    )
    ~mainTable [mapping::db]employeeFirmDenormTable
    name: [mapping::db]employeeFirmDenormTable.name,
    firm
    (
      legalName: [mapping::db]employeeFirmDenormTable.legalName
    )
  }
)
```
- Copy the above grammar into text mode and go to form mode
- Go to the mapping and click on class mapping person
- The embedded class mapping `firm` will be deleted
- Go to text mode to confirm the embedded class mapping was deleted. 
### Expected behavior
Embedded class mappings are not supported in form mode but they should be preserved. We should disallow editing of this property mapping if its an embedded class mapping and save whats inside.

akphi:
My bad, I think we still missed this case

```
###Relational
Database mapping::db
(
  Table employeeFirmDenormTable
  (
    id INTEGER PRIMARY KEY,
    name VARCHAR(200),
    firmId INTEGER,
[...]

PR #311, opened 2021-06-30
Cleanup codebase, error messages, and expression processing logic

akphi:
Fixes https://github.com/finos/legend-studio/issues/285

Besides, I did some cleanups for error messages throughout the app to make them more consistent. The template for error messages will be something like `Can't ...: <extra reason>` or `Can't ... . Error: <error message>`. I also, go through the codebase and remove usage of naked `Error` and replace them with assertions to make code look cleaner and more readable. Also, this is a pattern we would definitely want to think of adopting in the future if we want to throw some contextual error while building the graph to handle them separately downstream.

The most important change though is to formalize the processing of function expression. Before, we will attempt to build all expressions optimistically, which probably only works for super simple expressions; only functions that we explicitly support such as `project`, `filter`, `getAll` will get special treatment. This special treatment is somewhat similar to how `engine` does type-inferencing and function-matching. Now, this special-treatment logic will be moved into `query builder`. The rationale is that core Studio should not try to handle expression, even the simplest type, this is the sole job of query builder - **Studio does not do function-matching/type-inferencing, this is the job of engine**.

Due to this, I modularize the builder for function expression to allow taking specialised processing logics from `query builder` and because of this, I promoted `query builder` to become a preset instead of a plugin like before. This is a **breaking change**.

A small bonus is that now `query builder` will be ok if users' input include full-path functions (e.g. `meta::pure::tds::groupBy`)


packages/legend-studio/src/models/metamodels/pure/model/packageableElements/domain/Package.ts
```diff
@@ -99,15 +99,14 @@ export class Package extends PackageableElement implements Hashable {
...
-        // create the node if it is not in parent package
-        node = Package.createPackageFromParent(str, parent);
-        parent.addChild(node);
-      } else {
+      if (!insert) {
         throw new GraphError(
           `Can't find packageable element '${str}' in package '${packageName}'`,
         );
       }
+      // create the node if it is not in parent package
```
MauricioUyaguari, line 107: any specific reason to switch the oder?

akphi: more idiomatic coding style - assertion

packages/legend-studio-manual-tests/src/__tests__/roundtrip-grammar/RoundtripGrammar.test.ts
```diff
@@ -128,6 +128,23 @@ const checkGrammarRoundtrip = async (
         change.oldPath !== '__internal__::SectionIndex',
     ).length,
   ).toBe(0);
+
+  // compose grammar and compare that with original grammar text
+  // NOTE: this is optional test as `grammar text <-> protocol` test should be covered
+  // in engine already.
+  // Here, we do it just so we might be able to detect problem in the grammar roundtrip in engine
+  const transformJsonToGrammarResult = await axios.post(
```
MauricioUyaguari, line 136: This is under the assumption the grammar is formatted in the correct order right?

packages/legend-studio-manual-tests/src/__tests__/roundtrip-grammar/cases/basic-M2M.pure
```diff
@@ -39,7 +39,7 @@ Mapping model::mapping
   *model::target::_Firm: Pure
   {
     ~src model::Firm
-    employees[model_target__Person, x]: $src.employees,
+    employees[model_target__Person]: $src.employees,
```
MauricioUyaguari, line 42: why lose the target id in the test.

packages/legend-studio-preset-query-builder/src/models/protocols/pure/v1/V1_QueryBuilder_FunctionExpressionBuilder.ts
```diff
@@ -0,0 +1,427 @@
...
+export const V1_buildProjectFunctionExpression = (
+  functionName: string,
+  parameters: V1_ValueSpecification[],
+  openVariables: string[],
+  compileContext: V1_GraphBuilderContext,
+  processingContext: V1_ProcessingContext,
+): [SimpleFunctionExpression, ValueSpecification[]] => {
+  assertTrue(
+    parameters.length === 2 || parameters.length === 3,
+    `Can't build project() expression: project() expects at 1-2 arguments`,
```
MauricioUyaguari, line 251: 2-3rguments?

packages/legend-studio-manual-tests/src/__tests__/roundtrip-grammar/RoundtripGrammar.test.ts
akphi: yeah, a bit strict, but I thought we might as well to cover more cases

packages/legend-studio-manual-tests/src/__tests__/roundtrip-grammar/cases/basic-M2M.pure
akphi: it will fail when we do rountrip

packages/legend-studio-preset-query-builder/src/models/protocols/pure/v1/V1_QueryBuilder_FunctionExpressionBuilder.ts
akphi: yes, there are 2 forms:

```
project([col(x|$x.name, 'name'])
project([x|$x.name], ['name'])
```

packages/legend-studio-manual-tests/src/__tests__/roundtrip-grammar/cases/basic-M2M.pure
MauricioUyaguari: bit confused why is the x swallowed?

packages/legend-studio-preset-query-builder/src/models/protocols/pure/v1/V1_QueryBuilder_FunctionExpressionBuilder.ts
MauricioUyaguari: yeah i meant your error message says 1-2 arguments, but should be 2-3 no?

packages/legend-studio-manual-tests/src/__tests__/roundtrip-grammar/cases/basic-M2M.pure
akphi: 🤷

packages/legend-studio-preset-query-builder/src/models/protocols/pure/v1/V1_QueryBuilder_FunctionExpressionBuilder.ts
akphi: oh, because the first param is the preceding expression.

packages/legend-studio-manual-tests/src/__tests__/roundtrip-grammar/cases/basic-M2M.pure
MauricioUyaguari: can you explain :)

MauricioUyaguari (approved):

packages/legend-studio-manual-tests/src/__tests__/roundtrip-grammar/cases/basic-M2M.pure
akphi: just as confused as you are, but pretty sure we need to check how `engine` swallow `x`, and yeah, why do we have `x` there in the first place?

PR #313, opened 2021-07-02
feat: add button for `quotedIdentifiersIgnoreCase` flag

hardikmaheshwari:
Add button (UI support) for quotedIdentifiersIgnoreCase flag property of snowflake datasource specification.
Details about quotedIdentifiersIgnoreCase flag - https://docs.snowflake.com/en/sql-reference/identifiers-syntax.html#label-identifier-casing-parameter-howto
Update description for quoteIdentifiers property of the relational database connection.

| Tests added + passed?   | No                               |

### DIFF
--- a/.changeset/giant-keys-stare.md
+++ b/.changeset/giant-keys-stare.md
@@ -2,4 +2,4 @@
 "@finos/legend-studio": patch
 ---
 
-Add button for quotedIdentifiersIgnoreCase flag.
+Add button for `quotedIdentifiersIgnoreCase` flag.
--- a/.changeset/giant-keys-stare.md
+++ b/.changeset/giant-keys-stare.md
@@ -1,0 +1,5 @@
+---
+"@finos/legend-studio": patch
+---
+
+Add button for quotedIdentifiersIgnoreCase flag.
--- a/packages/legend-studio/src/components/editor/edit-panel/connection-editor/RelationalDatabaseConnectionEditor.tsx
+++ b/packages/legend-studio/src/components/editor/edit-panel/connection-editor/RelationalDatabaseConnectionEditor.tsx
@@ -527,14 +527,15 @@ const SnowflakeDatasourceSpecificationEditor = observer(
           }
         />
         {/* TODO: we should reconsider adding this field, it's an optional boolean, should we default it to `undefined` when it's `false`?*/}
-        {/* <ConnectionEditor_BooleanEditor
+        <ConnectionEditor_BooleanEditor
           isReadOnly={isReadOnly}
           value={sourceSpec.quotedIdentifiersIgnoreCase}
-          propertyName={'cloud type'}
-          update={(value: string | undefined): void =>
-            sourceSpec.setCloudType(value)
+          propertyName={'quoted identifiers ignore case'}
+          description={'Set this when account has this flag set as true'}
+          update={(value: boolean | undefined): void =>
+            sourceSpec.setQuotedIdentifiersIgnoreCase(Boolean(value))
           }
-        /> */}
+        />
       </>
     );
   },
@@ -1215,8 +1216,7 @@ const RelationalConnectionGeneralEditor = observer(
                   isReadOnly={isReadOnly}
                   value={connection.quoteIdentifiers}
                   propertyName="Quote identifiers"
-                  // TODO: change the description
-                  description="Use quote identifiers"
+                  description="Use this to quote sql identifiers"
                   update={(value?: boolean): void =>
                     connection.setQuoteIdentifiers(Boolean(value))
                   }
--- a/packages/legend-studio/src/components/editor/edit-panel/connection-editor/RelationalDatabaseConnectionEditor.tsx
+++ b/packages/legend-studio/src/components/editor/edit-panel/connection-editor/RelationalDatabaseConnectionEditor.tsx
@@ -1216,7 +1216,7 @@ const RelationalConnectionGeneralEditor = observer(
                   isReadOnly={isReadOnly}
                   value={connection.quoteIdentifiers}
                   propertyName="Quote identifiers"
-                  description="Use this to quote sql identifiers"
+                  description="Use this to quote SQL identifiers"
                   update={(value?: boolean): void =>
                     connection.setQuoteIdentifiers(Boolean(value))
                   }


.changeset/giant-keys-stare.md
```diff
@@ -0,0 +1,5 @@
+---
+"@finos/legend-studio": patch
+---
+
+Add button for quotedIdentifiersIgnoreCase flag.
```
akphi, line 5:
```suggestion
Add button for `quotedIdentifiersIgnoreCase` flag.
```

packages/legend-studio/src/components/editor/edit-panel/connection-editor/RelationalDatabaseConnectionEditor.tsx
```diff
@@ -1215,8 +1216,7 @@ const RelationalConnectionGeneralEditor = observer(
                   isReadOnly={isReadOnly}
                   value={connection.quoteIdentifiers}
                   propertyName="Quote identifiers"
-                  // TODO: change the description
-                  description="Use quote identifiers"
+                  description="Use this to quote sql identifiers"
```
akphi, line 1219:
```suggestion
                  description="Use this to quote SQL identifiers"
```

packages/legend-studio/src/components/editor/edit-panel/connection-editor/RelationalDatabaseConnectionEditor.tsx
```diff
@@ -527,14 +527,15 @@ const SnowflakeDatasourceSpecificationEditor = observer(
           }
         />
         {/* TODO: we should reconsider adding this field, it's an optional boolean, should we default it to `undefined` when it's `false`?*/}
```
akphi, line 529: Can you also remove this comment?

packages/legend-studio/src/components/editor/edit-panel/connection-editor/RelationalDatabaseConnectionEditor.tsx
```diff
@@ -527,14 +527,15 @@ const SnowflakeDatasourceSpecificationEditor = observer(
...
-        {/* <ConnectionEditor_BooleanEditor
+        <ConnectionEditor_BooleanEditor
           isReadOnly={isReadOnly}
           value={sourceSpec.quotedIdentifiersIgnoreCase}
-          propertyName={'cloud type'}
-          update={(value: string | undefined): void =>
-            sourceSpec.setCloudType(value)
+          propertyName={'quoted identifiers ignore case'}
+          description={'Set this when account has this flag set as true'}
```
akphi, line 534: This description is a bit unclear to me. What do you think @MauricioUyaguari?

akphi:
@hardikmaheshwari Could you please update the PR description so we can have some context when looking back at this in the future?

packages/legend-studio/src/components/editor/edit-panel/connection-editor/RelationalDatabaseConnectionEditor.tsx
hardikmaheshwari: I intentionally didn't remove this because I felt it's a valid point and maybe if in the future we have a better strategy we may want to use that here as well

packages/legend-studio/src/components/editor/edit-panel/connection-editor/RelationalDatabaseConnectionEditor.tsx
hardikmaheshwari: For a bit more context, quoted Identifiers Ignore Case is the name of an actual flag that can be configured at the snowflake account/session level
https://docs.snowflake.com/en/sql-reference/identifiers-syntax.html#label-identifier-casing-parameter-howto

akphi: Then should we try to use a summarized version of the description there as our description instead?

akphi: @hardikmaheshwari also, maybe you have missed the comment I made, we should update our PR description with better context.

hardikmaheshwari:
@akphi Added in some details. Please let me know if I am not following some convention or missing some detail that you think might be useful

akphi:
> @akphi Added in some details. Please let me know if I am not following some convention or missing some detail that you think might be useful

Great! Just one more thing is to update the table in the description, we don't have any tests here.

packages/legend-studio/src/components/editor/edit-panel/connection-editor/RelationalDatabaseConnectionEditor.tsx
akphi: Maybe something like `Treats alphabetic characters in double-quoted identifiers as uppercase`

hardikmaheshwari: Actually, I not very sure here as to what would be easy for users.
I am not sure if users would really realize when they need to set this flag if we have a functional explanation in the description. We need this because alloy expects case-sensitive resultSet from db by default but this flag is an exception.

I am not saying that users modeling in studio would be database experts but I expect if we have this description they either check their db config or get details from their db experts

hardikmaheshwari:
> > @akphi Added in some details. Please let me know if I am not following some convention or missing some detail that you think might be useful
> 
> Great! Just one more thing is to update the table in the description, we don't have any tests here.

@akphi done

packages/legend-studio/src/components/editor/edit-panel/connection-editor/RelationalDatabaseConnectionEditor.tsx
akphi: @hardikmaheshwari I see what you mean, but instructing them with `Set this when the account has this flag set as true` is not that helpful either because if you're targeting non-expert, that description sounds rather cryptic to me. I think we should assume that we are making a form editor for folks who know what they are doing with their Snowflake connection.

That aside, in the future, I believe we have a plan to allow people to scan the spaces for already created connections, maybe this will help with non-expert users.

hardikmaheshwari: I agree. 
Ok so to clarify setting this flag true/false does not make any difference to results from the user's side.
It's like a DB config which if set wrong, alloy won't be able to fetch results for the user.

So I believe, `Snowflake account config parameter` as the description makes sense to me.
I am still a bit resistant to the functional description `Treats alphabetic characters in double-quoted identifiers as uppercase` as this gives a feeling that setting this would change results.

Let me know your thoughts on this

akphi: Except this is the description for a boolean field, maybe `Instructs Snowflake to treat alphabetic characters in double-quoted identifiers as uppercase`

hardikmaheshwari: It seems this flag would be used by legend at session-level for services which is not the case.
Would give 1 last try at it (maybe we are overthinking this a bit as we can change the description with the user's feedback over time)

akphi: @pierredebelen what do you think?

akphi (approved):
Will make some modification afterwards to make sure description messages are consistent.

packages/legend-studio/src/components/editor/edit-panel/connection-editor/RelationalDatabaseConnectionEditor.tsx
akphi: Anyway, I'll merge this, but will change this to make it consistent with what we do in Studio in general. Will consult with @pierredebelen

Issue #324, opened 2021-07-08
Feature request: Query builder: Previewing data for mapped properties


akphi:
## 🚀 Feature Request

Allow previewing data for mapped properties.

> This assumes that the user already selected a `mapping` and a `runtime`

### Implementation ideas

- [x] ~~If the user haven't selected a runtime, disable preview or print warning notification~~ - right now we can't do this since `runtime` is always specified
- [x] Only support mapped properties and properties that do not branch from derived properties
- [x] For `non-numeric`: we ony show the values and their count: `->groupBy([x|$x.property], [agg(x|$x.property, x|$x.count())], ['Value', 'Count']])->take(10)->sort([desc('Count'), asc('Value')]`
- [x] For `numeric`: we show most of the aggregation for number: `->groupBy([], [count, ...], ['Count', 'Distinct Count', 'Sum', 'Min', 'Max', 'Average', 'Std Dev (Population)', 'Std Dev (Sample)'])`
- [x] Show preview result in a modal

PR #328, opened 2021-07-12
Query builder: add support for preview data

akphi:
Resolves https://github.com/finos/legend-studio/issues/324


packages/legend-studio-preset-query-builder/src/components/QueryBuilderProjectionPanel.tsx
```diff
@@ -278,7 +322,26 @@ const QueryBuilderProjectionColumnEditor = observer(
...
+            }
+            disabled={
+              !(
+                projectionColumnState instanceof
+                QueryBuilderSimpleProjectionColumnState
+              )
+            }
+            className={clsx('query-builder__projection__column__context-menu', {
+              'query-builder__projection__column--selected-from-context-menu':
+                isSelectedFromContextMenu,
```
MauricioUyaguari, line 339: testing

packages/legend-studio-preset-query-builder/src/stores/QueryBuilderProjectionState.ts
```diff
@@ -555,4 +597,125 @@ export class QueryBuilderProjectionState {
...
+        case PRIMITIVE_TYPE.NUMBER:
+        case PRIMITIVE_TYPE.INTEGER:
+        case PRIMITIVE_TYPE.DECIMAL:
+        case PRIMITIVE_TYPE.FLOAT: {
+          const previewResult =
+            (yield this.editorStore.graphState.graphManager.executeMapping(
+              this.editorStore.graphState.graph,
+              this.queryBuilderState.querySetupState.mapping,
+              this.queryBuilderState.buildRawLambdaFromLambdaFunction(
+                buildNumericPreviewDataQuery(
```
MauricioUyaguari, line 643: In preparation for the db builder, wouldn't it be better to have a more generic function we call that would do this switch case? something like `buildPreviewDataQuery` that calls the Numeric and NonNumericQuery builder.

akphi: yeah but I was not coding against that yet, trying to get this out quick and (arguably not so) dirty. We will co-op when you do that part

MauricioUyaguari (approved):

Issue #302, opened 2021-06-28
Bug: Query builder: Allow execute/build-query on entities from dependencies/generated elements


akphi:
## 🐛 Bug Report

Right-clicking to execute (i.e. opening query builder) right now is not supported for elements from dependencies and generated elements

MauricioUyaguari:
Reopening as the original MR enables query builder but doesn't total enable query builder as we don't fetch mappings/runtimes in the dependency/generation models.

akphi:
@MauricioUyaguari is this fixed now?

MauricioUyaguari:
Yes @akphi  closed by https://github.com/finos/legend-studio/pull/588

PR #359, opened 2021-07-21
enable query builder for generated/dependency elements

MauricioUyaguari:
Enable opening query builder from generated and dependency elements.
Fixes: https://github.com/finos/legend-studio/issues/302

### DIFF
--- a/packages/legend-studio/src/components/editor/side-bar/Explorer.tsx
+++ b/packages/legend-studio/src/components/editor/side-bar/Explorer.tsx
@@ -80,10 +80,11 @@ const ExplorerContextMenu = observer(
   (
     props: {
       node?: PackageTreeNodeData;
+      nodeIsImmutable?: boolean;
     },
     ref: React.Ref<HTMLDivElement>,
   ) => {
-    const { node } = props;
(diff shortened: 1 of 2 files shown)


packages/legend-studio/src/components/editor/side-bar/Explorer.tsx
```diff
@@ -300,7 +304,9 @@ const PackageTreeNodeContainer = observer(
 
     return (
       <ContextMenu
-        content={<ExplorerContextMenu node={node} />}
+        content={
+          <ExplorerContextMenu node={node} nodeIsImmutable={readOnlyContext} />
```
akphi, line 308: why don't we add this as property of the node?

akphi: more work I guess, but probably cleaner than this. `isImmutable` or `isReadOnly` on the node itself might be useful for other things as well.

akphi (approved):

akphi:
The way we organize the state is a bit messy imo, we probably should come back to clean this up when we have the time 😭

Issue #280, opened 2021-06-23
Feature request: Allow excluding certain system elements in element selectors 


akphi:
## 🚀 Feature Request

As we allow plugins to add more system elements, we need to be careful in what to show or not to show in the element selector (e.g. `class selector` in `class editor > properties`). This would be a list of element paths that the system uses for filtering. Also, we should create a new method per graph manager plugin to allow plugins to specify their own `allowedlist`.

PR #320, opened 2021-07-07
Fix: Association mapping sourceID omission and hashing issue

kshradhan:
| Q                       | A                                |
| ----------------------- | -------------------------------- |
| Fixed issues?           | Fixes #278  |
| Tests added + passed?   | Yes                              |
| Changeset added?        | Yes                              |
| Documentation PR link   |                                  |
| Any dependency changes? | No                               |

Fixes #278. 
Solves the issue where the source Id in grammar text is omitted from studio's protocol classes.
This change also requires a change in engine. PR for which is already out here: [engine-#300](https://github.com/finos/legend-engine/pull/301)

PR #363, opened 2021-07-22
Some more code cleanups

akphi:
Fixes https://github.com/finos/legend-studio/issues/280

- [x] **BREAKING CHANGE** Rename methods in `BasicModel` to avoid confusion with methods coming from `PureModel`: We add the keyword `own` to indicate methods coming from `BasicModel`
- [x] Also allow double-clicking label to edit property holder view in diagram editor.
- [x] Cleanup usage of `explicit reference` when building the graph. In theory, we should not have to use any of those (except for very special cases). For cases we haven't been able to convert from `explicit` to `implicit`, we have put a note to explain the risk and mark it to revise in the future as they can cause bugs in roundtrips.
- [x] Addressed residuals from https://github.com/finos/legend-studio/pull/320#pullrequestreview-710508396


.changeset/hot-carpets-relax.md
```diff
@@ -0,0 +1,5 @@
+---
+'@finos/legend-studio': patch
+---
+
+**BREAKING CHANGE:** Rename methods in `BasicModel`: add `own` to methods name to avoid confusion on when consumers are interacting with elements from the whole graph (in `PureModel`) or elements just from the graph itself.
```
MauricioUyaguari, line 5: You break you fix.

MauricioUyaguari (approved):

PR #364, opened 2021-07-22
feat:Added support for redshift connector

varunmaddipati:
Adding support for a redshift connector. 
Reference : https://docs.aws.amazon.com/redshift/latest/mgmt/welcome.html


packages/legend-studio/src/components/editor/edit-panel/connection-editor/RelationalDatabaseConnectionEditor.tsx
```diff
@@ -40,13 +40,16 @@ import {
   DelegatedKerberosAuthenticationStrategy,
   OAuthAuthenticationStrategy,
   SnowflakePublicAuthenticationStrategy,
+  TestDatabaseAuthenticationStrategy,
```
hardikmaheshwari, line 43: Why is this import needed - `TestDatabase...`

packages/legend-studio/src/stores/__tests__/roundtrip/RelationalRoundtripTestData.ts
```diff
@@ -3812,16 +3812,17 @@ export const testRelationalDatabaseConnectionRoundtrip = [
       connectionValue: {
         _type: 'RelationalDatabaseConnection',
         authenticationStrategy: {
-          _type: 'oauth',
-          oauthKey: 'dummy',
-          scopeName: 'UserPass',
+          _type: 'userPassword',
```
hardikmaheshwari, line 3815: We should not be modifying existing test
We need to append our test-data

packages/legend-studio/src/stores/editor-state/element-editor-state/connection/ConnectionEditorState.ts
```diff
@@ -178,6 +184,12 @@ export class RelationalDatabaseConnectionValueState extends ConnectionValueState
         );
         return;
       }
+      case CORE_DATASOURCE_SPEC_TYPE.REDSHIFT: {
+        this.connection.setDatasourceSpecification(
+          new RedshiftDatasourceSpecification('', '', 5439),
```
MauricioUyaguari, line 189: can we default to port 80 to make it consistent with the other default port numbers, or is there a reason we default to 5439?

varunmaddipati: I used 5439 because that is the port that Redshift uses by default

MauricioUyaguari (approved):

Issue #373, opened 2021-07-27
RFC: Execution Plan Cleanup


MauricioUyaguari:
## 💬 Request for Comments

Cleanup and bug fixes related to https://github.com/finos/legend-studio/issues/249
and pull request 

### Clean ups: 
- [x] Move out execution plan viewer to its own component
- [x] Share across test/execute/service
- [x] Fix Graph Fetch Crash on opening tests
- [x] use  PackageableElementImplicitReference for TDS Column type
- [x] Move executionPlan, executionPlanMeta inside the executionPlanState

PR #374, opened 2021-07-27
Execution Plan Cleanups

MauricioUyaguari:
Execution Plan Clean ups.
Fixes https://github.com/finos/legend-studio/issues/249
Fixes https://github.com/finos/legend-studio/issues/373

### DIFF
--- a/packages/legend-studio/src/components/editor/edit-panel/mapping-editor/MappingTestEditor.tsx
+++ b/packages/legend-studio/src/components/editor/edit-panel/mapping-editor/MappingTestEditor.tsx
@@ -17,7 +17,6 @@
 import { Fragment, useState, useEffect, useCallback } from 'react';
 import { observer } from 'mobx-react-lite';
 import { ReflexContainer, ReflexElement, ReflexSplitter } from 'react-reflex';
-import Dialog from '@material-ui/core/Dialog';
 import type { MappingTestState } from '../../../../stores/editor-state/element-editor-state/mapping/MappingTestState';
 import {
   MAPPING_TEST_EDITOR_TAB_TYPE,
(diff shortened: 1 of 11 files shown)


akphi:
Don't forget to mention this fixes #249 and #373

packages/legend-studio/src/components/editor/edit-panel/mapping-editor/MappingTestEditor.tsx
```diff
@@ -643,7 +630,7 @@ export const MappingTestEditor = observer(
               className="mapping-test-editor__header__action"
               onClick={generatePlan}
               tabIndex={-1}
-              title="View Execution Plan"
+              title="View Execution Plancc"
```
akphi, line 633:
```suggestion
              title="View Execution Plan"
```

packages/legend-studio/src/stores/ExecutionPlanState.ts
```diff
@@ -42,30 +48,41 @@ export class ExecutionPlanState {
...
+  isGenerating = false;
 
   constructor(editorStore: EditorStore) {
     makeObservable(this, {
       displayData: observable,
       displayDataJson: observable,
       sqlSelectedTab: observable,
       viewMode: observable,
+      isGenerating: observable,
+      plan: observable,
```
akphi, line 61: what do you think if we separate this and make `plan?: ExecutionPlan` and `rawPlan!: object` and do `observable.ref` for the latter? When we can't build the plan, we can make `plan = undefined` and thus fall back to display JSON?

packages/legend-studio/src/components/editor/edit-panel/mapping-editor/execution-plan-viewer/ExecutionPlanViewer.tsx
```diff
@@ -0,0 +1,400 @@
...
+          root: 'editor-modal__root-container',
+          container: 'editor-modal__container',
+          paper: 'editor-modal__content',
+        }}
+      >
+        <div className="modal modal--dark editor-modal">
+          <div className="modal__header">
+            <div className="modal__title">Execution Plan</div>
+          </div>
+          {plan instanceof ExecutionPlan ? (
```
akphi, line 348: gracias! 🙏

akphi (dismissed):

packages/legend-studio/src/stores/ExecutionPlanState.ts
MauricioUyaguari: So I would favor having the following:

`
plan?: ExecutionPlan
rawPlan?: object
`
Thus when closing the modal we set both to undefined. When generating the plan we populate rawPlan always and if possible we also pollute the plan. 

Thoughts?

akphi: oh so I forgot we use `.plan` as dialog trigger. Then yeah I agree with your suggestion there. But we should do something like this, when `rawPlan` is set, we will show the plan viewer, but as we build it into `plan` we should show some loading indicator. If we load successfully, we will show the proper execution plan viewer, otherwise, fallback to the old JSON plan viewer.

pierredebelen (approved):

Issue #297, opened 2021-06-28
RFC: Viewer mode improvements


akphi:
## 💬 Request for Comments

Viewer mode allows user to enter a read-only workspace that corresponds to the current HEAD of the project, we should come up with a set of features that we should allow in this mode. Modifications to models should be blocked. This can set us up on the right path should we update review mode to also allow users to interact with the read-only workspace.

### Implementation Plan
- [ ] Allow seeing project configuration
- [ ] Allow hackermode/text-mode
- [ ] Allow opening model loader
- [ ] Allow using query builder

PR #379, opened 2021-07-29
Enable project configuration viewing in viewer mode

umarphaarook:
| Fixed issues?           | Related to #297 |


packages/legend-studio/src/stores/ViewerStore.tsx
```diff
@@ -198,6 +199,16 @@ export class ViewerStore {
             undefined,
           )) as Entity[];
       }
+
+      // Get Project Config
+      const projectConfiguration: PlainObject<ProjectConfiguration> =
+        flowResult(
+          yield this.editorStore.applicationStore.networkClientManager.sdlcClient.getConfiguration(
```
MauricioUyaguari, line 206: Did we account for viewer mode where workspaceId would not be defined like through a version or revisionID?

umarphaarook: Nope, the logic within `ProjectConfigurationEditor` component always expects `projectConfiguration` to be present in the store. `EditorStore` does load the `projectConfiguration` in standard mode while `ViewerStore` doesn't, hence the crash on the app whenever user try to view project `config`  in viewer mode.

MauricioUyaguari: yeah understood, my question was more when no workspace is provided, which occurs when you are in viewer mode for released version or on specific revision.

umarphaarook: I couldn't find any logic that handles that, so I assumed we don't

MauricioUyaguari: lets test your changes locally with a revision/version id and make sure this still works.

packages/legend-studio/src/stores/ViewerStore.tsx
```diff
@@ -194,12 +197,35 @@ export class ViewerStore {
...
+            ),
+          ])) as [Entity[], PlainObject<ProjectConfiguration>];
+          entities = result[0];
+          projectConfiguration = result[1];
+          this.editorStore.projectConfigurationEditorState.setProjectConfiguration(
+            ProjectConfiguration.serialization.fromJson(projectConfiguration),
+          );
+          // make sure we set the original project configuration to a different object
+          this.editorStore.projectConfigurationEditorState.setOriginalProjectConfiguration(
+            ProjectConfiguration.serialization.fromJson(projectConfiguration),
```
MauricioUyaguari, line 219: we should only be doing the project configuration once before setting them

packages/legend-studio/src/stores/ViewerStore.tsx
```diff
@@ -152,7 +153,9 @@ export class ViewerStore {
           `Can't have both version ID and revision ID specified for viewer mode`,
         );
       }
+
       let entities: Entity[] = [];
+      let projectConfiguration: PlainObject<ProjectConfiguration>;
```
MauricioUyaguari, line 158: can we name this `rawProjectConfiguration`
and the serialized one `projectConfiguration`
thats more inline with what we do in the rest of the app

packages/legend-studio/src/stores/ViewerStore.tsx
```diff
@@ -194,12 +196,39 @@ export class ViewerStore {
...
+              undefined,
+            ),
+            this.editorStore.applicationStore.networkClientManager.sdlcClient.getConfiguration(
+              this.editorStore.sdlcState.currentProjectId,
+              undefined,
+            ),
+          ])) as [Entity[], PlainObject<ProjectConfiguration>];
+          entities = result[0];
+          const rawProjectConfiguration = result[1];
+          const serializedProjectConfiguration =
```
MauricioUyaguari, line 213: `serializedProjectConfiguration` -> `projectConfiguration`

MauricioUyaguari (approved):

PR #505, opened 2021-09-15
Add graph support for DSL Serializer

gayathrir11:
| Q                       | A                                |
| ----------------------- | -------------------------------- |
| Fixed issues?           | Closes https://github.com/finos/legend-studio/issues/507 |
| Tests added + passed?   | Yes                              |
| Changeset added?        | Yes                              |
| Documentation PR link   |                                  |
| Any dependency changes? | No                               |

PR #534, opened 2021-09-29
Add Form Support for SchemaSet

gayathrir11:

pic1
pic2


akphi:
Looks like you need to properly rebase your branch locally. Also, we forgot to mention in the last PR, could you add a grammar roundtrip test for DSL serializer?

gayathrir11:
Where can I add grammar roundtrip test?

akphi:
Why service store? Also, could you fix your description to use proper markdown syntax for code, i.e. the triple backticks? - See https://guides.github.com/features/mastering-markdown/

packages/legend-extension-dsl-serializer/style/index.scss
```diff
@@ -0,0 +1,278 @@
...
+
+.icon--schema-set {
+  color: var(--color-schema-set);
+}
+
+.icon--binding {
+  color: var(--color-binding);
+}
+
+.format-editor {
```
akphi, line 32: if we have a lot like this, probably it's better to move these to another file and import back into this main one

packages/legend-extension-dsl-serializer/style/index.scss
```diff
@@ -0,0 +1,278 @@
...
+  height: 100%;
```
akphi, line 33: now that I notice it, you should spell this out to `external-format-connection-editor` or something like that

packages/legend-extension-dsl-serializer/src/stores/studio/BindingEditorState.ts
```diff
@@ -19,7 +19,7 @@ import type { EditorStore } from '@finos/legend-studio';
 import { guaranteeType } from '@finos/legend-shared';
 import { ElementEditorState } from '@finos/legend-studio';
 import type { PackageableElement } from '@finos/legend-graph';
-import { Binding } from '../models/metamodels/pure/model/packageableElements/store/Binding';
+import { Binding } from '../../models/metamodels/pure/model/packageableElements/store/Binding';
 
 export class BindingEditorState extends ElementEditorState {
```
akphi, line 24: As commented before, since you don't have an editor for `Binding` yet, why do we still keep this file?

packages/legend-extension-dsl-serializer/src/models/metamodels/pure/model/packageableElements/schemaSet/Schema.ts
```diff
@@ -24,11 +24,14 @@ export class Schema implements Hashable {
   location?: string | undefined;
   content: string;
 
-  constructor(id: string, location: string, content: string) {
+  constructor(content: string, id?: string, location?: string) {
```
akphi, line 27: Let's avoid this, passing optional parameters to constructor like this is less than ideal, also remember what we said last time about using a `dumb` constructor, let's get rid of the parameters and populate them after instantiation

akphi:
@gayathrir11 generally, I'm okay with this PR, but I think this is simple enough for us to adopt dark theme. We are moving towards converting Studio to be 100% dark-themed to be consistent, could you refer to service editor or runtime editor for example and follow the styling there? Also when you finish that, please update the preview accordingly. Thanks!

akphi:
@gayathrir11 also, next time, you should try to `rebase` your branch off the default branch `master` instead of creating a merge commit on your branch, your commit history would be clearer then.

akphi:
@gayathrir11 also, you don't need to fill in the `fixes issue` box in your PR description next time if you don't have an issue. And the right syntax is to write `fixes #534` instead of `#534` as the keyword there will trigger Github to close the issue when this PR is closed.

akphi:
@gayathrir11 So a few concerns:
- Could we get rid of the border around the list of schemas?
- Let's add 0.5rem padding for the content of the schema viewer panel (right)
- Why can't we use `TextInputEditor` instead of `<input>` for the content?

gayathrir11:
@akphi do you want me to get rid of expand/collapse toggle for content?

akphi:
> @akphi do you want me to get rid of expand/collapse toggle for content?

Just use `TextInputEditor` component, it does not include the expand toggle

akphi:
@gayathrir11 I tried the roundtrip test you posted in the thread in Studio and I got the error message `Compilation failed: Unknown schema format: FlatData`

gayathrir11:
> @gayathrir11 I tried the roundtrip test you posted in the thread in Studio and I got the error message `Compilation failed: Unknown schema format: FlatData`

@akphi dependency for FlatData is missing in pom.xml file of legend-engine-server

akphi:
> > @gayathrir11 I tried the roundtrip test you posted in the thread in Studio and I got the error message `Compilation failed: Unknown schema format: FlatData`
> 
> @akphi dependency for FlatData is missing in pom.xml file of legend-engine-server

So we should fix that on engine side first

packages/legend-extension-dsl-serializer/src/models/protocols/pure/DSLSerializer_PureProtocolProcessorPlugin.ts
```diff
@@ -148,7 +149,13 @@ export class DSLSerializer_PureProtocolProcessorPlugin extends DSLMapping_PurePr
           );
           const element = getSchemaSet(path, context.graph);
           element.format = elementProtocol.format;
-          element.schemas = elementProtocol.schemas;
+          element.schemas = elementProtocol.schemas.map((schema) => {
+            const schemaElement = new Schema();
```
akphi, line 153: You lack the assertions here, you should refer to how we do it here - https://github.com/finos/legend-studio/pull/557/commits/7ab13787309722be6a58a5ea0f5c4a7ec73c1718

akphi: Notice how back in #505, we missed a few places where `V1_` objects are assigned to the metamodel which is quite bad.

packages/legend-extension-dsl-serializer/src/models/metamodels/pure/model/packageableElements/store/ModelUnit.ts
```diff
@@ -23,18 +23,15 @@ export class ModelUnit implements Hashable {
...
+  constructor() {
     makeObservable(this, {
       packageableElementIncludes: observable,
       packageableElementExcludes: observable,
       hashCode: computed,
     });
 
-    this.packageableElementIncludes = packageableElementIncludes;
-    this.packageableElementExcludes = packageableElementExcludes;
+    this.packageableElementIncludes = [];
```
akphi, line 33: Let's follow the convention and do when you specify the field - after rebasing, you don't need this change due to #557

packages/legend-extension-dsl-serializer/src/models/metamodels/pure/model/packageableElements/store/Binding.ts
```diff
@@ -44,7 +44,7 @@ export class Binding extends Store implements Hashable {
     });
 
     this.contentType = '';
-    this.modelUnit = new ModelUnit([], []);
+    this.modelUnit = new ModelUnit();
```
akphi, line 47: This should not be needed after #557

packages/legend-extension-dsl-serializer/src/models/metamodels/pure/model/packageableElements/schemaSet/Schema.ts
```diff
@@ -24,17 +24,30 @@ export class Schema implements Hashable {
...
+      setId: action,
+      setLocation: action,
+      setContent: action,
       hashCode: computed,
     });
 
-    this.id = id;
-    this.location = location;
-    this.content = content;
+    this.content = '';
```
akphi, line 38: Let's avoid initializing the value in constructor

packages/legend-extension-dsl-serializer/src/models/metamodels/pure/model/packageableElements/schemaSet/Schema.ts
```diff
@@ -24,17 +24,30 @@ export class Schema implements Hashable {
   location?: string | undefined;
   content: string;
```
akphi, line 25: Do `content!: string` instead

packages/legend-extension-dsl-serializer/src/components/studio/SchemaSetElementEditor.tsx
```diff
@@ -0,0 +1,285 @@
...
+    );
+  },
+);
+
+export const SchemaSetEditor = observer(() => {
+  const editorStore = useEditorStore();
+  const editorState = editorStore.getCurrentEditorState(SchemaSetEditorState);
+  const schemaSet = editorState.schemaSet;
+  const isReadOnly = editorState.isReadOnly;
+  schemaSet.setFormat(FORMAT_TYPE.FLAT_DATA);
```
akphi, line 141: This is a `no-no`, you're modifying the state in the render method.

akphi: Instead, when you initialize the state, you could have set this if this is not set already, if you do this, it means whatever the format is, it's always reset to `FLAT_DATA` which I believe is not what we intend to do

packages/legend-extension-dsl-serializer/src/models/metamodels/pure/model/packageableElements/store/ModelUnit.ts
gayathrir11: reverted those changes

packages/legend-extension-dsl-serializer/src/components/studio/SchemaSetElementEditor.tsx
```diff
@@ -0,0 +1,280 @@
...
+  (props: { schema: Schema; isReadOnly: boolean }) => {
+    const { schema, isReadOnly } = props;
+    const changeId: React.ChangeEventHandler<HTMLInputElement> = (event) =>
+      schema.setId(event.target.value);
+    const changeLocation: React.ChangeEventHandler<HTMLInputElement> = (
+      event,
+    ) => schema.setLocation(event.target.value);
+    const editorOptions: editor.IEditorOptions & editor.IGlobalEditorOptions = {
+      lineNumbers: 'off',
+      fontFamily: 'Arial',
```
akphi, line 94: why do we use `Arial` instead of using the default config for `TextInputEditor`

packages/legend-extension-dsl-serializer/src/models/protocols/pure/DSLSerializer_PureProtocolProcessorPlugin.ts
```diff
@@ -170,11 +175,22 @@ export class DSLSerializer_PureProtocolProcessorPlugin extends DSLMapping_PurePr
             elementProtocol.name,
           );
           const element = getSchemaSet(path, context.graph);
-          element.format = guaranteeNonEmptyString(
-            elementProtocol.format,
-            `Schema set 'format' field is missing or empty`,
-          );
-          element.schemas = elementProtocol.schemas;
+          element.format =
+            Object.values(FORMAT_TYPE).find(
```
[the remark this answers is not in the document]
gayathrir11: You would get an error message from engine side if we don't support the format type. Do you want me to include the condition here as well?

packages/legend-extension-dsl-serializer/src/components/studio/SchemaSetElementEditor.tsx
gayathrir11: > why do we use `Arial` instead of using the default config for `TextInputEditor`. You should check out the API of `TextInputEditor` to see the options

Id, location uses different fontFamily from the default config for TextInputEditor. It was looking really weird. So I changed it.

packages/legend-extension-dsl-serializer/src/models/protocols/pure/DSLSerializer_PureProtocolProcessorPlugin.ts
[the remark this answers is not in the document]
gayathrir11: > I think this is not proper, since we are building the graph from the input, if the input contains a format type that we don't support, we should throw.

Compilation fails and you would get an error message from engine side if we don't support the format type. Do you want me to include the condition here as well?

packages/legend-extension-dsl-serializer/src/components/studio/SchemaSetElementEditor.tsx
akphi: what do you mean by `location` uses a different font? Could you attach a screenshot so I could see how weird it looks?

packages/legend-extension-dsl-serializer/src/models/protocols/pure/DSLSerializer_PureProtocolProcessorPlugin.ts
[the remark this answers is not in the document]
akphi: yes, I think that's fair to throw if they give us a messed up protocol model

packages/legend-extension-dsl-serializer/src/components/studio/SchemaSetElementEditor.tsx
gayathrir11: > what do you mean by `location` uses a different font? Could you attach a screenshot so I could see how weird it looks?

pic3

Issue #474, opened 2021-09-12
RFC: Refactor to use SDLC `Workflow` API


akphi:
## 💬 Request for Comments

`Build` API has been deprecated, we should adapt to use `Workflow` instead 

### Implementation plan
- [x] Add `Workflow` and `WorkflowJob` models
- [x] Add APIs for `Workflow` in `SDLC client`
- [x] Refactor `Builds` side bar to use `Workflow`
- [x] Rename panel to `Workflow Jobs` or `Workflow` instead

PR #568, opened 2021-10-11
Group workspace support and workspace workflows

MauricioUyaguari:
Add support for group workspaces. 
Fixes https://github.com/finos/legend-studio/issues/461. 
Fixes https://github.com/finos/legend-studio/issues/474.

### DIFF
--- a/packages/legend-server-sdlc/src/SDLCServerClient.ts
+++ b/packages/legend-server-sdlc/src/SDLCServerClient.ts
@@ -16,10 +16,7 @@
 
 import type { Project, SdlcMode, ProjectType } from './models/project/Project';
 import type { ImportReport } from './models/project/ImportReport';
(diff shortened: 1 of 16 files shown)


akphi:
So Kevin made a fairly valid point for this. SDLC `Jackson` object mapper will serialize `null` to `null`. This is because we want the server's response to be lossless, in order to be able to distinguish between `missing` and `null` cases. This is fair. But it is a limitation to the SDLC models that we have: some of the fields should have been typed with `Optional<T>` instead of `String` to indicate its optionality. So with this, I think in Studio, we keep the model as `someField?: string | undefined` and do some surgery on the deserializer to coerce `nullish` value to `undefined`. I want the model in Studio at least to make the distinction between `missing` and `set to null`.

PR #629, opened 2021-10-29
Input to ServiceTestAssertLambda is updated to regular string

### LINKED ISSUE #586: Bug: Escape character is not correctly handled in Service Test

Returns the service test assert as Fail, and the comparison for actual & expected shows no difference.
Reason: Additional backslash is being added to the expected result string in Studio prior to triggering the doTest() endpoint.

### Steps to reproduce

- Switch to text mode and copy the entire code-block to your workspace to replicate.
- Run the tests under PersonService service.
- Tests will fail, check the comparison.

### Model data
```
###FlatData
FlatData model::FlatDataStore
{
  section default: DelimitedWithHeadings
  {
    scope.untilEof;
    delimiter: ',';
    quoteChar: '"';
    nullString: '';

    Record
    {
      DATE_COL: DATE(format='MM/dd/yyyy', optional);
    }
  }
}


###Service
Service model::PersonService
{
  pattern: '/a4a67e91-1f4f-4226-9949-43f872b8c8a8';
  documentation: '';
  autoActivateUpdates: true;
  execution: Single
  {
    query: |model::Person.all()->graphFetchChecked(#{model::Person{DOB}}#)->serialize(#{model::Person{DOB}}#, ^meta::pure::graphFetch::execution::AlloySerializationConfig(typeKeyName='@type' , removePropertiesWithNullValues=true));
    mapping: model::PersonMapping;
    runtime:
    #{
      mappings:
      [
        model::PersonMapping
      ];
      connections:
      [
        model::FlatDataStore:
        [
          connection_1:
          #{
            FlatDataConnection
            {
              store: model::FlatDataStore;
              url: 'executor:default';
            }
          }#
        ]
      ];
    }#;
  }
  test: Single
  {
    data: 'DATE_COL\r\n ,0001-01-01';
    asserts:
    [
      { [], res: meta::pure::mapping::Result[1]|$res.values->toOne()->toString()->equalJsonStrings('{"defects":[],"source":{"defects":[{"message":"Failed to read \'DATE_COL\' with value:  ,0001-01-01, error: ParseException Unparseable date: \\" ,0001-01-01\\"","enforcementLevel":"Error","ruleType":"InvalidInput","ruleDefinerPath":"model::FlatDataStore","path":[]}],"source":{"number":1,"lineNumber":2,"record":" ,0001-01-01","recordValues":[{"address":"DATE_COL","rawValue":" ,0001-01-01"}]},"value":{"typeName":"model::FlatDataStore.default.default","values":[]}},"value":{}}') }
    ];
  }
}


###Pure
Class model::Person
{
  DOB: StrictDate[0..1];
}


###Mapping
Mapping model::PersonMapping
(
  *model::Person[accounts_classA]: FlatData
  {
    ~src model::FlatDataStore.default
    DOB: $src['DATE_COL']
  }

  MappingTests
  [
    test_1
    (
      query: |model::Person.all()->graphFetchChecked(#{model::Person{DOB}}#)->serialize(#{model::Person{DOB}}#);
      data:
      [
        <FlatData, model::FlatDataStore, 'DATE_COL\r\n ,0001-01-01'>
      ];
      assert: '{"defects":[],"source":{"defects":[{"id":null,"externalId":null,"message":"Failed to read \'DATE_COL\' with value:  ,0001-01-01, error: ParseException Unparseable date: \" ,0001-01-01\"","enforcementLevel":"Error","ruleType":"InvalidInput","ruleDefinerPath":"model::FlatDataStore","path":[]}],"source":{"number":1,"lineNumber":2,"record":" ,0001-01-01","recordValues":[{"address":"DATE_COL","rawValue":" ,0001-01-01"}]},"value":{"typeName":"model::FlatDataStore.default.default","values":[]}},"value":{"DOB":null}}';
    )
  ]
)

```

### Additional Context 

We (@MauricioUyaguari, @mrudula-gs, and @akphi) have summarised the root cause in https://github.com/finos/legend-engine/issues/429. For now, because we noticed this discrepancy in the behaviour of mapping and service test runners, we will compensate for that in Studio. We will change the way we handle assertion data for service test lambda by removing the `toGrammarString()` and `fromGrammarString()` calls when we read-in and write-out the assertion string.

In our fix, we will mention this and the issue in engine for context of the fix/workaround

mrudula-gs:
## Summary

Closes #586 

Due to discrepancies in the test runners for mapping and service, we have don't need to do any (un)escaping for service test assertion data like what we do for mapping test assertion data. For better context:
See https://github.com/finos/legend-studio/issues/586 
See https://github.com/finos/legend-engine/issues/429

Check with the test case in the linked issue

### DIFF
--- a/.changeset/little-turkeys-hammer.md
+++ b/.changeset/little-turkeys-hammer.md
@@ -1,5 +1,4 @@
 ---
-'@finos/legend-graph': patch
 '@finos/legend-studio': patch
 ---
 
--- a/.changeset/little-turkeys-hammer.md
+++ b/.changeset/little-turkeys-hammer.md
@@ -1,0 +1,8 @@
+---
+'@finos/legend-graph': patch
+'@finos/legend-studio': patch
+---
+
+Due to discrepancies in the test runners for mapping and service, we don't need to do any (un)escaping for the expectedResult assertion data of the service test during its initialization and generation like what we do for mapping test assertion data. So, removing the toGrammarString & fromGrammarString pair for service test runner to avoid it adding additional escape characters. For better context:
+// See https://github.com/finos/legend-studio/issues/586
+// See https://github.com/finos/legend-engine/issues/429
--- a/.changeset/little-turkeys-hammer.md
+++ b/.changeset/little-turkeys-hammer.md
@@ -5,4 +5,3 @@
 
 Due to discrepancies in the test runners for mapping and service, we don't need to do any (un)escaping for the expectedResult assertion data of the service test during its initialization and generation like what we do for mapping test assertion data. So, removing the toGrammarString & fromGrammarString pair for service test runner to avoid it adding additional escape characters. For better context:
 // See https://github.com/finos/legend-studio/issues/586
-// See https://github.com/finos/legend-engine/issues/429
--- a/.changeset/little-turkeys-hammer.md
+++ b/.changeset/little-turkeys-hammer.md
@@ -4,4 +4,3 @@
 ---
 
 Due to discrepancies in the test runners for mapping and service, we don't need to do any (un)escaping for the expectedResult assertion data of the service test during its initialization and generation like what we do for mapping test assertion data. So, removing the toGrammarString & fromGrammarString pair for service test runner to avoid it adding additional escape characters. For better context:
-// See https://github.com/finos/legend-studio/issues/586
--- a/.changeset/little-turkeys-hammer.md
+++ b/.changeset/little-turkeys-hammer.md
@@ -3,4 +3,4 @@
 '@finos/legend-studio': patch
 ---
 
-Due to discrepancies in the test runners for mapping and service, we don't need to do any (un)escaping for the expectedResult assertion data of the service test during its initialization and generation like what we do for mapping test assertion data. So, removing the toGrammarString & fromGrammarString pair for service test runner to avoid it adding additional escape characters. For better context:
+Fix a problem with escaping of single quote character which causes service tests created in Studio fail ([#586](See https://github.com/finos/legend-studio/issues/586)), this can be considered a workaround until we figure out a strategy for the discrepancies in mapping test and service test runners in `Engine` (see [finos/legend-engine#429](https://github.com/finos/legend-engine/issues/429))
--- a/packages/legend-graph/src/graphManager/AbstractPureGraphManager.ts
+++ b/packages/legend-graph/src/graphManager/AbstractPureGraphManager.ts
@@ -345,6 +345,7 @@ export abstract class AbstractPureGraphManager {
   // structurally in Studio
 
   abstract HACKY_createGetAllLambda(_class: Class): RawLambda;
+  // NOTE: Input should be a regular string, can't be escaped.
   abstract HACKY_createServiceTestAssertLambda(assertData: string): RawLambda;
   abstract HACKY_extractServiceTestAssertionData(
     query: RawLambda,
--- a/packages/legend-graph/src/graphManager/AbstractPureGraphManager.ts
+++ b/packages/legend-graph/src/graphManager/AbstractPureGraphManager.ts
@@ -345,7 +345,6 @@ export abstract class AbstractPureGraphManager {
   // structurally in Studio
 
   abstract HACKY_createGetAllLambda(_class: Class): RawLambda;
-  // NOTE: Input should be a regular string, can't be escaped.
   abstract HACKY_createServiceTestAssertLambda(assertData: string): RawLambda;
   abstract HACKY_extractServiceTestAssertionData(
     query: RawLambda,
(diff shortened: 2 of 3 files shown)


akphi:
Did we file a bug for this? We should link it in the PR description and the changeset.

.changeset/short-camels-retire.md
```diff
@@ -0,0 +1,5 @@
+---
+"@finos/legend-studio": patch
+---
+
+Input to ServiceTestAssertLambda is updated to regular string
```
MauricioUyaguari, line 5: There should only be one changeset.

packages/legend-graph/src/graphManager/AbstractPureGraphManager.ts
```diff
@@ -345,6 +345,7 @@ export abstract class AbstractPureGraphManager {
   // structurally in Studio
 
   abstract HACKY_createGetAllLambda(_class: Class): RawLambda;
+  // NOTE: Input should be a regular string, can't be escaped.
```
MauricioUyaguari, line 348: We need to provide a full explanation of what this function does.
Alongside a more descriptive note on what the input `assertData` should be

akphi:
@mrudula-gs 2 things
1. Could you properly link your PR to the issue? - See https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword
2. For the 2 places that we make modifications, let's use the following comment as clarification

```
// NOTE: due to discrepancies in the test runners for mapping and service, we have don't need 
// to do any (un)escaping here like what we do for mapping test assertion data. For better context:
// See https://github.com/finos/legend-studio/issues/586 
// See https://github.com/finos/legend-engine/issues/429
```

Thanks!

packages/legend-graph/src/graphManager/AbstractPureGraphManager.ts
akphi: @mrudula-gs @MauricioUyaguari see my comment for the comment you need to add.

MauricioUyaguari: Lets remove this note as we have the comments directly on the service test

.changeset/modern-cups-arrive.md
```diff
@@ -0,0 +1,6 @@
+---
+"@finos/legend-graph": patch
+"@finos/legend-studio": patch
+---
+
+Removed toGrammarString & fromGrammarString pair for service test runner
```
MauricioUyaguari, line 6: Can we be more descriptive on why we are doing this. We should also link up the engine issues. Changesets will be part of the change logs when we do a release so think about what your change is doing overall. The message should explain why the change was made.

.changeset/little-turkeys-hammer.md
```diff
@@ -0,0 +1,8 @@
+---
+'@finos/legend-graph': patch
+'@finos/legend-studio': patch
+---
+
+Due to discrepancies in the test runners for mapping and service, we don't need to do any (un)escaping for the expectedResult assertion data of the service test during its initialization and generation like what we do for mapping test assertion data. So, removing the toGrammarString & fromGrammarString pair for service test runner to avoid it adding additional escape characters. For better context:
```
akphi, line 6:
```suggestion
Fix a problem with escaping of single quote character which causes service tests created in Studio fail ([#586](See https://github.com/finos/legend-studio/issues/586)), this can be considered a workaround until we figure out a strategy for the discrepancies in mapping test and service test runners in `Engine` (see [finos/legend-engine#429](https://github.com/finos/legend-engine/issues/429))
```

.changeset/little-turkeys-hammer.md
```diff
@@ -0,0 +1,8 @@
...
+// See https://github.com/finos/legend-studio/issues/586
```
akphi, line 7:
```suggestion
```

.changeset/little-turkeys-hammer.md
```diff
@@ -0,0 +1,8 @@
...
+// See https://github.com/finos/legend-engine/issues/429
```
akphi, line 8:
```suggestion
```

.changeset/little-turkeys-hammer.md
```diff
@@ -0,0 +1,6 @@
+---
+'@finos/legend-graph': patch
```
akphi, line 2:
```suggestion
```

akphi:
@mrudula-gs Sorry, took the liberty myself to update the description and the changeset for a clearer explanation since this problem is quite confusing. I would also note that this is just the temporary solution on Studio side for now until we have a more definitive strategy for https://github.com/finos/legend-engine/issues/429 as well as when @gs-gunjan and @MauricioUyaguari refactored the mapping test flow to use `engine` mapping test runner, we would need to re-evaluate this again.

akphi (approved):

PR #642, opened 2021-11-03
studio: rework handling of multiple SDLC instances

### LINKED ISSUE #149: RFC: Rework UX and workflow for handling multiple SDLC servers
### Overview

In the past, we need to do a quick pass (#147) to allow Studio to work with multiple instances of SDLC servers. However, that quick attempt causes many loose ends. For example, right now whenever the user launches Studio, regardless of the URL route, they are being asked which SDLC server to choose in case where multiple servers are available in the config. This is inconvenient. Also, with the naive implementation we had, we force all sub-routes of Studio to be vocal about which SDLC instances it uses. This seriously limit the extensibility of our app in that conceptually, we cannot have any part of Studio that can work without caring too much about SDLC (see #617 for example)

### Implementation plan

- [x] Rework the approach done in https://github.com/finos/legend-studio/pull/171 as it disallows us to have non-SDLC base URL pattern (i.e. part of Studio that might not need SDLC at all)
  - [x] Refactor to accept only 2 types of SDLC-instance patterns: `studio/-/...` and `studio/sdlc-:sdlcServerKey/...`
  - [x] Remove the decorator around `extraApplicationPageRenderEntries` URL patterns
  - [x] Remove `_currentSdlcServerOption` in `StudioConfig`
- [x] `MAINTENANCE` Be consistent and rename everything to `SDLC`
- [x] Since we have the `default` SDLC option, we can remove the `isConfigured` flag in `StudioConfig` and the configuration screen altogether
- [x] For SDLC setup, always use the default so we don't block the app
- [x] In `LegendStudioApplication`, separate the concern:
  - [x] We will scan the URL and predict the `SDLCServerOption` to pick: update the current scanner to distinguish between 3 cases:
    - [x] When the URL matches SDLC-instance pattern: and the key is found -> standard flow
    - [x] When the URL matches SDLC-instance pattern: and the key is **not** found -> redirect to use `defaultSdlcServerOption`
    - [x] When the URL **does not** match SDLC-instance pattern: pass-through and just use the default SDLC, i.e. `/-/`
  - [x] Update `<Redirect>` in `<Router>` of `LegendStudioApplicationRoot` to use `defaultSdlcServerOption`
  - [x] Show the `SDLC` selection on the top right (like `Location` for shopping pages) - replace the URL with the new key and reload the page
- [x] Adjust `/-/view/gav/:gav` url pattern to `/view/:gav` - see the work that has been done in https://github.com/finos/legend-studio/pull/639 for https://github.com/finos/legend-studio/issues/638

### Additional context

_No response_

akphi:
## Summary

- [x] Closes #149 

Added/Updated integration tests

Re. the new SDLC dropdown

https://user-images.githubusercontent.com/13574879/140401687-e67a0b79-a6b6-403c-a363-af4c17ef1b68.mov


MauricioUyaguari:
@akphi  can we add a screenshot of the dropdown config :)

akphi:
Of course, it's only fair :) Updated

packages/legend-studio/src/components/editor/side-bar/Explorer.tsx
```diff
@@ -222,7 +222,7 @@ const ExplorerContextMenu = observer(
         applicationStore.navigator.openNewWindow(
           applicationStore.navigator.generateLocation(
             generateViewEntityRoute(
-              applicationStore.config.sdlcServerKey,
+              applicationStore.config.currentSDLCServerOption,
```
MauricioUyaguari, line 225: Does this account for when you are viewing the explorer tree in gav mode. Seems like we are generating the viewr mode link with the default sdlc server option which is not necessarily right

packages/legend-studio/src/components/editor/side-bar/Explorer.tsx
```diff
@@ -236,7 +236,7 @@ const ExplorerContextMenu = observer(
           .copyTextToClipboard(
             applicationStore.navigator.generateLocation(
               generateViewEntityRoute(
-                applicationStore.config.sdlcServerKey,
+                applicationStore.config.currentSDLCServerOption,
```
MauricioUyaguari, line 239: Please fix the generation of the link copied

MauricioUyaguari (approved):

PR #663, opened 2021-11-12
Fix bug in filegeneration editor not resetting while changing elements

### LINKED ISSUE #142: Bug: File Generation Editor does not reset when changing elements
### Step to Reproduce

Create two classes (Class A, Class B). 
Open the generated view for Class A to avro. 
Open the generated view for Class B to avro. 
File generation editor for class b will not auto open the avro generated file (it will for class A). 

### Expected behavior

When opening the generated view for Class B, the file generation editor should open the first avro generated file.

gayathrir11:
## Summary
Closes #142 

https://user-images.githubusercontent.com/87973126/141459981-c43317f7-4221-456c-84f1-503e12df3b4a.mp4

### DIFF
--- a/.changeset/clever-seas-study.md
+++ b/.changeset/clever-seas-study.md
@@ -1,0 +1,5 @@
+---
+"@finos/legend-studio": patch
+---
+
+Fix bug in `FileGenerationEditor` not resetting while changing elements ([#142](https://github.com/finos/legend-studio/issues/142)]).
--- a/packages/legend-studio/src/stores/editor-state/FileGenerationState.ts
+++ b/packages/legend-studio/src/stores/editor-state/FileGenerationState.ts
@@ -180,11 +180,10 @@ export class FileGenerationState {
     reprocessOpenNodes(treeData, this.filesIndex, this.root, openedNodeIds);
     // select the current file node if available, else select the first output
     const selectedFileNodePath =
-      this.selectedNode?.fileNode.path ??
-      (generationResult.length === 1
+      generationResult.length === 1 || this.selectedNode === undefined
         ? (generationResult[0] as GenerationOutputResult).generationOutput
             .fileName
-        : undefined);
+        : this.selectedNode.fileNode.path;
     if (selectedFileNodePath) {
       const file = this.filesIndex.get(selectedFileNodePath);
       if (file) {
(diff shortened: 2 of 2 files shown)


.changeset/clever-seas-study.md
```diff
@@ -0,0 +1,5 @@
+---
+"@finos/legend-studio": patch
+---
+
+Fix bug in filegeneration editor not resetting while changing elements
```
MauricioUyaguari, line 5:
````suggestion
Fix bug in `FileGenerationEditor` not resetting while changing elements (see [#142](https://github.com/finos/legend-studio/issues/142)] for more details).
````

File: packages/legend-studio/src/stores/editor-state/FileGenerationState.ts [hunk 2]
```diff
@@ -180,11 +180,10 @@ export class FileGenerationState {
     reprocessOpenNodes(treeData, this.filesIndex, this.root, openedNodeIds);
     // select the current file node if available, else select the first output
     const selectedFileNodePath =
-      this.selectedNode?.fileNode.path ??
-      (generationResult.length === 1
+      generationResult.length === 1
```
MauricioUyaguari, line 183: This is moving in the right direction but we are not reopening the same file in a regenerate that the user had opened instead we are always opening the first element in the generation output. 
I think we should follow the below logic. 
if `this.selectedNode?.fileNode.path` exists and the corresponding file exists in the file index open that. 
If not open the first generationOutput if there is one.

MauricioUyaguari:
I think we are close with this change, however we don't handle processing of file generations with no generation results. 
Use the grammar below. If you open the file generation editor, you will see an error. So when there is no selectedNode we should only set the selectedNode to the first generation result if there are results. 
```
###Text
Text model::A
{
  type: plainText;
  content: '';
}


###FileGeneration
Avro model::MyFilegeneration
{
  scopeElements: [model];
}

```

MauricioUyaguari:
Appreciate the bug fix @gayathrir11 . Could we check one more thing? Let's investigate how we can add for tests for this. Im okay with filing a separate issue for tests if you think it will be too involved and leave it for another commit.

MauricioUyaguari (commented):
Left some comments!

gayathrir11:
> Appreciate the bug fix @gayathrir11 . Could we check one more thing? Let's investigate how we can add for tests for this. Im okay with filing a separate issue for tests if you think it will be too involved and leave it for another commit.

I think we should leave it for another commit

MauricioUyaguari (approved):

Issue #660, opened 2021-11-11
Feature request: Properly handle duplicated attributes (in enumerations, classes, associations, profiles...)


akphi:
### Similar requests

- [X] I have searched and found no existing similar requests

### How are you using Studio?

Legend Studio

### What problems are you trying to solve?

Right now, `Studio` allows users to make duplication mistakes without any sort of warnings because we rely on `engine` to throw compilation errors, however, this does not seem to be the case (but that's a separate issue that we would file on `engine` side). Studio should do some validation when building the graph as well as giving users feedback when they are working in form mode.

### Describe the solution you would like to see implemented

- [ ] Fail graph builder (add tests for this) and show warnings in form mode when encounter:
  - [ ] Class properties with the same name
  - [ ] Enumeration with same values
  - [ ] Association properties with the same name
  - [ ] Profile tags with the same value
  - [ ] Profile stereotypes with the same value

> Note that for the case of class properties, due to class inheritance: this analysis might not be trivial as we have to account for overriding properties - which I suppose we support right now in `Pure`, so perhaps, Studio will only do a dumb check for duplicated properties in the list of a class's own properties

### Describe alternatives you have considered

We should also do this validation in `engine`, we thought this is enough, but UX-wise, this isn't, users should get early feedback

### Contribution

- [ ] I would like to work on this feature

akphi:
I think we can close this as https://github.com/finos/legend-studio/issues/941 is closed with an adequate approach, @gayathrir11's change in  #680 has now been integrated to `strict-mode` introduced in #1540

PR #680, opened 2021-11-15
Properly handle duplicated attributes

gayathrir11:

- [ ] Test(s) added

### DIFF
--- a/packages/legend-studio/src/components/editor/edit-panel/uml-editor/AssociationEditor.tsx
+++ b/packages/legend-studio/src/components/editor/edit-panel/uml-editor/AssociationEditor.tsx
@@ -76,13 +76,22 @@ const AssociationPropertyBasicEditor = observer(
     association: Association;
     property: Property;
     selectProperty: () => void;
+    checkDuplicateValue: (val: string) => void;
     isReadOnly: boolean;
   }) => {
-    const { association, property, selectProperty, isReadOnly } = props;
+    const {
+      association,
+      property,
+      selectProperty,
+      checkDuplicateValue,
+      isReadOnly,
+    } = props;
     const editorStore = useEditorStore();
     // Name
-    const changeValue: React.ChangeEventHandler<HTMLInputElement> = (event) =>
+    const changeValue: React.ChangeEventHandler<HTMLInputElement> = (event) => {
+      checkDuplicateValue(event.target.value);
       property.setName(event.target.value);
+    };
     // Generic Type
     const [isEditingType, setIsEditingType] = useState(false);
     // FIXME: make this so that association can only refer to classes from the same space
@@ -342,6 +351,16 @@ export const AssociationEditor = observer(
         association.deleteTaggedValue(val);
     // Property
     const deselectProperty = (): void => setSelectedProperty(undefined);
+    const checkDuplicateValue = (val: string): void => {
+      if (
+        association.properties[0].name === val ||
+        association.properties[1].name === val
+      ) {
+        editorStore.applicationStore.notifyWarning(
+          `Duplicated property '${val}' in association '${association.path}'`,
+        );
+      }
+    };
     // Drag and Drop
     const handleDropTaggedValue = useCallback(
       (item: UMLEditorElementDropTarget): void => {
@@ -451,6 +470,7 @@ export const AssociationEditor = observer(
                       selectProperty={selectProperty(
                         association.getFirstProperty(),
                       )}
+                      checkDuplicateValue={checkDuplicateValue}
                       isReadOnly={isReadOnly}
                     />
                     <AssociationPropertyBasicEditor
@@ -459,6 +479,7 @@ export const AssociationEditor = observer(
                       selectProperty={selectProperty(
                         association.getSecondProperty(),
                       )}
+                      checkDuplicateValue={checkDuplicateValue}
                       isReadOnly={isReadOnly}
                     />
                   </div>
--- a/packages/legend-studio/src/components/editor/edit-panel/uml-editor/ClassEditor.tsx
+++ b/packages/legend-studio/src/components/editor/edit-panel/uml-editor/ClassEditor.tsx
@@ -89,18 +89,27 @@ const PropertyBasicEditor = observer(
     property: Property;
     selectProperty: () => void;
     deleteProperty: () => void;
+    checkDuplicateProperty: (val: string) => void;
     isReadOnly: boolean;
   }) => {
-    const { property, _class, selectProperty, deleteProperty, isReadOnly } =
-      props;
+    const {
+      property,
+      _class,
+      selectProperty,
+      deleteProperty,
+      checkDuplicateProperty,
+      isReadOnly,
+    } = props;
     const editorStore = useEditorStore();
     const isInheritedProperty =
       property.owner instanceof Class && property.owner !== _class;
     const isPropertyFromAssociation = property.owner instanceof Association;
     const isIndirectProperty = isInheritedProperty || isPropertyFromAssociation;
     // Name
-    const changeValue: React.ChangeEventHandler<HTMLInputElement> = (event) =>
+    const changeValue: React.ChangeEventHandler<HTMLInputElement> = (event) => {
+      checkDuplicateProperty(event.target.value);
       property.setName(event.target.value);
+    };
     // Generic Type
     const [isEditingType, setIsEditingType] = useState(false);
     const propertyTypeOptions = editorStore.classPropertyGenericTypeOptions;
@@ -892,6 +901,16 @@ export const ClassFormEditor = observer(
           setSelectedProperty(undefined);
         }
       };
+    const checkDuplicateProperty = (val: string): void => {
+      if (
+        _class.properties.find((property) => property.name === val) !==
+        undefined
+      ) {
+        editorStore.applicationStore.notifyWarning(
+          `Duplicated property '${val}' in class '${_class.path}'`,
+        );
+      }
+    };
     // Add button
     let addButtonTitle = '';
     switch (selectedTab) {
@@ -1201,6 +1220,7 @@ export const ClassFormEditor = observer(
                           _class={_class}
                           deleteProperty={deleteProperty(property)}
                           selectProperty={selectProperty(property)}
+                          checkDuplicateProperty={checkDuplicateProperty}
                           isReadOnly={isReadOnly}
                         />
                       ))}
--- a/packages/legend-studio/src/components/editor/edit-panel/uml-editor/EnumerationEditor.tsx
+++ b/packages/legend-studio/src/components/editor/edit-panel/uml-editor/EnumerationEditor.tsx
@@ -63,11 +63,15 @@ const EnumBasicEditor = observer(
     _enum: Enum;
     selectValue: () => void;
     deleteValue: () => void;
+    checkDuplicateValue: (val: string) => void;
     isReadOnly: boolean;
   }) => {
-    const { _enum, selectValue, deleteValue, isReadOnly } = props;
-    const changeValue: React.ChangeEventHandler<HTMLInputElement> = (event) =>
+    const { _enum, selectValue, deleteValue, checkDuplicateValue, isReadOnly } =
+      props;
+    const changeValue: React.ChangeEventHandler<HTMLInputElement> = (event) => {
+      checkDuplicateValue(event.target.value);
       _enum.setName(event.target.value);
+    };
 
     return (
       <div className="enum-basic-editor">
@@ -419,6 +423,15 @@ export const EnumerationEditor = observer(
         editorStore.openElement(enumeration.generationParentElement);
       }
     };
+    const checkDuplicateValue = (val: string): void => {
+      if (
+        enumeration.values.find((value) => value.name === val) !== undefined
+      ) {
+        editorStore.applicationStore.notifyWarning(
+          `Duplicated value '${val}' in enumeration '${enumeration.path}'`,
+        );
+      }
+    };
 
     return (
       <div
@@ -493,6 +506,7 @@ export const EnumerationEditor = observer(
                         _enum={enumValue}
                         deleteValue={deleteValue(enumValue)}
                         selectValue={selectValue(enumValue)}
+                        checkDuplicateValue={checkDuplicateValue}
                         isReadOnly={isReadOnly}
                       />
                     ))}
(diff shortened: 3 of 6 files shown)


packages/legend-studio/src/components/editor/edit-panel/uml-editor/AssociationEditor.tsx
```diff
@@ -76,13 +76,22 @@ const AssociationPropertyBasicEditor = observer(
     association: Association;
     property: Property;
     selectProperty: () => void;
+    checkDuplicateValue: (val: string) => void;
```
akphi, line 79: I'm not so convinced that we need to pass this in as a prop like this. We can construct this function right in the body of `AssociationPropertyBasicEditor`. Also, I am thinking about the general strategy that you used, it's not bad to do a warning, but kinda disruptive in terms of UX, imagine, we already have a property called `prop1`, if user is creating `prop12`, while typing, they would get this notification, that's not great. So instead, let's make this input box red, we can discuss the styling later, but you can follow some example we have (see `ServiceEditor` runtime/mapping field for example, we do some validation there).

akphi:
@gayathrir11 General comment here. @MauricioUyaguari made a point that we should not do a `hard-fail` when we build the graph just because we're overstepping engine here, we should let engine handle this first, when they are throwing errors, we can change this to what you have.

So you should print out a warning instead. Also, the added tests are super valuable so let's keep them and temporarily `skip` them. You should add comments where we left these things off like this:

`/* @MARKER: Temporary until we resolve https://github.com/finos/legend-studio/issues/660 */`

Also, please update the description to not close #660 

@mrudula-gs Could you let us know when this validation for engine is committed? Thanks!!

packages/legend-studio/src/components/editor/edit-panel/uml-editor/AssociationEditor.tsx
```diff
@@ -80,9 +84,25 @@ const AssociationPropertyBasicEditor = observer(
...
     const { association, property, selectProperty, isReadOnly } = props;
     const editorStore = useEditorStore();
+    const checkDuplicateValue = (val: string, id: string): void => {
+      if (
+        association.properties[0].name === val ||
+        association.properties[1].name === val
+      ) {
+        guaranteeNonNullable(
+          document.getElementById(id),
+        ).style.backgroundColor = 'red';
```
akphi, line 94: :( no, let's not do this, let's use a boolean and set the style on the element. Did you check how we do that in `ServiceExecutionEditor` for runtime and mapping?

packages/legend-studio/src/components/editor/edit-panel/uml-editor/AssociationEditor.tsx
```diff
@@ -80,9 +84,25 @@ const AssociationPropertyBasicEditor = observer(
...
+      } else {
+        guaranteeNonNullable(
+          document.getElementById(id),
+        ).style.backgroundColor = 'var(--color-dark-grey-300)';
+      }
+    };
     // Name
-    const changeValue: React.ChangeEventHandler<HTMLInputElement> = (event) =>
+    const changeValue: React.ChangeEventHandler<HTMLInputElement> = (event) => {
+      checkDuplicateValue(event.target.value, event.target.id);
```
akphi, line 103: this is not ideal, we want this validation to be automatically done, not reactive. If the user starts with a class that already has duplicated properties, this won't work right?

packages/legend-studio/src/components/editor/edit-panel/uml-editor/AssociationEditor.tsx
```diff
@@ -161,6 +181,7 @@ const AssociationPropertyBasicEditor = observer(
       <div className="property-basic-editor">
         <input
           className="property-basic-editor__name"
+          id={property.uuid}
```
akphi, line 184: why do we need this

gayathrir11: We want only the input attribute with duplicate property to go red so using `id` to differentiate that

gayathrir11: Won't be needing with updated logic

packages/legend-studio/src/components/editor/edit-panel/uml-editor/AssociationEditor.tsx
gayathrir11: If you think the styling of the input box is looking weird in light theme. I have a screenshot of how it looks like without that background color. Do you want me to change the styling to this?
pic2

packages/legend-studio/src/components/editor/edit-panel/uml-editor/AssociationEditor.tsx
```diff
@@ -80,9 +80,19 @@ const AssociationPropertyBasicEditor = observer(
   }) => {
     const { association, property, selectProperty, isReadOnly } = props;
     const editorStore = useEditorStore();
+    const checkDuplicateValue = (val: string): boolean => {
```
akphi, line 83: --> `checkDuplicatedValues`

packages/legend-studio/src/components/editor/edit-panel/uml-editor/ClassEditor.tsx
```diff
@@ -98,9 +98,18 @@ const PropertyBasicEditor = observer(
       property.owner instanceof Class && property.owner !== _class;
     const isPropertyFromAssociation = property.owner instanceof Association;
     const isIndirectProperty = isInheritedProperty || isPropertyFromAssociation;
+    const checkDuplicateProperty = (val: string): boolean => {
```
akphi, line 101: --> `isPropertyDuplicated = (val: Property): ...`

packages/legend-studio/src/components/editor/edit-panel/uml-editor/EnumerationEditor.tsx
```diff
@@ -63,16 +63,23 @@ const EnumBasicEditor = observer(
...
+      props;
+    const changeValue: React.ChangeEventHandler<HTMLInputElement> = (event) => {
       _enum.setName(event.target.value);
+    };
 
     return (
       <div className="enum-basic-editor">
         <input
-          className="enum-basic-editor__name"
+          className={
```
akphi, line 78: --> `isEnumValueDuplicated = (val: Enum): ...`

packages/legend-studio/src/components/editor/edit-panel/uml-editor/ClassEditor.tsx
```diff
@@ -98,9 +98,18 @@ const PropertyBasicEditor = observer(
...
+      let count = 0;
+      _class.properties.forEach((property) => property.name === val && count++);
```
akphi, line 103: also, why don't we do `_class.properties.filter((prop) => prop.name === val.name).length >= 2`? Could we do this for other places?

packages/legend-studio/src/components/editor/edit-panel/uml-editor/EnumerationEditor.tsx
akphi: Also, let's do this for other places

packages/legend-studio/style/components/editor/uml-editor/_class-editor.scss
```diff
@@ -93,6 +93,12 @@
     margin-right: 0.5rem;
     width: calc(100% - 36rem);
     min-width: calc(100% - 36rem);
+
+    &__duplicate {
```
akphi, line 97: this is a BEM modifier, we should do `--duplicated` instead. Read more here http://getbem.com/naming/

akphi: Also, as mentioned before, could we add a little `TimesCircle` icon at the end of `<input>` tag. Also, it's not bad to use `input-group` (See `input.scss`) to have an error message below `<input>`.

akphi (changes requested):

packages/legend-studio/src/components/editor/edit-panel/uml-editor/AssociationEditor.tsx
gayathrir11: I think the value of the function shouldn't be of type `property` because each `property` will have different `uuid` so this function will always return `false`

.changeset/honest-jokes-move.md
```diff
@@ -0,0 +1,6 @@
+---
+"@finos/legend-studio": patch
+"@finos/legend-graph": patch
+---
+
+Properly handle duplicated attributes in `profile`, `enumeration`, `class`, `association`.
```
akphi, line 6: Actually due to the workaround we introduced, I would split this into 2 changesets:

1. `@finos/legend-studio`: Show warning about duplicated attributes in UML editors.
2. `@finos/legend-graph`: Show warning when detecting duplicated attributes in `profile`, `enumeration`, `class`, `association`.

packages/legend-art/src/components/Input.tsx
```diff
@@ -19,12 +19,14 @@ import { TimesCircleIcon } from './Icon';
 
 const InputGroup = styled.div`
   position: relative;
+  width: 100%;
```
akphi, line 22: have you tested how this affect query builder projection panel?

packages/legend-studio/src/components/editor/edit-panel/uml-editor/AssociationEditor.tsx
```diff
@@ -159,14 +170,19 @@ const AssociationPropertyBasicEditor = observer(
...
       <div className="property-basic-editor">
-        <input
-          className="property-basic-editor__name"
-          disabled={isReadOnly}
-          value={property.name}
-          spellCheck={false}
-          onChange={changeValue}
-          placeholder={`Property name`}
-        />
+        <div className="input-group__input property-basic-editor__input">
```
akphi, line 173: why do we still need this `wrapper` `<div>`?

packages/legend-studio/src/components/editor/edit-panel/uml-editor/ClassEditor.tsx
```diff
@@ -179,15 +184,22 @@ const PropertyBasicEditor = observer(
...
-          <input
-            className="property-basic-editor__name"
-            disabled={isReadOnly}
-            value={property.name}
-            spellCheck={false}
-            onChange={changeValue}
-            placeholder={`Property name`}
-            name={`Property name`}
-          />
+          <div className="input-group__input property-basic-editor__input">
```
akphi, line 187: why do we still need this `wrapper` `<div>`? - please check other palces as well. Thanks!!

packages/legend-studio/src/components/editor/edit-panel/uml-editor/ProfileEditor.tsx
```diff
@@ -64,22 +73,30 @@ const StereotypeBasicEditor = observer(
   (props: {
     stereotype: Stereotype;
     deleteStereotype: () => void;
+    isStereotypeDuplicated: (val: Stereotype) => boolean;
```
akphi, line 76: why do we need to pass this when we can derived this via `stereotype.profile`?

packages/legend-studio/src/components/editor/edit-panel/uml-editor/ProfileEditor.tsx
```diff
@@ -21,28 +21,37 @@ import {
...
 import { Tag, Stereotype } from '@finos/legend-graph';
 
 const TagBasicEditor = observer(
-  (props: { tag: Tag; deleteValue: () => void; isReadOnly: boolean }) => {
-    const { tag, deleteValue, isReadOnly } = props;
-    const changeValue: React.ChangeEventHandler<HTMLInputElement> = (event) =>
+  (props: {
+    tag: Tag;
+    deleteValue: () => void;
+    isTagDuplicated: (val: Tag) => boolean;
```
akphi, line 34: why do we need to pass this when we can derived this via `stereotype.profile`?

packages/legend-studio/src/components/editor/edit-panel/uml-editor/EnumerationEditor.tsx
```diff
@@ -63,22 +64,33 @@ const EnumBasicEditor = observer(
     _enum: Enum;
     selectValue: () => void;
     deleteValue: () => void;
+    isEnumValueDuplicated: (val: Enum) => boolean;
```
akphi, line 67: why do we need to pass this when we can derived this via `_enum.owner`?

packages/legend-studio/style/components/editor/uml-editor/_class-editor.scss
```diff
@@ -95,6 +95,22 @@
     min-width: calc(100% - 36rem);
   }
 
+  &__name__duplicated {
```
akphi, line 98: why is this class named so when it's applied to the component even when there is no error?

packages/legend-art/src/components/Input.tsx
gayathrir11: yes. It was looking the same

packages/legend-studio/src/components/editor/edit-panel/uml-editor/AssociationEditor.tsx
gayathrir11: Oh that is have a margin to the right of the input box. Same at other places too. Could have added it to the `Input.tsx` but I think that will mess the styling at other places

packages/legend-studio/style/components/editor/uml-editor/_class-editor.scss
```diff
@@ -95,6 +95,22 @@
     min-width: calc(100% - 36rem);
   }
 
+  &__input__withValidation {
```
akphi, line 98: per `BEM` this should be `__input--with-validation`.

packages/legend-studio/src/components/editor/edit-panel/uml-editor/AssociationEditor.tsx
akphi: then we should add the style to the `class` you set for the `<input>` right?

packages/legend-studio/style/components/editor/uml-editor/_class-editor.scss
```diff
@@ -95,6 +95,22 @@
...
+  &__input--with-validation {
+    @include flexVCenter;
+
+    flex: auto;
+    height: 2.8rem;
+    background: var(--color-dark-grey-300);
+    color: var(--color-light-grey-100);
+    padding: 0 1rem;
+    border-radius: 0.2rem;
+    margin-right: 0.5rem;
```
akphi, line 107:
```suggestion
```

PR #729, opened 2021-12-08
Fix processing of bitemporal milestoning stereotype

gayathrir11:

- [ ] Test(s) added

### DIFF
--- a/packages/legend-query/src/components/QueryBuilderMilestoneEditor.tsx
+++ b/packages/legend-query/src/components/QueryBuilderMilestoneEditor.tsx
@@ -150,25 +150,23 @@ const MilestoningParameterEditor = observer(
           <div className="query-builder__parameter-editor__parameter">
             {milestoningParameter.genericType.value.rawType.name ===
               PRIMITIVE_TYPE.STRICTDATE && (
-              <div className="query-builder__parameter-editor__parameter">
-                <DatePrimitiveInstanceValueEditor
-                  valueSpecification={milestoningParameter}
-                />
-              </div>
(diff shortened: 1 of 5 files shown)


packages/legend-query/src/components/QueryBuilderMilestoneEditor.tsx
```diff
@@ -206,20 +204,20 @@ const BiTemporalMilestoneEditor = observer(
...
           />
         </div>
         <div className="panel__content__form__section">
           <div className="panel__content__form__section__header__label">
-            Processing Date
+            Business Date
           </div>
           <MilestoningParameterEditor
-            key="ProcessingDate"
+            key="BusinessDate"
```
akphi, line 220: Sorry if I asked before, but why do we need to have this prop? We're not using a `list` anywhere

packages/legend-query/src/components/QueryBuilderMilestoneEditor.tsx
```diff
@@ -206,20 +204,20 @@ const BiTemporalMilestoneEditor = observer(
...
             queryBuilderState={queryBuilderState}
             index={1}
```
akphi, line 222: Could we spend a little bit of time to rename this to `parameterIndex` or something? `index` is kinda vague

packages/legend-query/src/components/QueryBuilderMilestoneEditor.tsx
gayathrir11: We are using `MilestoningParameterEditor` twice for a `bitemporal` class, so adding a key

Issue #683, opened 2021-11-15
Feature request:  Handle processing of milestoning classes 


YannanGao-gs:
### Similar issues

- [X] I have searched and found no existing similar issues

### How are you using Studio?

Legend Studio

### Current and expected behavior

The model below should work well in the Studio

### Steps to reproduce

use the below model

### Model data

```text
###Relational
Database my::db
(
  Table PersonTable
  (
    NAME CHAR(200),
    FIRMID INTEGER
  )
  Table FirmTable
  (
    FIRMNAME CHAR(200),
    ID INTEGER
  )

  Join Firm_Person(PersonTable.FIRMID = FirmTable.ID)
)


###Pure
Class <<temporal.businesstemporal>> my::Person
{
  name: String[1];
  firmID: Integer[1];
}

Class my::Firm
{
  legalName: String[1];
  id: Integer[1];
}

Association my::Employment
{
  firm: my::Firm[0..1];
  employees: my::Person[*];
}


###Mapping
Mapping my::map
(
  my::Person[per1]: Relational
  {
    ~mainTable [my::db]PersonTable
    name: [my::db]PersonTable.NAME
  }
  my::Firm[fir1]: Relational
  {
    ~mainTable [my::db]FirmTable
    legalName: [my::db]FirmTable.FIRMNAME
  }

  my::Employment: Relational
  {
    AssociationMapping
    (
      employeesAllVersions[fir1,per1]: [my::db]@Firm_Person,
      firm[per1,fir1]: [my::db]@Firm_Person
    )
  }
)
```

### Environment

```text
System:
    OS: macOS 12.0.1
  Binaries:
    Node: 16.10.0 - /private/var/folders/7l/8nlbhlj55c9_lp8c23_61n8h0000gn/T/xfs-ac754eb3/node
    Yarn: 3.0.1 - /private/var/folders/7l/8nlbhlj55c9_lp8c23_61n8h0000gn/T/xfs-ac754eb3/yarn
    npm: 7.24.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 95.0.4638.69
```

### Contribution

- [X] I would like to work on the fix for this issue

MauricioUyaguari:
Moving this to backlog as Mapping defined above is incorrect. Users should not be allowed to map the `allVersion` property in a mapping. 
If there is a requirement for studio to generate milestoning properties we will look to implement this.

akphi:
[...]

PR #766, opened 2022-01-05
Support date propagation for milestoning properties in QueryBuilder

gayathrir11:

- [ ] Test(s) added


packages/legend-graph/src/helpers/DomainHelper.ts
```diff
@@ -49,3 +59,270 @@ export const getMilestoneTemporalStereotype = (
...
+              _type: 'var',
+              class: 'Date',
+              name: DEFAULT_MILESTONING_PARAMETERS.BUSINESS_DATE,
+              multiplicity: {
+                lowerBound: 1,
+                upperBound: 1,
+              },
+            },
+          ];
+          dateProperty.body = [
```
MauricioUyaguari, line 96: is the body required? do we gain anything from filling this in ? If not let's leave it empty. Let's followt his for the other ones.

packages/legend-graph/src/helpers/DomainHelper.ts
```diff
@@ -49,3 +59,270 @@ export const getMilestoneTemporalStereotype = (
...
+            `${property.name}`,
+            graph.getTypicalMultiplicity(TYPICAL_MULTIPLICITY_TYPE.ONE),
+            GenericTypeExplicitReference.create(
+              new GenericType(
+                graph.getClass(property.genericType.value.rawType.path),
+              ),
+            ),
+            property.owner,
+          );
+          dateProperty.parameters = [
```
MauricioUyaguari, line 85: Instead of doing this could we build `VariableExpression` then serialize to JSON ? its a lot cleaner. lets do this for the rest as well.

packages/legend-graph/src/helpers/DomainHelper.ts
```diff
@@ -49,3 +59,270 @@ export const getMilestoneTemporalStereotype = (
...
+              ],
+              property: `${property.name}`,
+            },
+          ];
+          const milestonedAllVersions = new Property(
+            `${property.name}AllVersions`,
+            graph.getTypicalMultiplicity(TYPICAL_MULTIPLICITY_TYPE.ONEMANY),
+            GenericTypeExplicitReference.create(
+              new GenericType(
+                graph.getClass(property.genericType.value.rawType.path),
```
MauricioUyaguari, line 113: same comment we probably don't need the `getClass`

packages/legend-graph/src/helpers/DomainHelper.ts
```diff
@@ -49,3 +59,270 @@ export const getMilestoneTemporalStereotype = (
...
+              ),
+            ),
+            property.owner,
+          );
+          const milestonedAllVersionsInRange = new DerivedProperty(
+            `${property.name}AllVersionsInRange`,
+            graph.getTypicalMultiplicity(TYPICAL_MULTIPLICITY_TYPE.ONE),
```
MauricioUyaguari, line 120: should this also be `ONEMANY` MULTIPLICITY ?

packages/legend-graph/src/models/metamodels/pure/packageableElements/domain/Association.ts
```diff
@@ -79,6 +79,14 @@ export class Association
   taggedValues: TaggedValue[] = [];
   derivedProperties: DerivedProperty[] = [];
 
+  //To store the properties generated while processing the milestoning properties. The properties
+  //generated are `allVersions`, `allVersionsInRange` and derived property with date parameter.
+  //Engine does the processing differently by adding the derived properties generated to the list
+  // of qualified properties and the actual milestoning properties to the list of _originalMilestonedProperties
+  // and the properties to the list of actual proeprties. We are not doing that here to easy the process of
+  //handling milestoning properties in query builder.
```
MauricioUyaguari, line 87: Lets follow the format of other comments and lets reword this `We are not doing that here to easy the process of
  //handling milestoning properties in query builder.`

packages/legend-graph/src/helpers/DomainHelper.ts
```diff
@@ -49,3 +59,270 @@ export const getMilestoneTemporalStereotype = (
...
+            milestonedAllVersions,
+          );
+          break;
+        }
+        default:
+          break;
+      }
+    }
+  });
+};
```
MauricioUyaguari, line 328: Could we add a test case in `GraphBuildSuccess.test` testing the graph. Specifically testing the milestoned properties are generated.

MauricioUyaguari (changes requested):
Have reviewed `legend-graph`. Willl move onto the query part now.

packages/legend-graph/src/helpers/DomainHelper.ts
gayathrir11: Just realized that multiplicity of all these generated properties is dependent on the actual milestoning property. Will fix that

packages/legend-query/src/stores/QueryBuilderPropertyEditorState.ts
```diff
@@ -275,6 +416,18 @@ export class QueryBuilderPropertyExpressionState {
       ) {
         requiresExistsHandling = true;
       }
+      if (
+        currentExpression.func.genericType.value.rawType instanceof Class &&
+        currentExpression.func.owner._generatedMilestonedProperties.length !== 0
```
MauricioUyaguari, line 421: Is it possible for the `_generatedMilestonedProperties` to be populated but the currentExpression not be a milestone expression? we should make sure its there in the list before replacing the current expression.func.

MauricioUyaguari (changes requested):

packages/legend-query/src/stores/QueryBuilderPropertyEditorState.ts
```diff
@@ -88,55 +92,187 @@ const fillDerivedPropertyArguments = (
...
+          .rawType,
+        derivedPropertyExpressionState.queryBuilderState.graphManagerState
+          .graph,
+      );
+    }
+    if (temporalTarget) {
+      switch (temporalTarget) {
+        case MILESTONING_STEROTYPES.BUSINESS_TEMPORAL: {
+          let parameter;
+          if (querySetupState.classMilestoningTemporalValues.length === 0) {
```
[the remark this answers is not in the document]
gayathrir11: i thought of doing the same initially but that's not how we are doing it internally. Can give you a quick demo on that in the morning

packages/legend-query/src/stores/QueryBuilderPropertyEditorState.ts
```diff
@@ -88,55 +92,187 @@ const fillDerivedPropertyArguments = (
...
+              DEFAULT_MILESTONING_PARAMETERS.PROCESSING_DATE,
+            );
+            derivedPropertyExpressionState.queryBuilderState.buildMilestoningParameter(
+              DEFAULT_MILESTONING_PARAMETERS.BUSINESS_DATE,
+            );
+          } else if (
+            querySetupState.classMilestoningTemporalValues.length === 1
+          ) {
+            if (
+              querySetupState.classMilestoningTemporalValues[0] instanceof
```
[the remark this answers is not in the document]
gayathrir11: actually this scenario arises only when parent is not `bitemporal`. If parent is already `bitemporal` `classMilestoningTemporalValues.length` will be 2 and it will just populate the default values.

packages/legend-query/src/stores/QueryBuilderLambdaBuilder.ts
```diff
@@ -101,21 +102,55 @@ export const buildLambdaFunction = (
...
+        let parameter;
+        if (
+          queryBuilderState.querySetupState.classMilestoningTemporalValues
+            .length === 1
+        ) {
+          parameter =
+            queryBuilderState.querySetupState.classMilestoningTemporalValues[0];
+        } else {
+          parameter =
+            queryBuilderState.querySetupState.classMilestoningTemporalValues[1];
```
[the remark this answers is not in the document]
gayathrir11: discussed offline

packages/legend-query/src/stores/QueryBuilderPropertyEditorState.ts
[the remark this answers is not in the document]
gayathrir11: discussed offline

MauricioUyaguari:
lets remove  #683. we dont want this unless its necessary for the query builder.

MauricioUyaguari:
I think we have an incorrect behavior based on the second video. 
When a user is editing a parameter, they can change the **TYPE** of the value specification, not when they are executing the query and selecting a value for the parameter. 

So if they select parameter of type `DATE`
when they go to execute, they can switch between types of dates because `STRICTDATE`, `LATEST` etc. are types of DATE. 

However if they choose the parameter to be type `STRICTDATE`, then they are locked on the execute parameter value setter to `STRICTDATE`. They can't change it to another type like `LATEST`. If they want to change it they must go to the parameter editor to change the type there to either `LATEST` of more less strict type like `DATE`

MauricioUyaguari:
For the second video can you include editing processingDate which is of type StrictDate

packages/legend-query/src/components/QueryBuilderValueSpecificationEditor.tsx
```diff
@@ -487,6 +495,76 @@ export const LatestDatePrimitiveInstanceValueEditor: React.FC = () => (
...
+            ),
+            DATE_FORMAT,
+          ),
+        ];
+      }
+    };
+
+    return (
+      <div className="query-builder-value-spec-editor__date">
+        {(valueSpecification.genericType.value.rawType.name ===
```
MauricioUyaguari, line 539: could we instead of comparing names compares the types themselves. You can grab
````suggestion
const strictDtae = graph.getPrimitiveType(PRIMITETYPE.STRICTDATE);
rawType ===strictDate
```

MauricioUyaguari (changes requested):
small change request + video update, then good to go

.changeset/spicy-trains-sit.md
```diff
@@ -0,0 +1,5 @@
+---
+"@finos/legend-query": patch
```
akphi, line 2: --> `minor`

MauricioUyaguari (approved):

PR #768, opened 2022-01-06
Add default error message when received empty error message for compilation failure

### LINKED ISSUE #708: Bug: Studio fails to navigate to error location in text mode if error does not contain error messgae
### Similar issues

### How are you using Studio?

Legend Studio

Use Grammar below in text mode. There is a compilation error without an error message. The error message contains source information. Because there is no error message Studio fails to navigate user to location of error message.

Although engine should return an error message, Studio should navigate users to where the error occurred.
Corresponding Engine Issue - https://github.com/finos/legend-engine/issues/455

Engine Response to Studio
```
{
  "errorType": "COMPILATION",
  "sourceInformation": {
    "endColumn": 156,
    "endLine": 5,
    "sourceId": "",
    "startColumn": 3,
    "startLine": 5
  },
  "stackTrace": [],
  "suppressed": []
}

```

### Steps to reproduce

Compile code below in text mode. 

### Model data

```text
Class model::Person
{
  firstName: String[1];
  lastName: String[1];
  isEqual(fullName: String[1], noCheck: Boolean[1]) {if($noCheck, |true, |if(($this.firstName + $this.lastName) == $fullNameL, |true, |false))}: Boolean[1];
}
```

### Environment

```text
Windows
```

_No response_

### Contribution

gayathrir11:
## Summary

Closes #708 

- [ ] Test(s) added

pic3

### DIFF
--- a/.changeset/slow-kings-yell.md
+++ b/.changeset/slow-kings-yell.md
@@ -1,0 +1,7 @@
+---
+"@finos/legend-studio": patch
+"@finos/legend-query": patch
+"@finos/legend-application": patch
+---
+
+Handle empty error messages from Engine.
--- a/packages/legend-application/src/components/LambdaEditor.tsx
+++ b/packages/legend-application/src/components/LambdaEditor.tsx
@@ -327,7 +327,7 @@ const LambdaEditorInline = observer(
         if (error?.sourceInformation) {
           setErrorMarkers(
             editorModel,
-            error.message,
+            error.message === '' ? '(no error message)' : error.message,
             error.sourceInformation.startLine,
             error.sourceInformation.startColumn,
             error.sourceInformation.endLine,
--- a/packages/legend-studio/src/components/editor/edit-panel/GrammarTextEditor.tsx
+++ b/packages/legend-studio/src/components/editor/edit-panel/GrammarTextEditor.tsx
@@ -199,7 +199,7 @@ export const GrammarTextEditor = observer(() => {
       if (error?.sourceInformation) {
         setErrorMarkers(
           editorModel,
-          error.message,
+          error.message === '' ? '(no error message)' : error.message,
           error.sourceInformation.startLine,
           error.sourceInformation.startColumn,
           error.sourceInformation.endLine,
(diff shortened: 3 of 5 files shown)


packages/legend-application/src/components/LambdaEditor.tsx
```diff
@@ -325,6 +325,9 @@ const LambdaEditorInline = observer(
         editorModel.updateOptions({ tabSize: TAB_SIZE });
         const error = parserError ?? compilationError;
         if (error?.sourceInformation) {
+          if (error.message === '') {
```
MauricioUyaguari, line 328: Rethinking this I dont like us replacing the orginal value of the error. Could we may add a getter property to `EngineError` called nonEmptyMessage where we return either the orginal message if not empty and if empty `No error message provided by engine`.

.changeset/slow-kings-yell.md
```diff
@@ -0,0 +1,5 @@
+---
+"@finos/legend-studio": patch
+"@finos/legend-query": patch
+"@finos/legend-application": patch
+---
```
MauricioUyaguari, line 5: Handle empty error messages from Engine.

packages/legend-studio/src/components/editor/edit-panel/GrammarTextEditor.tsx
```diff
@@ -199,7 +199,7 @@ export const GrammarTextEditor = observer(() => {
       if (error?.sourceInformation) {
         setErrorMarkers(
           editorModel,
-          error.message,
+          error.message === '' ? error.nonEmptyMessage : error.message,
```
akphi, line 202: I think we don't need this, let's just hard code the error here `(no error message)`

gayathrir11: will update it.

packages/legend-graph/src/graphManager/action/EngineError.ts
```diff
@@ -34,6 +34,10 @@ export class EngineError extends ApplicationError {
   setSourceInformation(sourceInformation: SourceInformation | undefined): void {
     this.sourceInformation = sourceInformation;
   }
+
+  get nonEmptyMessage(): string {
```
akphi, line 38: don't think we need this, let's just hard code the error in various places with the message `(no error message)`

gayathrir11: will remove this

akphi (approved):
@MauricioUyaguari and I think the best way is to handle things like this at `ApplicationError` level, but we will take of it. Thanks a lot for doing this!

Issue #630, opened 2021-10-30
Feature request: Handle subtypes in query builder


akphi:
### Similar requests

- [X] I have searched and found no existing similar requests

### What problems are you trying to solve?

In query builder, in the explorer tree, we should be able to navigate properties coming from sub-types.

### Describe the solution you would like to see implemented

- [x] Support sub-typing for graph-fetch (with test) - #789 
- [x] Support sub-typing for projection (with test) - #740 
- [x] Change the current way we handle subtypes. If there's `Firm2 extends Firm1 extends Firm` we will form a tree with nested levels according to supertype hierarchy. We need to make the subtype a sub-node of the node that represents the supertype. For example - #741 
```
Firm (+)
  |__ employees: Person (+)
  |       |__ name: String
  |       |__ ProductManager (+)
  |       |       |__ product
  |       |__ Technologist (+)
  |       |       |__ techID
  |       |__ Studio Technologist (+)
  |               |__ githubID
  |__ legalName: String
  |
  |__ @Startup
          |__ fundingRound: FundingRound

Where we have:
Startup extends Firm
ProductManager extends Person
Technologist extends Person
StudioTechnologist extends Technologist
```
- [x] DnD property with subtype to compute the lambda with `->subType...` - #741 
- [x] `DISCUSS` Reconsider the mapped-ness algorithm now that we account for subtype - #812 
- [x] `DISCUSS` Reconsider how this impact the rendering of `property expression editor`
- [ ] For completeness sake, we need to support the syntax `->subType(...).someSupertypeProperty`, but let's leave this out of the explorer tree for now. We should add a test for this as well. - #812 

Example of subtypes

```
test::Person.all()->project(
  [
    x|$x->subType(
    @test::Employee
  ).age,
    x|$x->subType(
    @test::Employee
  ).lastName
  ],
  [
    'Age',
    'Last Name'
  ]
)
```

### Describe alternatives you have considered

N/A

### Contribution

- [ ] I would like to work on this feature

YannanGao-gs:
Demo video:

https://user-images.githubusercontent.com/73408381/153242544-c5dcdbf6-5c8e-4b89-a06a-63ef4cc6bdd4.mov

YannanGao-gs:
The template model used in the video above:

```
###Service
Service model::MyService
{
  pattern: '/d2c48a9c-70fa-46e3-8173-c355e774004f';
  documentation: '';
  autoActivateUpdates: true;
  execution: Single
  {
[...]

PR #812, opened 2022-01-24
Display subclasses in the explorer tree only if mapped

YannanGao-gs:

- [ ] Test(s) added

https://user-images.githubusercontent.com/73408381/150880851-b9e89803-171b-490b-85c9-558e482dbe3b.mov


packages/legend-query/src/stores/QueryBuilderExplorerState.ts
```diff
@@ -337,6 +337,24 @@ export const getPropertyNodeMappingData = (
   return { mapped: false, skipMappingCheck: false };
 };
 
+export const getSubTypeNodeMappingData = (
+  subclass: Class,
+  parentMappingData: QueryBuilderPropertyMappingData,
+  mapping: Mapping | undefined,
+): QueryBuilderPropertyMappingData => {
+  const rootSetImpl = mapping
+    ? getRootSetImplementation(mapping, subclass)
```
akphi, line 346: I believe we don't just find the root set implementation for the subclass but all of the subclass of that subclasses.

akphi (changes requested):
@MauricioUyaguari  Is it me or it seems like the test doesn't really test for mappedness?

packages/legend-query/src/stores/QueryBuilderExplorerState.ts
MauricioUyaguari: @akphi what do you mean by this comment?

MauricioUyaguari: Rethinking about this and this algo seems to be correct. I think the handling for the `extends` is separate after thinking about it a bit more. `extends` is when you look at stuff mapped to the supertype.
So here we are going with only showing subclasses if they are mapped.

MauricioUyaguari:
> @MauricioUyaguari Is it me or it seems like the test doesn't really test for mappedness?

Correct. @YannanGao-gs  please add tests to `QueryBuilder_ExplorerState.test`

akphi:
@YannanGao-gs since this marks the end of the subtype support #630, let's also add a test for the syntax `->subType(...).someSupertypeProperty`, i.e. when people have a query with projection like that, we need to support that in form mode.

packages/legend-query/src/components/__tests__/QueryBuilder_FetchStructure.test.tsx
```diff
@@ -370,6 +370,10 @@ test(
     const _personClass = mockedQueryStore.graphManagerState.graph.getClass(
       'model::pure::tests::model::simple::Person',
     );
+    const _personExtensionClass =
```
akphi, line 373: So after discussing with @MauricioUyaguari I think mapped-ness algorithm is getting more and more heavy, which demands its own test file. Let's create a new test file and do **unit-testing** just the `mapped-ness algorithm`.

packages/legend-query/src/stores/QueryBuilderExplorerState.ts
akphi: Let's check with @pierredebelen or John Pentelow to see how we should implement this. Let say we have 


```
Class Person {
  name: String[1];
}

Class Teacher extends Person {
  school: String[1];
}

Class Student extends Person {
  class: String[1];
}
```

I see 2 potential approach to determine if property `name` of `Person` is mapped:

1. Find root class mapping for `Person`, if there is none, `name` is considered not mapped
2. In any of the subclasses of `Person` is mapped, if `name` is mapped there even once, it's considered mapped

MauricioUyaguari: @akphi Tried the execution and does not work on the above cases. 
@pierredebelen confirmed that this is would work if you have the extends defined which we already addressed with https://github.com/finos/legend-studio/pull/822. 
@pierredebelen confirmed for the true `mappedness` of a subtype property we would need to go into the logic of reading and understanding operational class mappings. He does not want us to do this as it is complex. Instead we should only not show the property if we are confident it is not mapped. 
I opt to close this PR and leave it as showing all properties under subtype.

YannanGao-gs:
Close this pr based on conversations above.  
https://github.com/finos/legend-studio/issues/834

Issue #780, opened 2022-01-13
Bug: Query builder not showing mapped properties through inheritance mapping


MauricioUyaguari:
### Similar issues

- [X] I have searched and found no existing similar issues

### How are you using Studio?

Legend Query

### Current and expected behavior

When a property is mapped through an inhertiance mapping, the property does not show up in the explorer tree for query builder. The property should show up as it is mapped.

### Steps to reproduce

- Open test on the mapping
- Firm Explorer tree does not show `legalName`. It has been mapped through the supertype LegalEnity

![image](https://user-images.githubusercontent.com/29690824/151485984-84e8ade5-c0fc-48a5-aebb-a4dbe3e86419.png)

### Model data

```text
###Service
Service model::MyService
{
  pattern: '/d2c48a9c-70fa-46e3-8173-c355e774004f';
  documentation: '';
  autoActivateUpdates: true;
  execution: Single
  {
    query: |model::Firm.all()->project([x|$x.employeeSize, x|$x.legalName], ['Employee Size', 'Legal Name']);
    mapping: model::NewMapping;
    runtime: model::Runtime;
  }
  test: Single
  {
    data: '';
    asserts:
    [
    ];
  }
}


###Relational
Database model::Test
(
  Table FirmTable
  (
    id INTEGER PRIMARY KEY,
    Legal_name VARCHAR(200)
  )
  Table PersonTable
  (
    id INTEGER PRIMARY KEY,
    firm_id INTEGER,
    firstName VARCHAR(200),
    lastName VARCHAR(200)
  )

  Join FirmPerson(PersonTable.firm_id = FirmTable.id)
)


###Pure
Profile model::MyExtension
{
  stereotypes: [important];
  tags: [doc];
}

Enum model::IncType
{
  Corp,
  LLC
}

Class model::LegalEntity
{
  legalName: String[1];
}

Class model::Person
{
  firstName: String[1];
  lastName: String[1];
}

Class <<model::MyExtension.important>> {model::MyExtension.doc = 'This is a model of a firm'} model::Firm extends model::LegalEntity
[
  validName: $this.legalName->startsWith('_')
]
{
  employees: model::Person[1..*];
  incType: model::IncType[1];
  employeeSize() {$this.employees->count()}: Number[1];
}


###Mapping
Mapping model::NewMapping
(
  *model::Firm extends [model_LegalEntity]: Relational
  {
    ~primaryKey
    (
      [model::Test]FirmTable.id
    )
[...]

PR #822, opened 2022-01-28
Display properties mapped through inheritance mapping

YannanGao-gs:
## Summary
Fixes https://github.com/finos/legend-studio/issues/780

- [x] Test(s) added

https://user-images.githubusercontent.com/73408381/151884898-c17c6912-3d83-404a-b826-b0ab8ac1f552.mov

### DIFF
--- a/packages/legend-graph/src/helpers/MappingHelper.ts
+++ b/packages/legend-graph/src/helpers/MappingHelper.ts
@@ -21,6 +21,7 @@ import type { Class } from '../models/metamodels/pure/packageableElements/domain
 import type { Enumeration } from '../models/metamodels/pure/packageableElements/domain/Enumeration';
 import type { Mapping } from '../models/metamodels/pure/packageableElements/mapping/Mapping';
 import { AggregationAwareSetImplementation } from '../models/metamodels/pure/packageableElements/mapping/aggregationAware/AggregationAwareSetImplementation';
+import { RootRelationalInstanceSetImplementation } from '../models/metamodels/pure/packageableElements/store/relational/mapping/RootRelationalInstanceSetImplementation';
 
 export const getAllClassMappings = (mapping: Mapping): SetImplementation[] =>
   uniq(
@@ -117,3 +118,25 @@ export const getEnumerationMappingsByEnumeration = (
     (enumerationMapping) =>
       enumerationMapping.enumeration.value === enumeration,
   );
+
+export const getAllSuperSetImplementations = (
+  currentSetImpl: SetImplementation,
+): SetImplementation[] => {
+  if (
+    currentSetImpl instanceof RootRelationalInstanceSetImplementation &&
+    currentSetImpl.superSetImplementationId
+  ) {
+    const superSetImpl = [
+      getClassMappingById(
+        currentSetImpl.parent,
+        currentSetImpl.superSetImplementationId,
+      ),
+    ];
+    const superSetImplFromParents = superSetImpl
+      .map((s) => getAllSuperSetImplementations(s))
+      .flat();
+    return superSetImpl.concat(superSetImplFromParents);
+  } else {
+    return [];
+  }
+};
--- a/packages/legend-query/src/stores/QueryBuilderExplorerState.ts
+++ b/packages/legend-query/src/stores/QueryBuilderExplorerState.ts
@@ -48,6 +48,7 @@ import {
   VARIABLE_REFERENCE_TOKEN,
   ARROW_FUNCTION_TOKEN,
   Multiplicity,
+  getAllSuperSetImplementations,
 } from '@finos/legend-graph';
 import type { QueryBuilderState } from './QueryBuilderState';
 import { action, makeAutoObservable, observable } from 'mobx';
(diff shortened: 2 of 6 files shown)


packages/legend-query/src/stores/__tests__/QueryBuilder_ExplorerState.test.ts
```diff
@@ -109,12 +111,23 @@ const cases: TestCase[] = [
       entities: TEST_DATA__Relational_Inline,
     },
   ],
+  [
+    'Simple relational inheritance mapping using extends',
```
MauricioUyaguari, line 115: Simple relational mapping with inhertiance class mapping

packages/legend-query/src/components/QueryBuilderExplorerPanel.tsx
```diff
@@ -704,6 +705,7 @@ const QueryBuilderExplorerTree = observer(
               queryBuilderState.graphManagerState,
               property,
               node,
+              guaranteeNonNullable(queryBuilderState.querySetupState.mapping),
```
MauricioUyaguari, line 708: Can you make sure we are safe with this NonNullable check. Ideally we don't show a tree if no mapping but lets make sure we don't break the app with this mapping input.

packages/legend-query/src/stores/QueryBuilderExplorerState.ts
```diff
@@ -305,30 +335,35 @@ export const getPropertyNodeMappingData = (
         graphManagerState,
         parentTargetSetImpl,
       );
+      const propertyMappingsFromSuperSetImpls = superSetImplementations
+        .map((s) => resolvePropertyMappingsForSetImpl(graphManagerState, s))
+        .flat();
       const mappedProperties = propertyMappings
```
MauricioUyaguari, line 341: can we have one array of mappedProperties concating superSetMappedProperties

packages/legend-query/src/stores/QueryBuilderExplorerState.ts
```diff
@@ -305,30 +335,35 @@ export const getPropertyNodeMappingData = (
...
-            };
+          let propertyMappingArray;
+          // propertyMappings from the current level is the highest priority
+          if (mappedProperties.includes(property.name)) {
+            propertyMappingArray = propertyMappings;
+          } else {
+            propertyMappingArray = propertyMappingsFromSuperSetImpls;
           }
+          return generatePropertyNodeMappingDataForNonPrimitveProperty(
+            property,
```
MauricioUyaguari, line 364: Lets keep the logic as is and only `resolveTargetSetImplementationForPropertyMapping` if a propertyMapping exists (remove the guaranteNonNullable above please :))

MauricioUyaguari (changes requested):

packages/legend-query/src/stores/QueryBuilderExplorerState.ts
YannanGao-gs: If one property is mapped both in the current class Mapping **(A)** and the class mapping **(B)** which it entends from, I think A takes precedence over B. Therefore I chose not to concatenate them. :)

packages/legend-query/src/stores/QueryBuilderExplorerState.ts
YannanGao-gs: there is if condition above which checks if the mappedProperties or mappedPropertiesFromSuperSetImpls contains property name, ideally if this returns true, a propertyMapping exits correspondingly.

packages/legend-query/src/stores/QueryBuilderExplorerState.ts
MauricioUyaguari: yeah but concat doesn't change this. you add the superset to the end. the find() grabs the first one. if its A it would grab A.

packages/legend-graph/src/helpers/MappingHelper.ts
```diff
@@ -117,3 +118,25 @@ export const getEnumerationMappingsByEnumeration = (
     (enumerationMapping) =>
       enumerationMapping.enumeration.value === enumeration,
   );
+
+export const getAllSuperSetImplementations = (
+  currentSetImpl: SetImplementation | undefined,
```
MauricioUyaguari, line 123: both `currentSetImpl` and `mapping` should be required by this function.

packages/legend-query/src/stores/QueryBuilderExplorerState.ts
```diff
@@ -282,12 +283,33 @@ const isAutoMappedProperty = (
...
+};
+
 export const getPropertyNodeMappingData = (
   graphManagerState: GraphManagerState,
   property: AbstractProperty,
   parentMappingData: QueryBuilderPropertyMappingData,
+  mapping: Mapping | undefined,
 ): QueryBuilderPropertyMappingData => {
   const parentTargetSetImpl = parentMappingData.targetSetImpl;
+  const superSetImplementations = getAllSuperSetImplementations(
```
MauricioUyaguari, line 309: thinking about it why do you need mapping?
can't you use `parentTargetSetImpl.parent` as an input?

packages/legend-graph/src/helpers/MappingHelper.ts
MauricioUyaguari, line 123:
````suggestion
currentSetImpl: SetImplementation;
````

File: packages/legend-query/src/stores/QueryBuilderExplorerState.ts [hunk 7]
```diff
@@ -282,12 +283,30 @@ const isAutoMappedProperty = (
...
+  };
+};
+
 export const getPropertyNodeMappingData = (
   graphManagerState: GraphManagerState,
   property: AbstractProperty,
   parentMappingData: QueryBuilderPropertyMappingData,
 ): QueryBuilderPropertyMappingData => {
   const parentTargetSetImpl = parentMappingData.targetSetImpl;
+  const superSetImplementations =
```
MauricioUyaguari, line 308: move this further down to after we check for parentTargetSetImpl

MauricioUyaguari (approved):

packages/legend-query/src/stores/QueryBuilderExplorerState.ts
```diff
@@ -282,6 +283,22 @@ const isAutoMappedProperty = (
   return false;
 };
 
+export const generatePropertyNodeMappingDataForNonPrimitveProperty = (
```
akphi, line 286: --> `NonPrimitive`

akphi (commented):
minor cleanups that I will do in https://github.com/finos/legend-studio/pull/830

PR #827, opened 2022-02-02
cancel PR running jobs from previous CI builds to save build resource

akphi:
## Summary

- [x] Re-disable SonarCloud scan in PR CI
- [x] Use [trivy-action](https://github.com/aquasecurity/trivy-action) instead of [Azure/container-scan](https://github.com/Azure/container-scan) for checking vulnerabilities in Docker images - see blocker issue https://github.com/Azure/container-scan/issues/122
- [x] Uninstall `scanitizer` app
- [x] Cancel running jobs from previous pipelines of the same workflow on PR to save resource when commits are pushed quickly.
Note that we don't want this behavior on default branch

- [ ] Test(s) added

### DIFF
--- a/.github/workflows/check-build.yml
+++ b/.github/workflows/check-build.yml
@@ -9,6 +9,13 @@ on:
     branches:
       - '**'
 
+# Cancel running jobs from previous pipelines of the same workflow on PR to save resource when commits are pushed quickly
+# NOTE: we don't want this behavior on default branch
+# See https://stackoverflow.com/a/68422069
+concurrency:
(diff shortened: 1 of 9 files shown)


maoo:
> @maoo Can you help me uninstall the app https://github.com/apps/scanitizer now that we have switched to use `trivy` to scan for CVEs in Docker images? Thanks!

Scanitizer uninstalled. Well done with `trivy`! Is this going to be mentioned/documented somewhere? I'd love to link it as part of our project onboarding docs, so that other projects can use the same approach. If so, please keep me posted. TY!

Issue #819, opened 2022-01-27
Feature request: Add function explorer panel in Query Builder for user defined functions


YannanGao-gs:
### Similar requests

- [X] I have searched and found no existing similar requests

### How are you using Studio?

Legend Studio

### What problems are you trying to solve?

To enhance our user experience in using user-defined functions in the query builder. 
Adding a function explorer panel would give users ability to explorer functions defined in the project and drag and drop into fetch structure panel. 

### Implementation Ideas/Steps
- [x] Enable a new panel for user-defined functions 
- [x] Two modes of displaying by package or flatten out by function name
- [x] `CONSIDER` Leverage the `doc.doc` tag value so producers of function can add documentation for the function and consumers can read tag value and learn what function does
- [x] For functions whom return type is primitive users can DND directly into fetch structure. 
- [x] Drag and drop into derivation panel 
- [ ] `CONSIDER` include on the fly `calculated attributes` where users can reuse defined attributes within query builder: @akphi - I have thought about this and I think we can support it but it doesn't have to persistable, meaning that within one editing session, the user can create their calculated attribute to save time, then DnD it etc. but this will not exist in the next editing session (I have seen this behavior in Google Sheets before with number format)
- [x] `CONSIDER` add a dropdown somewhere to enable/disable panels to show as we start to have too many panels (parameters, setup, explorer, functions, calculated attributes, etc.)

akphi:
I have updated the list a bit to add my own refinement to the implementation plan

Issue #627, opened 2021-10-29
Feature Request: support derivation as LHS for filter, post-filter, aggregation


akphi:
### Similar requests

- [X] I have searched and found no existing similar requests

## Context and Motivation
We want to support use cases like the following:

```
|model::Person.all()->filter(
  p|$p.firstName + $p.lastName == 'John Doe'
)->project(
  [
    x|$x.firstName,
    x|$x.lastName
  ],
  [
    'First Name',
    'Last Name'
  ]
)
```
Currently, this type of query triggers the unsupported view on the user which is very user-friendly. 

### What problems are you trying to solve?

This is similar to https://github.com/finos/legend-studio/issues/254, but this time we want to do it for `filter`, `post-filter`, and `aggregation`

### Describe the solution you would like to see implemented

This differs from https://github.com/finos/legend-studio/issues/254 a bit since a filter is not a list like projections, but a tree structure. As such, some special handling must be in place. For example:
 - [ ] `?` Add a new method to `QueryBuilderFilterOperator`
 - [ ] Create another visitor that extends `V1_ValueSpecificationBuilder` visitor that handles the building within the `filter` function in a special way, because unlike `projection`, `filter` is a tree structure.
- [ ] When we introduce `derivation` for filter, we would need to support something similar to https://github.com/finos/legend-studio/pull/888
- [ ] We need to take care of the case in #839 where after DnD to create the column, the user change the projection derivation changing the return type, do we need to reevaluate to check if post-filter condition still valid or not

### Describe alternatives you have considered

N/A

### Contribution

- [ ] I would like to work on this feature

PR #888, opened 2022-02-25
add functions explorer panel in query builder

YannanGao-gs:
## Summary
Fixes https://github.com/finos/legend-studio/issues/819

- [ ] Test(s) added

https://user-images.githubusercontent.com/73408381/157114489-4d54b6e5-04fd-4c12-a99a-23ba299a373c.mov

### DIFF
--- a/.changeset/cuddly-rockets-kick.md
+++ b/.changeset/cuddly-rockets-kick.md
@@ -1,0 +1,5 @@
+---
+'@finos/legend-query': minor
+---
+
+Add functions explorer panel to show user-defined functions
--- a/.changeset/curly-swans-glow.md
+++ b/.changeset/curly-swans-glow.md
@@ -1,0 +1,5 @@
+---
+'@finos/legend-graph': patch
+---
+
+Added helper to generate function signature in text.
--- a/packages/legend-query/src/components/QueryBuilder.tsx
+++ b/packages/legend-query/src/components/QueryBuilder.tsx
@@ -24,6 +24,13 @@ import {
   ResizablePanel,
   ResizablePanelSplitter,
   HackerIcon,
+  DropdownMenu,
+  MenuContent,
+  MenuContentItem,
+  MenuContentItemIcon,
+  MenuContentItemLabel,
+  CheckIcon,
+  CaretDownIcon,
 } from '@finos/legend-art';
 import { QueryBuilderFilterPanel } from './QueryBuilderFilterPanel';
 import { QueryBuilderExplorerPanel } from './QueryBuilderExplorerPanel';
@@ -39,6 +46,7 @@ import { QueryBuilderUnsupportedQueryEditor } from './QueryBuilderUnsupportedQue
 import { useApplicationStore } from '@finos/legend-application';
 import { QueryBuilderParameterPanel } from './QueryBuilderParameterPanel';
 import { QueryBuilderPostFilterPanel } from './QueryBuilderPostFilterPanel';
+import { QueryBuilderFunctionsExplorerPanel } from './QueryBuilderFunctionsExplorerPanel';
 
 enum QUERY_BUILDER_HOTKEY {
   COMPILE = 'COMPILE',
@@ -122,11 +130,66 @@ export const QueryBuilder = observer(
         );
       },
     };
+    const toggleShowFunctionPanel = (): void => {
+      queryBuilderState.setShowFunctionPanel(
+        !queryBuilderState.showFunctionPanel,
+      );
+    };
+    const toggleShowParameterPanel = (): void => {
+      queryBuilderState.setShowParameterPanel(
+        !queryBuilderState.showParameterPanel,
+      );
+    };
+
     return (
       <div
         data-testid={QUERY_BUILDER_TEST_ID.QUERY_BUILDER}
         className="query-builder"
       >
+        <div className="query-builder__sub-header">
+          <div className="query-builder__sub-header__actions">
+            <DropdownMenu
+              className="query-builder__sub-header__custom-action"
+              content={
+                <MenuContent>
+                  <MenuContentItem onClick={toggleShowFunctionPanel}>
+                    <MenuContentItemIcon>
+                      {queryBuilderState.showFunctionPanel ? (
+                        <CheckIcon />
+                      ) : null}
+                    </MenuContentItemIcon>
+                    <MenuContentItemLabel className="query-builder__sub-header__menu-content">
+                      Show Functions Explorer Panel
+                    </MenuContentItemLabel>
+                  </MenuContentItem>
+                  <MenuContentItem onClick={toggleShowParameterPanel}>
+                    <MenuContentItemIcon>
+                      {queryBuilderState.showParameterPanel ? (
+                        <CheckIcon />
+                      ) : null}
+                    </MenuContentItemIcon>
+                    <MenuContentItemLabel className="query-builder__sub-header__menu-content">
+                      Show Parameter Panel
+                    </MenuContentItemLabel>
+                  </MenuContentItem>
+                </MenuContent>
+              }
+              menuProps={{
+                anchorOrigin: { vertical: 'bottom', horizontal: 'left' },
+                transformOrigin: { vertical: 'top', horizontal: 'left' },
+                elevation: 7,
+              }}
+            >
+              <button
+                className="query-builder__sub-header__custom-action-content"
+                title="Show Advanced Menu..."
+              >
+                Advanced
+                <CaretDownIcon />
+              </button>
+            </DropdownMenu>
+          </div>
+        </div>
         <GlobalHotKeys keyMap={keyMap} handlers={handlers}>
           <Backdrop className="backdrop" open={queryBuilderState.backdrop} />
           <div className="query-builder__content">
@@ -157,11 +220,31 @@ export const QueryBuilder = observer(
                           />
                         </ResizablePanel>
                         <ResizablePanelSplitter />
-                        <ResizablePanel minSize={40} direction={-1}>
-                          <QueryBuilderParameterPanel
-                            queryBuilderState={queryBuilderState}
-                          />
-                        </ResizablePanel>
+                        {queryBuilderState.showFunctionPanel && (
+                          <ResizablePanel
+                            minSize={40}
+                            direction={
+                              queryBuilderState.showParameterPanel
+                                ? [1, -1]
+                                : -1
+                            }
+                          >
+                            <QueryBuilderFunctionsExplorerPanel
+                              queryBuilderState={queryBuilderState}
+                            />
+                          </ResizablePanel>
+                        )}
+                        {queryBuilderState.showFunctionPanel &&
+                        queryBuilderState.showParameterPanel ? (
+                          <ResizablePanelSplitter />
+                        ) : null}
+                        {queryBuilderState.showParameterPanel && (
+                          <ResizablePanel minSize={40} direction={-1}>
+                            <QueryBuilderParameterPanel
+                              queryBuilderState={queryBuilderState}
+                            />
+                          </ResizablePanel>
+                        )}
                       </ResizablePanelGroup>
                     </ResizablePanel>
                     <ResizablePanelSplitter />
--- a/packages/legend-query/src/components/QueryBuilderFunctionsExplorerPanel.tsx
+++ b/packages/legend-query/src/components/QueryBuilderFunctionsExplorerPanel.tsx
@@ -1,0 +1,599 @@
+/**
+ * Copyright (c) 2020-present, Goldman Sachs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { observer } from 'mobx-react-lite';
+import {
+  type TooltipPlacement,
+  type TreeNodeContainerProps,
+  FolderIcon,
+  FolderOpenIcon,
+  InfoCircleIcon,
+  SubjectIcon,
+  Tooltip,
+  ViewHeadlineIcon,
+  ChevronDownIcon,
+  ChevronRightIcon,
+  PURE_FunctionIcon,
+  CompressIcon,
+  TreeView,
+  BlankPanelContent,
+  MenuContentItem,
+  DropdownMenu,
+  MenuContent,
+  CheckIcon,
+  MenuContentItemIcon,
+  MenuContentItemLabel,
+  MoreVerticalIcon,
+} from '@finos/legend-art';
+import {
+  type QueryBuilderFunctionsExplorerTreeNodeData,
+  type QueryBuilderFunctionsExplorerDragSource,
+  QUERY_BUILDER_FUNCTIONS_EXPLORER_TREE_DND_TYPE,
+  generateFunctionsExplorerTreeNodeData,
+  getFunctionsExplorerTreeNodeChildren,
+} from '../stores/QueryFunctionsExplorerState';
+import { useDrag, useDragLayer } from 'react-dnd';
+import {
+  ConcreteFunctionDefinition,
+  generateFunctionSignature,
+  Package,
+  ROOT_PACKAGE_NAME,
+} from '@finos/legend-graph';
+import { getEmptyImage } from 'react-dnd-html5-backend';
+import { useEffect, useState } from 'react';
+import type { QueryBuilderState } from '../stores/QueryBuilderState';
+import { getMultiplicityDescription } from './shared/QueryBuilderUtils';
+
+const isDependencyTreeNode = (
+  node: QueryBuilderFunctionsExplorerTreeNodeData,
+): boolean =>
+  node.packageableElement.getRoot().path ===
+  ROOT_PACKAGE_NAME.PROJECT_DEPENDENCY_ROOT;
+
+const QueryBuilderFunctionInfoTooltip: React.FC<{
+  element: ConcreteFunctionDefinition;
+  children: React.ReactElement;
+  placement?: TooltipPlacement | undefined;
+}> = (props) => {
+  const { element, children, placement } = props;
+  return (
+    <Tooltip
+      arrow={true}
+      {...(placement !== undefined ? { placement } : {})}
+      classes={{
+        tooltip: 'query-builder__tooltip',
+        arrow: 'query-builder__tooltip__arrow',
+        tooltipPlacementRight: 'query-builder__tooltip--right',
+      }}
+      TransitionProps={{
+        timeout: 0,
+      }}
+      title={
+        <div className="query-builder__tooltip__content">
+          <div className="query-builder__tooltip__item">
+            <div className="query-builder__tooltip__item__label">
+              Function Name
+            </div>
+            <div className="query-builder__tooltip__item__value">
+              {element.path}
+            </div>
+          </div>
+          <div className="query-builder__tooltip__item">
+            <div className="query-builder__tooltip__item__label">
+              Parameters
+            </div>
+            <div className="query-builder__tooltip__item__value">
+              {element.parameters
+                .map(
+                  (t) =>
+                    `${t.name}: ${
+                      t.type.value.name
+                    }${getMultiplicityDescription(t.multiplicity)}`,
+                )
+                .join('; ')}
+            </div>
+          </div>
+          <div className="query-builder__tooltip__item">
+            <div className="query-builder__tooltip__item__label">
+              Documentation
+            </div>
+            <div className="query-builder__tooltip__item__value">
+              {element.taggedValues
+                .filter((t) => t.tag.value.owner.name === 'doc')
+                .map((t) => t.value)
+                .join('; ')}
+            </div>
+          </div>
+          <div className="query-builder__tooltip__item">
+            <div className="query-builder__tooltip__item__label">
+              Return Type
+            </div>
+            <div className="query-builder__tooltip__item__value">
+              {element.returnType.value.name}
+            </div>
+          </div>
+        </div>
+      }
+    >
+      <span>{children}</span>
+    </Tooltip>
+  );
+};
+
+const QueryBuilderFunctionDragLayer = observer(
+  (props: { queryBuilderState: QueryBuilderState }) => {
+    const { itemType, item, isDragging, currentPosition } = useDragLayer(
+      (monitor) => ({
+        itemType:
+          monitor.getItemType() as QUERY_BUILDER_FUNCTIONS_EXPLORER_TREE_DND_TYPE,
+        item: monitor.getItem<QueryBuilderFunctionsExplorerDragSource | null>(),
+        isDragging: monitor.isDragging(),
+        initialOffset: monitor.getInitialSourceClientOffset(),
+        currentPosition: monitor.getClientOffset(),
+      }),
+    );
+    if (
+      !isDragging ||
+      !item ||
+      !Object.values(QUERY_BUILDER_FUNCTIONS_EXPLORER_TREE_DND_TYPE).includes(
+        itemType,
+      ) ||
+      item.node.dndType ===
+        QUERY_BUILDER_FUNCTIONS_EXPLORER_TREE_DND_TYPE.PACKAGE
+    ) {
+      return null;
+    }
+    return (
+      <div className="query-builder__functions-explorer__tree__drag-preview-layer">
+        <div
+          className="query-builder__functions-explorer__tree__drag-preview"
+          style={
+            !currentPosition
+              ? { display: 'none' }
+              : {
+                  transform: `translate(${currentPosition.x + 20}px, ${
+                    currentPosition.y + 10
+                  }px)`,
+                }
+          }
+        >
+          {generateFunctionSignature(
+            item.node.packageableElement as ConcreteFunctionDefinition,
+            true,
+          )}
+        </div>
+      </div>
+    );
+  },
+);
+
+const QueryBuilderFunctionsExplorerListEntry = observer(
+  (props: {
+    queryBuilderState: QueryBuilderState;
+    element: ConcreteFunctionDefinition;
+    rootPackageName: ROOT_PACKAGE_NAME;
+  }) => {
+    const { queryBuilderState, element, rootPackageName } = props;
+    const node = generateFunctionsExplorerTreeNodeData(
+      queryBuilderState,
+      element,
+      rootPackageName,
+    );
+    const functionSignature = generateFunctionSignature(element, true);
+    const [, dragConnector, dragPreviewConnector] = useDrag(
+      () => ({
+        type: QUERY_BUILDER_FUNCTIONS_EXPLORER_TREE_DND_TYPE.FUNCTION,
+        item: { node: node },
+      }),
+      [node],
+    );
+    // hide default HTML5 preview image
+    useEffect(() => {
+      dragPreviewConnector(getEmptyImage(), { captureDraggingState: true });
+    }, [dragPreviewConnector]);
+
+    return (
+      <div
+        className="query-builder__functions-explorer__function"
+        ref={dragConnector}
+      >
+        <div className="query-builder__functions-explorer__function__content">
+          <div className="query-builder__functions-explorer__function__icon">
+            <div className="query-builder__functions-explorer__function-icon">
+              <PURE_FunctionIcon />
+            </div>
+          </div>
+          <div
+            className="query-builder__functions-explorer__function__label"
+            title={functionSignature}
+          >
+            {functionSignature}
+          </div>
+        </div>
+        <div className="query-builder__functions-explorer__function__actions">
+          <div className="query-builder__functions-explorer__function__action">
+            <QueryBuilderFunctionInfoTooltip
+              element={node.packageableElement as ConcreteFunctionDefinition}
+            >
+              <div className="query-builder__functions-explorer__function__action query-builder__functions-explorer__function__node__info">
+                <InfoCircleIcon />
+              </div>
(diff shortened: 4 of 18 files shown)


YannanGao-gs:
@beekemarie Discussed with Beeke. 
Consider:

- Hide the parameter panel in the default page of Query Builder too
- Choose a meaningful color for the second header(functional band) as blue is already used by service registration

packages/legend-query/src/stores/QueryFunctionsState.ts
```diff
@@ -0,0 +1,468 @@
...
+        `${p.name}: ${p.type.value.name}${getSimpleMultiplicityDescription(
+          p.multiplicity,
+        )}`,
+    )
+    .join(', ')})` +
+  `: ${element.returnType.value.name}${getSimpleMultiplicityDescription(
+    element.returnMultiplicity,
+  )}`;
+
+export const generateDefaultValueForPrimitiveParameter = (
```
akphi, line 256: Can we reuse this from somewhere?

packages/legend-query/src/stores/QueryFunctionsState.ts
```diff
@@ -0,0 +1,468 @@
...
+        case ROOT_PACKAGE_NAME.PROJECT_DEPENDENCY_ROOT:
+          this.setDependencyTreeData({ ...data });
+          break;
+        default:
+          this.setTreeData({ ...data });
+      }
+    }
+  };
+
+  refreshTreeData(): void {
```
akphi, line 437: I'm not so sure if we need to refresh this tree

packages/legend-query/src/stores/QueryFunctionsState.ts
```diff
@@ -0,0 +1,468 @@
...
+    }
+    if (this.dependencyTreeData) {
+      this.dependencyTreeData = { ...this.dependencyTreeData };
+    }
+  }
+
+  get nonNullableTreeData(): TreeData<QueryBuilderPackageElementTreeNodeData> {
+    return guaranteeNonNullable(
+      this.treeData,
+      'Query builder function tree data has not been initialized',
```
akphi, line 407: --> `... functions explorer tree ...`

packages/legend-query/style/_query-builder.scss
```diff
@@ -35,6 +35,51 @@
     background: var(--color-dark-grey-50);
   }
 
+  &__fuctional-header {
+    @include flexVCenter;
+
+    height: 2.8rem;
+    background: #7399c6;
```
akphi, line 42: Let's use the _boring_ dark grey.

packages/legend-query/src/components/QueryBuilderFunctionPanel.tsx
```diff
@@ -0,0 +1,653 @@
...
+                  (t) =>
+                    `${t.name}: ${
+                      t.type.value.name
+                    }${getMultiplicityDescription(t.multiplicity)}`,
+                )
+                .join('; ')}
+            </div>
+          </div>
+          <div className="query-builder__tooltip__item">
+            <div className="query-builder__tooltip__item__label">
```
akphi, line 107: What do we show when there are no docs?

packages/legend-query/src/components/QueryBuilderFunctionPanel.tsx
```diff
@@ -0,0 +1,653 @@
...
+            </div>
+            <div className="query-builder__tooltip__item__value">
+              {element.taggedValues
+                .filter((t) => t.tag.value.owner.name === 'doc')
+                .map((t) => t.value)
+                .join('; ')}
+            </div>
+          </div>
+          <div className="query-builder__tooltip__item">
+            <div className="query-builder__tooltip__item__label">
```
akphi, line 118: Showing `tagged values` might be a little too cluttered. What do you think @MauricioUyaguari?

packages/legend-query/src/components/QueryBuilderFunctionPanel.tsx
```diff
@@ -0,0 +1,653 @@
...
+              queryBuilderState,
+              fromDependency: false,
+            }}
+          />
+        )}
+        {rootPackageName === ROOT_PACKAGE_NAME.PROJECT_DEPENDENCY_ROOT &&
+          dependencyTreeData &&
+          showDependencyFuncions && (
+            <>
+              <div className="query-builder__functions__separator" />
```
akphi, line 474: can we propagate the change I mentioned in naming classes `functions-explorer`?

packages/legend-query/src/components/QueryBuilderFunctionPanel.tsx
```diff
@@ -0,0 +1,653 @@
...
+  },
+);
+
+const QueryBuilderFunctionExplorerTreeNodeView = observer(
+  (
+    props: TreeNodeViewProps<
+      QueryBuilderPackageElementTreeNodeData,
+      {
+        queryBuilderState: QueryBuilderState;
+        fromDependency: boolean;
```
akphi, line 366: probably don't need this right? we can follow what we do in the project explorer tree.

packages/legend-query/src/components/QueryBuilderFunctionPanel.tsx
```diff
@@ -0,0 +1,653 @@
...
+                }}
+              />
+            </>
+          )}
+      </>
+    );
+  },
+);
+
+export const QueryBuilderFunctionPanel = observer(
```
akphi, line 497: `QueryBuilderFunctionsExplorerPanel`, let's also rename this file and its related states in `QueryBuilderState`

packages/legend-query/src/components/QueryBuilderFunctionPanel.tsx
```diff
@@ -0,0 +1,653 @@
...
+  (props: { queryBuilderState: QueryBuilderState }) => {
+    const { queryBuilderState } = props;
+    const queryFunctionsState = queryBuilderState.queryFunctionsState;
+    const [viewAsTree, setViewAsTree] = useState(false);
```
akphi, line 501: *clap*

akphi (changes requested):
Nice job! I just left minor comments on renaming things, also we should check the style again later. Thanks for this!!

packages/legend-query/src/components/QueryBuilderProjectionPanel.tsx
```diff
@@ -494,31 +515,64 @@ export const QueryBuilderProjectionPanel = observer(
...
     const handleDrop = useCallback(
-      (item: QueryBuilderExplorerTreeDragSource): void =>
-        projectionState.addColumn(
-          new QueryBuilderSimpleProjectionColumnState(
-            projectionState,
-            buildPropertyExpressionFromExplorerTreeNodeData(
-              queryBuilderState.explorerState.nonNullableTreeData,
-              item.node,
-              projectionState.queryBuilderState.graphManagerState.graph,
+      (
```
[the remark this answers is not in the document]
MauricioUyaguari: I dont think there is a difference between DnD into derivation and creating a new derivation. 
We also don't force users to compile during derivation so there are no special checks that run that wouldnt run with this. This has the same risk as doing it with an already defined derivation.

packages/legend-query/src/components/QueryBuilderFunctionPanel.tsx
MauricioUyaguari: Could be. What if we just show if it has documentation. `doc.doc`?

packages/legend-query/src/components/QueryBuilderFunctionPanel.tsx
YannanGao-gs: Screen Shot 2022-03-01 at 1 22 07 PM

packages/legend-query/src/stores/QueryFunctionsState.ts
YannanGao-gs: Tried to use `generateDefaultValueForPrimitiveType`, but didn't work as i expect

.changeset/cuddly-rockets-kick.md
```diff
@@ -0,0 +1,5 @@
+---
+'@finos/legend-query': patch
+---
+
+Add a function explorer panel in Query Builder for user-defined functions
```
akphi, line 5:
```suggestion
Add functions explorer panel to show user-defined functions
```

packages/legend-query/src/components/QueryBuilder.tsx
```diff
@@ -121,11 +132,58 @@ export const QueryBuilder = observer(
...
+            <DropdownMenu
+              className="query-builder__sub-header__custom-action"
+              content={
+                <MenuContent>
+                  <MenuContentItem onClick={toggleShowFunctionPanel}>
+                    <MenuContentItemIcon>
+                      {showFunctionPanel ? <CheckIcon /> : null}
+                    </MenuContentItemIcon>
+                    <MenuContentItemLabel className="query-builder__sub-header__menu-content">
+                      Show Function Panel
```
akphi, line 158: --> `Show Functions Explorer Panel`

packages/legend-query/src/components/QueryBuilder.tsx
```diff
@@ -121,11 +132,58 @@ export const QueryBuilder = observer(
...
+                anchorOrigin: { vertical: 'bottom', horizontal: 'left' },
+                transformOrigin: { vertical: 'top', horizontal: 'left' },
+                elevation: 7,
+              }}
+            >
+              <button
+                className="query-builder__sub-header__custom-action-content"
+                title="Show Advanced Menu..."
+              >
+                {'Advanced'}
```
akphi, line 181: --> `Advanced`

packages/legend-query/src/components/QueryBuilder.tsx
```diff
@@ -106,6 +115,8 @@ export const QueryBuilder = observer(
     const { queryBuilderState } = props;
     const applicationStore = useApplicationStore();
     const isQuerySupported = queryBuilderState.isQuerySupported();
+    const [showFunctionPanel, setShowFunctionPanel] = useState(false);
```
akphi, line 118: why do we use local states here? For these, we could add states to the `QueryBuilderState` I think.

packages/legend-query/style/index.scss
```diff
@@ -22,3 +22,4 @@
 @forward 'graph-fetch-tree';
 @forward 'query-setup';
 @forward 'query-editor';
+@forward 'query-builder-function-explorer';
```
akphi, line 25: --> `query-builder-functions-explorer`

packages/legend-query/src/components/QueryBuilderFunctionsExplorerPanel.tsx
```diff
@@ -0,0 +1,588 @@
...
+    </Tooltip>
+  );
+};
+
+const QueryBuilderFunctionDragLayer = observer(
+  (props: { queryBuilderState: QueryBuilderState }) => {
+    const { itemType, item, isDragging, currentPosition } = useDragLayer(
+      (monitor) => ({
+        itemType: monitor.getItemType() as QUERY_BUILDER_FUNCTION_TREE_DND_TYPE,
+        item: monitor.getItem() as QueryBuilderFunctionDragSource | null,
```
akphi, line 149: let's do this the proper way

getItem<...>

packages/legend-query/src/components/QueryBuilderFunctionsExplorerPanel.tsx
```diff
@@ -0,0 +1,588 @@
...
+              Documentation
+            </div>
+            <div className="query-builder__tooltip__item__value">
+              {element.taggedValues
+                .filter((t) => t.tag.value.owner.name === 'doc')
+                .map((t) => t.value)
+                .join('; ')}
+            </div>
+          </div>
+          <div className="query-builder__tooltip__item">
```
akphi, line 120: Let's remove `stereotypes` for now

packages/legend-query/src/components/QueryBuilderFunctionsExplorerPanel.tsx
```diff
@@ -0,0 +1,588 @@
...
+            showDependencyFuncions &&
+            queryFunctionsState.dependencyTreeData &&
+            queryFunctionsState.dependencyTreeData.rootIds.length !== 0 && (
+              <QueryBuilderPackableElementExplorerTree
+                queryBuilderState={queryBuilderState}
+                rootPackageName={ROOT_PACKAGE_NAME.PROJECT_DEPENDENCY_ROOT}
+                showDependencyFuncions={showDependencyFuncions}
+              />
+            )}
+          {/* list view */}
```
akphi, line 552: For list view, I don't think we need a tree

packages/legend-query/src/components/QueryBuilderFunctionsExplorerPanel.tsx
```diff
@@ -0,0 +1,588 @@
...
+          />
+          {((showDependencyFuncions &&
+            (!queryFunctionsState.dependencyTreeData ||
+              queryFunctionsState.dependencyTreeData.rootIds.length === 0)) ||
+            !showDependencyFuncions) &&
+            queryFunctionsState.treeData?.rootIds.length === 0 && (
+              <BlankPanelContent>No user-defined functions</BlankPanelContent>
+            )}
+          {/* tree view */}
+          {viewAsTree && queryFunctionsState.treeData && (
```
akphi, line 535: We can use the `Fragment` to group this better, i.e. `{viewAsTree && <>...</>}

packages/legend-query/src/components/QueryBuilderFunctionsExplorerPanel.tsx
```diff
@@ -0,0 +1,588 @@
...
+              </div>
+            </QueryBuilderFunctionInfoTooltip>
+          </div>
+        </div>
+      </div>
+    );
+  },
+);
+
+const QueryBuilderFunctionExplorerTreeNodeContainer = observer(
```
akphi, line 242: You should be consistent and scan for all `functionExplorer` and rename them to `functionsExplorer`

packages/legend-query/src/stores/QueryFunctionsState.ts
```diff
@@ -0,0 +1,457 @@
...
+  node: QueryBuilderFunctionsExplorerTreeNodeData;
+}
+
+export class QueryBuilderFunctionsExplorerTreeNodeData implements TreeNodeData {
+  id: string;
+  label: string;
+  dndType: string;
+  childrenIds: string[] = [];
+  isOpen?: boolean | undefined;
+  packageableElement: PackageableElement;
```
akphi, line 54: hm, so where is my `new line`?

akphi (changes requested):

packages/legend-query/src/components/QueryBuilderFunctionsExplorerPanel.tsx
akphi: my bad, but let's rename the container component

packages/legend-query/src/components/QueryBuilderProjectionPanel.tsx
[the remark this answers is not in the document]
akphi: I still don't think this feature makes sense unfortunately. We're basically saying we let you create a projection from a function whereas the simple path we want to shape users into doing is to create a derivation and to be able to DnD function there to save typings

Now, I understand the exploratory aspect of this where it goes like, people are curious and playing with the UI then they discover that the functions are dragable, and they are droppable in the projection panel.

But now, consider the consistency aspect of this behavior, when we have filter, how do you plan to support this? You add an adhoc condition?

packages/legend-query/src/stores/QueryBuilderState.ts
```diff
@@ -123,6 +124,7 @@ export class QueryBuilderState {
   querySetupState: QueryBuilderSetupState;
   explorerState: QueryBuilderExplorerState;
   queryParametersState: QueryParametersState;
+  queryFunctionsState: QueryFunctionsState;
```
akphi, line 127: let's be consistent and do `queryFunctionsExplorerState`

packages/legend-query/src/stores/QueryFunctionsState.ts
```diff
@@ -0,0 +1,458 @@
...
+    }
+    if (this.dependencyTreeData) {
+      this.dependencyTreeData = { ...this.dependencyTreeData };
+    }
+  }
+
+  get nonNullableTreeData(): TreeData<QueryBuilderFunctionsExplorerTreeNodeData> {
+    return guaranteeNonNullable(
+      this.treeData,
+      'Query builder function explorer tree data has not been initialized',
```
akphi, line 397: --> `functions explorer`

packages/legend-query/src/components/QueryBuilderValueSpecificationEditor.tsx
```diff
@@ -225,6 +225,7 @@ const NumberPrimitiveInstanceValueEditor = observer(
   },
 );
 
+//here
```
akphi, line 228: let's remove this :)

packages/legend-query/src/components/QueryBuilderValueSpecificationEditor.tsx
```diff
@@ -542,6 +543,7 @@ export const DateInstanceValueEditor = observer(
       }
     };
 
+    // heads up
```
akphi, line 546: interesting.

packages/legend-query/src/components/QueryBuilderFunctionsExplorerPanel.tsx
```diff
@@ -0,0 +1,580 @@
...
+            item.node.packageableElement as ConcreteFunctionDefinition,
+            true,
+          )}
+        </div>
+      </div>
+    );
+  },
+);
+
+const SimpleQueryBuilderFunctionsExplorerTreeNodeContainer = observer(
```
akphi, line 179: --> `QueryBuilderFunctionsExplorerListEntry`?

akphi (changes requested):
Looks good overall, except for some renaming and we need to make decision on the behavior of DnD. Thanks!

packages/legend-query/src/components/QueryBuilderProjectionPanel.tsx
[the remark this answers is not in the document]
akphi: After discussing with @MauricioUyaguari I think I can _bend_ on this one 😄

But as said, my biggest concerns are:
1. Adding a feature like this is nice, but adding new `easter-egg` features like this costs maintenance (both code and docs)
2. We have to be careful of the implication this has to the UX, to me derivation is a dangerous thing to use (could be me being too _shy_ because our support for derivation at this point is too basic), so in the course of exploring the panel, people can accidentally add a derivation in projection and they would wonder what is this thing.
3. To point (1), we now bear the burden of maintaining the consistency in UX, now when we support derivation better in filter, aggregation, etc. we need to make sure this works. I will add more infos to #627

packages/legend-query/src/components/QueryBuilderProjectionPanel.tsx
```diff
@@ -494,31 +517,64 @@ export const QueryBuilderProjectionPanel = observer(
...
+              ),
+            );
+          derivationProjectionColumn.derivationLambdaEditorState.setLambdaString(
+            `x|${generateFunctionLambdaString(
+              (item as QueryBuilderFunctionsExplorerDragSource).node
+                .packageableElement as ConcreteFunctionDefinition,
+            )}`,
+          );
+          projectionState.addColumn(derivationProjectionColumn);
+        } else {
```
akphi, line 548: --> so assuming in the `else` block like this is not wrong but could be dangerous in the future when we add more handling, perhaps it's better to rewrite this as a `switch-case` here

packages/legend-query/style/_query-builder.scss
```diff
@@ -35,6 +35,51 @@
     background: var(--color-dark-grey-50);
   }
 
+  &__sub-header {
+    @include flexVCenter;
+
+    height: 2.8rem;
+    background: darkgray;
```
akphi, line 42: Please don't use color that we don't define here. Also, we could make use of darker grey that we already have.

packages/legend-query/style/_query-builder-functions-explorer.scss
```diff
@@ -0,0 +1,203 @@
...
+  }
+
+  &__drag-preview {
+    @include flexCenter;
+
+    display: inline-flex;
+    height: 2rem;
+    padding: 0 0.5rem;
+    font-size: 1.2rem;
+    background: var(--color-blue-200);
```
akphi, line 93: Let's use `grey` for this. This is no primitive properties.

packages/legend-query/style/_query-builder-functions-explorer.scss
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    &-icon {
+      display: flex;
+      justify-content: center;
+      width: 1.5rem;
+      min-width: 1.5rem;
+    }
+
+    &__label {
```
akphi, line 173: I don't think I see `truncation` kicking in the demo video you showed

packages/legend-query/src/components/QueryBuilderFunctionsExplorerPanel.tsx
```diff
@@ -0,0 +1,590 @@
...
+        className="query-builder__functions-explorer__function"
+        ref={dragConnector}
+      >
+        <div className="query-builder__functions-explorer__function__content">
+          <div className="query-builder__functions-explorer__function__icon">
+            <div className="query-builder__functions-explorer__function-icon">
+              <PURE_FunctionIcon />
+            </div>
+          </div>
+          <div className="query-builder__functions__function__label">
```
akphi, line 218: We need to do truncation here and potentially for the tree as well, when the text is too long, people can nolonger read it, we need to allow them to hover and read the full signature, so let's add `title={generateFunctionSignature(element, true)}` here, which means, we'd probably should create a constant to hold this so we don't need to call it twice.

packages/legend-query/src/components/QueryBuilderFunctionsExplorerPanel.tsx
```diff
@@ -0,0 +1,590 @@
...
+        <div className="panel__content query-builder__functions-explorer__content">
+          <QueryBuilderFunctionDragLayer
+            queryBuilderState={queryBuilderState}
+          />
+          {((showDependencyFuncions &&
+            (!queryFunctionsState.dependencyTreeData ||
+              queryFunctionsState.dependencyTreeData.rootIds.length === 0)) ||
+            !showDependencyFuncions) &&
+            queryFunctionsState.treeData?.rootIds.length === 0 && (
+              <BlankPanelContent>No user-defined functions</BlankPanelContent>
```
akphi, line 527:
```suggestion
              <BlankPanelContent>No user-defined functions available</BlankPanelContent>
```

packages/legend-query/src/components/QueryBuilderFunctionsExplorerPanel.tsx
```diff
@@ -0,0 +1,590 @@
...
+            <button
+              className="panel__header__action"
+              onClick={collapseTree}
+              tabIndex={-1}
+            >
+              {viewAsTree && <CompressIcon title="Collapse Tree" />}
+            </button>
+            <div className="panel__header__action">
+              {!viewAsTree ? (
+                <SubjectIcon
```
akphi, line 480: I think we need to scale these icons a bit larger.

packages/legend-query/src/stores/QueryFunctionsExplorerState.ts
```diff
@@ -0,0 +1,456 @@
...
+      return true;
+    case PRIMITIVE_TYPE.FLOAT:
+    case PRIMITIVE_TYPE.DECIMAL:
+      return 0.0;
+    case PRIMITIVE_TYPE.NUMBER:
+    case PRIMITIVE_TYPE.INTEGER:
+      return 0;
+    case PRIMITIVE_TYPE.DATE:
+    case PRIMITIVE_TYPE.STRICTDATE:
+      return `%${format(new Date(2020, 10, 1), DATE_FORMAT)}`;
```
akphi, line 263: why not using today for these?

packages/legend-query/src/stores/QueryFunctionsExplorerState.ts
```diff
@@ -0,0 +1,456 @@
...
+    case PRIMITIVE_TYPE.DATETIME:
+      return `%${format(new Date(2020, 10, 1), DATE_TIME_FORMAT)}`;
+    case PRIMITIVE_TYPE.STRICTTIME:
+      return `%00:00:00`;
+    case PRIMITIVE_TYPE.STRING:
+      return "''";
+    default:
```
akphi, line 270: your function is not really doing what its name says, this method also handle `non-primitive` parameter. I suggest we call it `generateDefaultParameterValueForType` and let this handle `enumeration` as well as `class`. i.e. Let's do `instanceof Class`, `instanceof Enumeration`, `instanceof PrimitiveType`, etc.

packages/legend-query/src/stores/QueryFunctionsExplorerState.ts
```diff
@@ -0,0 +1,456 @@
...
+  element: ConcreteFunctionDefinition,
+): string => {
+  let lambdaString = '';
+  if (element.parameters.length > 0) {
+    for (let i = 0; i < element.parameters.length; i++) {
+      const paramType = element.parameters.at(i)?.type.value;
+      const separator = i !== element.parameters.length - 1 ? ', ' : '';
+      if (paramType instanceof Enumeration) {
+        lambdaString = `${lambdaString + paramType.path}.${
+          paramType.values.at(0)?.name
```
akphi, line 285: You haven't accounted for the case where the enumeration have no enum values here, I suggest we fall back to `param${index}`.

packages/legend-query/src/stores/QueryFunctionsExplorerState.ts
```diff
@@ -0,0 +1,456 @@
...
+    case PRIMITIVE_TYPE.STRICTTIME:
+      return `%00:00:00`;
+    case PRIMITIVE_TYPE.STRING:
+      return "''";
+    default:
+      return `param${index}`;
+  }
+};
+
+export const generateFunctionLambdaString = (
```
akphi, line 275: Let's rename this to `generateFunctionCallString`, it's not a lambda per se.

packages/legend-query/src/stores/QueryFunctionsExplorerState.ts
```diff
@@ -0,0 +1,456 @@
...
+};
+
+const getAllPackagesFromElement = (element: PackageableElement): Package[] => {
+  if (element.package) {
+    return [element.package].concat(getAllPackagesFromElement(element.package));
+  }
+  return [];
+};
+
+export const generateFunctionSignature = (
```
akphi, line 236: I could see we make use of this in core, let's move it and `generateDefaultValueForPrimitiveParameter` to `DomainHelper`

akphi (changes requested):

packages/legend-query/src/stores/QueryFunctionsExplorerState.ts
akphi: Would be nice if we could bring something similar to this to `DnD` of `PurePropertyMappingEditor` for derived properties you know

packages/legend-query/src/stores/QueryFunctionsExplorerState.ts
```diff
@@ -0,0 +1,456 @@
...
+    case PRIMITIVE_TYPE.DECIMAL:
+      return 0.0;
+    case PRIMITIVE_TYPE.NUMBER:
+    case PRIMITIVE_TYPE.INTEGER:
+      return 0;
+    case PRIMITIVE_TYPE.DATE:
+    case PRIMITIVE_TYPE.STRICTDATE:
+      return `%${format(new Date(2020, 10, 1), DATE_FORMAT)}`;
+    case PRIMITIVE_TYPE.DATETIME:
+      return `%${format(new Date(2020, 10, 1), DATE_TIME_FORMAT)}`;
```
akphi, line 265: I think here we have to do manually add `00:00:00` or so in the end, please check what we do in Pure for `DateTime`

packages/legend-query/style/_query-builder.scss
```diff
@@ -35,6 +35,51 @@
...
+    justify-content: flex-end;
+  }
+
+  &__sub-header__custom-action {
+    @include flexCenter;
+
+    height: 2.3rem;
+    padding: 0 1rem;
+    border-radius: 0.5rem;
```
akphi, line 51: --> `0.2rem`, we rarely do more than `0.2` :)

packages/legend-query/style/_query-builder-functions-explorer.scss
YannanGao-gs: I think we should use blue since this is for dragging preview. It has the same color as the explorer tree does.

packages/legend-query/src/stores/QueryFunctionsExplorerState.ts
YannanGao-gs: i also moved this to DomainHelper.ts. Pls me know if it is redundant.

packages/legend-studio/src/stores/shared/__tests__/MockDataUtil.test.ts
```diff
@@ -113,3 +117,15 @@ test(unitTest('Test mock data with classes cycle'), () => {
   // should not continue on to next depth
   expect(secondApplicantInstance).not.toContain('previousApplication');
 });
+
+test(unitTest('Generate default value for primitive parameter'), () => {
```
akphi, line 121: no! Let's create a separate test for these in `legend-graph`, something like `src/helpders/__tests__/DomainHelper.test.ts`

.changeset/curly-swans-glow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@finos/legend-graph': patch
+---
+
+Added utility functions for function signature and call string generation.
```
akphi, line 5:
```suggestion
Added helper to generate function signature in text.
```

.changeset/sixty-cameras-sip.md
```diff
@@ -0,0 +1,5 @@
+---
+'@finos/legend-studio': patch
+---
+
+Added tests for function signature and call string generation.
```
akphi, line 5: no need changeset for test.

akphi (changes requested):

akphi (approved):

PR #891, opened 2022-02-28
query builder: support editing of parameters for `milestoned` properties

### LINKED ISSUE #739: Support editing of parameters for milestoning properties in Query Builder
### Similar requests

Currently we don't give user an option to edit the date parameters for milestoned properties in projection/filter panel. If user were to edit them they have to go to text mode and add/edit those parameters.

- [x] `ENGINE` Support date propagation for milestoned properties in - finos/legend-engine#569
- [x] `PURE` Remove date propagation for milestoned properties during transforming of metamodel to protocol classes (transformer). Needed to test changes in engine; i.e. date propagation will only be done in `engine` now - finos/legend-pure#467
- [x] Propagate default date parameters to the milestoned properties
- [x] Allow users to edit the parameters for milestoned properties in projection/filter panel

### Grammar for testing

```
###Service
Service my::service
{
  pattern: '/4cf10b66-8243-457f-a8d1-2b6bf832b5dc';
  documentation: '';
  autoActivateUpdates: true;
  execution: Single
  {
    query: |my::Person.all(%latest)->project([x|$x.firm(%latest).legalName], ['Firm/Legal Name']);
    mapping: my::map;
    runtime:
    #{
      mappings:
      [
        my::map
      ];
    }#;
  }
  test: Single
  {
    data: '';
    asserts:
    [
    ];
  }
}


###Relational
Database my::db
(
  Table PersonTable
  (
    NAME CHAR(200),
    FIRMID INTEGER
  )
  Table FirmTable
  (
    FIRMNAME CHAR(200),
    ID INTEGER
  )

  Join Firm_Person(PersonTable.FIRMID = FirmTable.ID)
)


###Pure
Class <<temporal.businesstemporal>> my::Person
{
  name: String[1];
  firmID: Integer[1];
}

Class <<meta::pure::profiles::temporal.businesstemporal>> my::Firm
{
  legalName: String[1];
  id: Integer[1];
}

Association my::Employment
{
  firm: my::Firm[0..1];
  employees: my::Person[*];
}


###Mapping
Mapping my::map
(
  my::Person[per1]: Relational
  {
    ~mainTable [my::db]PersonTable
    name: [my::db]PersonTable.NAME
  }
  my::Firm[fir1]: Relational
  {
    ~mainTable [my::db]FirmTable
    legalName: [my::db]FirmTable.FIRMNAME
  }

  my::Employment: Relational
  {
    AssociationMapping
    (
      firm[per1,fir1]: [my::db]@Firm_Person
    )
  }
)
```

### Query to be supported in form mode

```
|my::Person.all(%latest)->project(
  [
    x|$x.firm(%latest).legalName
  ],
  ['Firm/Legal Name']
)
```
We should have an editor to edit the parameter for milestoning properties in projection panel and filter panel

### Contribution

gayathrir11:
## Summary

Closes #739 

- [ ] Test(s) added

***When top level class is not milestoned***

https://user-images.githubusercontent.com/87973126/159721730-3e7b198d-89ba-40fd-9706-f5369b7c1590.mp4

***When top level class is businesstemporal***

https://user-images.githubusercontent.com/87973126/159724745-eb2146b5-1922-4675-9f7e-5db75961b514.mp4

***When top level class is bitemporal***

https://user-images.githubusercontent.com/87973126/159727433-973cb77d-b3f5-4e39-935a-d0f9764635a0.mp4

***When to level class is processingtemporal***

https://user-images.githubusercontent.com/87973126/159733582-bd55782f-e17a-40cb-8f3f-b6ec999daab5.mp4

***Changing the value of bottom level milestoned property parameter when date propagation is supported***

https://user-images.githubusercontent.com/87973126/159730860-932ab783-d742-4b08-bebb-7f8c748bd341.mp4

***When default dates of top level property is changed***

https://user-images.githubusercontent.com/87973126/159731944-506dc1f5-4786-45e5-b724-63a121aa5923.mp4

***Flow diagram***
Milestoning


gayathrir11:
Model used for testing
```
###Service
Service my::service
{
  pattern: '/e70d2c08-4169-44f1-89d4-0fa9971c3995';
  documentation: '';
  autoActivateUpdates: true;
  execution: Single
  {
    query: businessDate: Date[1]|my::Person.all($businessDate);
    mapping: my::map;
    runtime:
    #{
      mappings:
      [
        my::map
      ];
    }#;
  }
  test: Single
  {
    data: '';
    asserts:
    [
    ];
  }
}

Service my::service1
{
  pattern: '/d8464027-6139-4e73-a225-ecd84c44334b';
  documentation: '';
  autoActivateUpdates: true;
  execution: Single
  {
    query: {businessDate: Date[1],processingDate: StrictDate[1]|my::Person1.all($processingDate, $businessDate)};
    mapping: my::map;
    runtime:
    #{
      mappings:
      [
        my::map
      ];
    }#;
  }
  test: Single
  {
    data: '';
    asserts:
    [
    ];
  }
}

Service my::service2
{
  pattern: '/80bfe84d-1f3a-4a59-a776-aa693db6729f';
  documentation: '';
  autoActivateUpdates: true;
  execution: Single
  {
    query: processingDate: StrictDate[1]|my::Person2.all($processingDate);
    mapping: my::map;
    runtime:
    #{
      mappings:
      [
        my::map
      ];
    }#;
  }
  test: Single
  {
    data: '';
    asserts:
    [
    ];
  }
}

Service my::service3
{
  pattern: '/80bfe84d-1f3a-4a59-a776-aa693db6729f';
  documentation: '';
  autoActivateUpdates: true;
  execution: Single
  {
    query: processingDate: StrictDate[1]|my::Person2.allVersions();
    mapping: my::map;
    runtime:
    #{
      mappings:
      [
        my::map
      ];
    }#;
  }
  test: Single
  {
    data: '';
    asserts:
    [
    ];
  }
}


###Relational
Database my::db
(
  Table PersonTable
  (
    NAME CHAR(200) PRIMARY KEY,
    FIRMID INTEGER
  )
  Table PersonTable1
  (
    NAME CHAR(200) PRIMARY KEY,
    FIRMID INTEGER
  )
  Table PersonTable2
  (
    NAME CHAR(200) PRIMARY KEY,
    FIRMID INTEGER
  )
  Table FirmTable
  (
    FIRMNAME CHAR(200) PRIMARY KEY,
    ID INTEGER
  )

  Join Firm_Person(PersonTable.FIRMID = FirmTable.ID)
  Join Person_Person1(PersonTable.FIRMID = PersonTable1.FIRMID)
  Join Person_Person2(PersonTable.FIRMID = PersonTable2.FIRMID)
  Join Person1_Person2(PersonTable1.FIRMID = PersonTable2.FIRMID)
)


###Pure
Class my::Firm
{
  legalName: String[1];
  id: Integer[1];
  businessTemporal: my::Person[1];
  biTemporal: my::Person1[1];
  processingTemporal: my::Person2[1];
}

Class <<temporal.businesstemporal>> my::Person
{
  name: String[1];
  firmID: Integer[1];
  date: DateTime[1];
  processingTemporal: my::Person2[1];
  businessTemporal: my::Person[1];
  biTemporal: my::Person1[1];
}

Class <<temporal.bitemporal>> my::Person1
{
  name: String[1];
  firmID: Integer[1];
  businessTemporal: my::Person[1];
  processingTemporal: my::Person2[1];
  biTemporal: my::Person1[1];
  prop(d: String[1]) {$this.name}: String[1];
}

Class <<temporal.processingtemporal>> my::Person2
{
  name: String[1];
  firmID: Integer[1];
  processingTemporal: my::Person2[1];
  biTemporal: my::Person1[1];
  businessTemporal: my::Person[1];
}


###Mapping
Mapping my::map
(
  *my::Person: Relational
  {
    ~primaryKey
    (
      [my::db]PersonTable.NAME
    )
    ~mainTable [my::db]PersonTable
    processingTemporal[my_Person2]: [my::db]@Person_Person2,
    firmID: 100,
    biTemporal[my_Person1]: [my::db]@Person_Person1,
    businessTemporal[my_Person]: [my::db]@Person_Person1,
    date: 2021
  }
  *my::Person1: Relational
  {
    ~primaryKey
    (
      [my::db]PersonTable1.NAME
    )
    ~mainTable [my::db]PersonTable1
    businessTemporal[my_Person]: [my::db]@Person_Person1,
    processingTemporal[my_Person2]: [my::db]@Person1_Person2,
    firmID: 101,
    biTemporal[my_Person1]: [my::db]@Person_Person1
  }
  *my::Person2: Relational
  {
    ~primaryKey
    (
      [my::db]PersonTable2.NAME
    )
    ~mainTable [my::db]PersonTable2
    biTemporal[my_Person1]: [my::db]@Person1_Person2,
    businessTemporal[my_Person]: [my::db]@Person_Person2,
    firmID: 101,
    processingTemporal[my_Person2]: [my::db]@Person_Person2
  }
  *my::Firm: Relational
  {
    ~primaryKey
    (
      [my::db]FirmTable.FIRMNAME
    )
    ~mainTable [my::db]FirmTable
    businessTemporal[my_Person]: [my::db]FirmTable.FIRMNAME,
    biTemporal[my_Person1]: [my::db]FirmTable.FIRMNAME,
    processingTemporal[my_Person2]: [my::db]FirmTable.FIRMNAME
  }
)
```

MauricioUyaguari:
@gayathrir11 I took a look at the videos. I think the logic makes sense for all of these. 
We may need to think about whether the edit value for the default milestone values are in the right place. 
I think what we may need though is some kind of logic diagram that walks us through the logic of choices to make it clear. Again from the videos I think you made the correct decisions. 
Code wise, is this ready for review ?

packages/legend-query/src/stores/QueryBuilderMilestoningHelper.ts
```diff
@@ -0,0 +1,603 @@
...
+    const parameter = new VariableExpression(
+      milestoningParameter.name,
+      multiplicity,
+    );
+    return parameter;
+  }
+  return undefined;
+};
+
+export const getPropertyExpressionState = (
```
MauricioUyaguari, line 447: can we change the name of this to what exactly the function is doing to the property expression state

packages/legend-query/src/stores/QueryBuilderLambdaProcessor.ts
```diff
@@ -387,20 +398,28 @@ export class QueryBuilderLambdaProcessor
...
+          );
+          this.queryBuilderState.querySetupState.setBusinessDate(
+            valueSpecification.parametersValues[2],
+          );
+        } else if (stereotype === MILESTONING_STEROTYPES.PROCESSING_TEMPORAL) {
+          acceptedNoOfParameters = 2;
+          assertTrue(
+            valueSpecification.parametersValues.length ===
+              acceptedNoOfParameters,
+            `Can't process getAll() expression: when used with a milestoned class getAll() expects a parameter of type 'Date'`,
```
MauricioUyaguari, line 419: The assert message doesn't match the assert. In the assert you check the length of the parameters not the type.

packages/legend-query/src/stores/QueryBuilderMilestoningHelper.ts
MauricioUyaguari: Seems like we may be able to do something similar to what we do in model to model mapping. We `decorate` the derived property when the date is propogated and we clean after the editor is closed.

MauricioUyaguari (changes requested):
A lot of logic to follow. Could we add some integration tests please. Would be good to have tests for the most common workflows in the videos.

packages/legend-graph/src/models/protocols/pure/v1/transformation/pureGraph/to/helpers/V1_MilestoneBuilderHelper.ts
```diff
@@ -0,0 +1,192 @@
...
+            ),
+            property.owner,
+          );
+          dateProperty.parameters = [
+            buildMilestoningParameter(
+              DEFAULT_MILESTONING_PARAMETERS.BUSINESS_DATE,
+              graph,
+            ),
+          ];
+          const milestonedAllVersions = new Property(
```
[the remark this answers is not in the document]
gayathrir11: No they won't show up in the explorer tree

packages/legend-graph/src/models/protocols/pure/v1/transformation/pureGraph/to/helpers/V1_MilestoneBuilderHelper.ts
```diff
@@ -0,0 +1,192 @@
...
+            buildMilestoningParameter(
+              DEFAULT_MILESTONING_PARAMETERS.PROCESSING_DATE,
+              graph,
+            ),
+            buildMilestoningParameter(
+              DEFAULT_MILESTONING_PARAMETERS.BUSINESS_DATE,
+              graph,
+            ),
+          ];
+          const milestonedAllVersions = new Property(
```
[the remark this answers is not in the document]
gayathrir11: no

packages/legend-query/src/stores/QueryBuilderSetupState.ts
```diff
@@ -46,12 +48,16 @@ export class QueryBuilderSetupState {
   constructor(queryBuilderState: QueryBuilderState) {
     makeAutoObservable(this, {
       queryBuilderState: false,
-      addClassMilestoningTemporalValues: action,
+      processingDate: observable,
```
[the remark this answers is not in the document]
gayathrir11: will modify that

akphi:
@gayathrir11 let's rebase 😛

packages/legend-query/src/components/__tests__/QueryBuilder_FetchStructure.test.tsx
```diff
@@ -487,3 +489,591 @@ test(
   },
   // TODO: add more test when we rework the graph fetch tree
 );
+
+test(
```
akphi, line 493: Can we move these test to a separate file, something like `QueryBuilder_Milestoning.test.tsx` and can we also do `Business Temporal` and `Processing Temporal` in the test name instead of `businesstemporal`? Thanks!

packages/legend-query/src/components/QueryBuilderValueSpecificationEditor.tsx
```diff
@@ -557,15 +562,15 @@ export const DateInstanceValueEditor = observer(
         valueSpecification.genericType.value.rawType.name !==
         PRIMITIVE_TYPE.LATESTDATE
       ) {
-        valueSpecification.values = [
+        valueSpecification.addValue(
```
akphi, line 565: please rebase so you can see how we handle this kind of modification now

gayathrir11:
> @gayathrir11 let's rebase 😛

Yes.

packages/legend-query/src/stores/QueryBuilderFilterState.ts
```diff
@@ -139,6 +140,12 @@ export class FilterConditionState {
     );
   }
 
+  get propertyExpression(): QueryBuilderPropertyExpressionState {
```
akphi, line 143: why do we do this? people can use `propertyExpression` from `this.propertyExpressionState.propertyExpression`. Doing something like this is confusing. If you meant to override the `propertyExpressionState` with a version that is equiped with `Milestoning` stuffs, you can probably do it right in the constructor.

packages/legend-graph/src/MetaModelConst.ts
```diff
@@ -320,3 +320,16 @@ export enum MILESTONING_STEROTYPES {
 }
 
 export const LATEST_DATE = '%latest';
+
+export enum DEFAULT_MILESTONING_PARAMETERS {
```
akphi, line 324: Let's split these into `DEFAULT_PROCESSING_DATE_MILESTONING_PARAMETER_NAME` and `DEFAULT_BUSINESS_DATE_MILESTONING_PARAMETER_NAME`

akphi: Also, please do me a favor and rename `MILESTONING_STEROTYPES` to `MILESTONING_STEROTYPE`

packages/legend-studio/src/stores/shared/__tests__/MockDataUtil.test.ts
```diff
@@ -20,13 +20,14 @@ import { unitTest } from '@finos/legend-shared';
...
 import type { Entity } from '@finos/legend-model-storage';
+import { TEST__buildGraphWithEntities } from '@finos/legend-graph';
 
 const editorStore = TEST__getTestEditorStore();
 
 beforeAll(async () => {
   await flowResult(
-    editorStore.graphManagerState.graphManager.buildGraph(
-      editorStore.graphManagerState.graph,
+    TEST__buildGraphWithEntities(
```
akphi, line 29: why are we doing this?

packages/legend-query/src/stores/QueryBuilderLambdaProcessor.ts
```diff
@@ -971,6 +987,10 @@ export class QueryBuilderLambdaProcessor
       let currentPropertyExpression: ValueSpecification = valueSpecification;
       while (currentPropertyExpression instanceof AbstractPropertyExpression) {
         const propertyExpression = currentPropertyExpression;
+        isValidMilestoningLambda(
```
akphi, line 990: I think the naming here is confusing. It gives me the impression that this is a checker, but looking at it, it returns `void` so I know the function is not pure, i.e. it modifies something, so I have to jump to that bit of the code and check. Would be great if we name these kinds of function more accurately (preferably with inline comment and function comment) to lessen the number of jump around we have to make while reading this code

packages/legend-query/src/stores/QueryBuilderMilestoningHelper.ts
```diff
@@ -0,0 +1,615 @@
...
+      derivedPropertyExpressionState.processingDate =
+        derivedPropertyExpressionState.queryBuilderState.querySetupState.ProcessingDate;
+      return parameter;
+    }
+    default:
+      return undefined;
+  }
+};
+
+export const isValidMilestoningLambda = (
```
akphi, line 281: Maybe, let's rename this to `processMilestoning...` is `Lambda` the right word here? Also, if you name the parameter `propertyExpression`, could we narrow down the type somehow? Due to its awkwardness to name and define this function, I think maybe instead of calling this on `propertyExpression.parametersValues[0]`, why don't we call it on `propertyExpression` itself, that way, we can name this function something like `processMilestoningPropertyExpression` instead. What do you think?

packages/legend-query/src/stores/QueryBuilderLambdaBuilder.ts
```diff
@@ -240,21 +241,47 @@ export const buildLambdaFunction = (
...
   const getAllFunction = buildGetAllFunction(_class, multiplicityOne);
-  if (
-    getMilestoneTemporalStereotype(
-      _class,
-      queryBuilderState.graphManagerState.graph,
-    )
-  ) {
-    queryBuilderState.querySetupState.classMilestoningTemporalValues.forEach(
-      (parameter) =>
+  const stereotype = getMilestoneTemporalStereotype(
```
akphi, line 244: --> milestoningStereotype

akphi: Let's add a blank line and a comment

```ts

// build milestoning parameter(s) for getAll()
```

packages/legend-query/src/stores/QueryBuilderProjectionState.ts
```diff
@@ -181,6 +182,12 @@ export class QueryBuilderSimpleProjectionColumnState extends QueryBuilderProject
     );
   }
 
+  get propertyExpression(): QueryBuilderPropertyExpressionState {
```
akphi, line 185: As said, this is the kind of thing we try to avoid. When you add something like this, especially when you don't have any documentation nor some intense alerting name like `DONTUSEME__getPropertyExpression`, you are creating yet another thing people can accidentally use without really understanding about it.

akphi (changes requested):
First round of comments. I will move on to the main changes tomorrow.

packages/legend-graph/src/models/metamodels/pure/packageableElements/domain/Association.ts
```diff
@@ -79,6 +79,12 @@ export class Association
   taggedValues: TaggedValue[] = [];
   derivedProperties: DerivedProperty[] = [];
 
+  /**
+   * To store the properties generated while processing the milestoning properties. The properties
+   * generated are `allVersions`, `allVersionsInRange` and derived property with date parameter.
+   */
+  _generatedMilestonedProperties: Property[] = [];
```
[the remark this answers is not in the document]
gayathrir11: No it doesn't correspond to `originalMilestonedProperties` in `legend-engine`. Will modify it to `AbstractProperty`

packages/legend-query/src/stores/QueryBuilderMilestoningHelper.ts
gayathrir11: Makes more sense

packages/legend-studio/src/stores/shared/__tests__/MockDataUtil.test.ts
gayathrir11: To build `SystemModels`. Otherwise it is unable to find `temporal profile` in the graph and throwing an error

packages/legend-query/src/stores/QueryBuilderMilestoningHelper.ts
```diff
@@ -0,0 +1,641 @@
...
+    const targetStereotype = getMilestoneTemporalStereotype(
+      property.genericType.value.rawType,
+      graph,
+    );
+    return sourceStereotype === targetStereotype;
+  }
+  return false;
+};
+
+export const removePropagatedDates = (
```
akphi, line 125: Let's add documentation to this helper method. Also, I need to check with you why we're doing this. In my mind, we can leave all of this as is if we use `propagated value` for these parameter. Now, we only need to `prune` these when we `build the lambda` for persistence/execution.

packages/legend-query/src/stores/QueryBuilderPropertyEditorState.ts
```diff
@@ -349,5 +456,8 @@ export class QueryBuilderPropertyExpressionState {
     }
     this.requiresExistsHandling = requiresExistsHandling;
     this.derivedPropertyExpressionStates = result.slice().reverse();
+    if (this.derivedPropertyExpressionStates.length) {
```
akphi, line 459: This is strange to me, see my comment for `removePropagatedDates`

packages/legend-query/src/stores/QueryBuilderPropertyEditorState.ts
```diff
@@ -185,17 +268,19 @@ const fillDerivedPropertyArguments = (
...
 };
 
 export class QueryBuilderDerivedPropertyExpressionState {
   queryBuilderState: QueryBuilderState;
   path: string;
   title: string;
-  readonly derivedProperty: DerivedProperty;
-  readonly propertyExpression!: AbstractPropertyExpression;
-  readonly parameters: VariableExpression[] = [];
+  propertyExpression: AbstractPropertyExpression;
```
akphi, line 279: why are we removing `readonly` for these? They are fixed like this purposely so we control how they got modified.

packages/legend-query/src/stores/QueryBuilderPropertyEditorState.ts
```diff
@@ -185,17 +268,19 @@ const fillDerivedPropertyArguments = (
...
+  derivedProperty: DerivedProperty;
+  parameters: VariableExpression[] = [];
+  processingDate: ValueSpecification | undefined;
```
akphi, line 282: why do we have a state here for these as well? I thought we control them via `QueryBuilderSetupState`

packages/legend-query/src/stores/QueryBuilderMilestoningHelper.ts
```diff
@@ -0,0 +1,641 @@
+/**
```
akphi, line 1: Honestly, I haven't had time to review fully the logic here yet, I'll come back later.

packages/legend-query/src/stores/QueryBuilderMilestoningHelper.ts
```diff
@@ -0,0 +1,641 @@
...
+      );
+    }
+  }
+  return undefined;
+};
+
+//sourceStereotype: Stereotype of source class of the previous property expression/ class which we are querying from if it is the
+// first expression getting processed.
+//targetStereotype: Stereotype of source class of current property expression.
+export const isDefaultDatePropagationSupported = (
```
akphi, line 100: Could we document this method better? The logic here is obscure to me and the parameters are not well-documented

packages/legend-query/src/stores/QueryBuilderMilestoningHelper.ts
```diff
@@ -0,0 +1,641 @@
...
+    case MILESTONING_STEROTYPE.PROCESSING_TEMPORAL:
+      return queryBuilderState.querySetupState.processingDate;
+    default:
+      return undefined;
+  }
+};
+
+// To populate default dates to the next level in the property chain if the date values of the current
+// level are changed when date propagation is supported to the next level.
+export const propagateDefaultDates = (
```
akphi, line 377: Let's dicuss about this. I wonder if this would be changed when we move to use `PropagatedMilestoningParameterValue` type instead

akphi (changes requested):

packages/legend-query/src/stores/QueryBuilderPropertyEditorState.ts
gayathrir11: that wasn't on purpose. Will add them back

packages/legend-graph/src/models/metamodels/pure/valueSpecification/PropagatedMilestoningParameterValue.ts
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+
+import {
+  ValueSpecification,
+  type ValueSpecificationVisitor,
+} from './ValueSpecification';
+
+export class PropagatedMilestoningParameterValue extends ValueSpecification {
+  isDefaultDatePropagationSupported!: boolean;
```
akphi, line 23: I kinda want this to be more like `PropagatedValue extends ValueSpecification`, but let's see.

akphi: Also, we have a convention for this kind of stuff to prefix it with `INTERNAL__`

packages/legend-query/src/components/QueryBuilderValueSpecificationEditor.tsx
```diff
@@ -696,6 +697,16 @@ export const QueryBuilderValueSpecificationEditor: React.FC<{
         className={className}
       />
     );
+  } else if (
+    valueSpecification instanceof PropagatedMilestoningParameterValue
+  ) {
+    return (
+      <QueryBuilderValueSpecificationEditor
```
akphi, line 704: yep, bravo :tada:

packages/legend-query/src/components/QueryBuilderPropertyExpressionEditor.tsx
```diff
@@ -166,7 +184,13 @@ const DerivedPropertyExpressionEditor = observer(
...
           {derivedPropertyExpressionState.title}
         </div>
-        {!parameterValues.length && (
+        {(isDefaultDatePropagationSupported(
+          derivedPropertyExpressionState,
+          derivedPropertyExpressionState.queryBuilderState.graphManagerState
+            .graph,
+        )
+          ? !derivedPropertyExpressionState.parameters.length
+          : !parameterValues.length) && (
```
akphi, line 193: Doesn't look right here, if we already filled the slots with `propagated values`, we should only care about `parameterValues.length`

packages/legend-query/src/stores/QueryBuilderSetupState.ts
```diff
@@ -31,12 +30,17 @@ import {
...
+  DEFAULT_PROCESSING_DATE_MILESTONING_PARAMETER_NAME,
+  MILESTONING_STEROTYPE,
+  observe_ValueSpecification,
 } from '@finos/legend-graph';
 
 export class QueryBuilderSetupState {
   queryBuilderState: QueryBuilderState;
   _class?: Class | undefined;
-  classMilestoningTemporalValues: ValueSpecification[] = [];
+  _businessDate: ValueSpecification | undefined;
```
akphi, line 42: Let's move these to the bottom of the list of attributes and put a `TODO` to indicate we're working towards changing this

packages/legend-query/src/stores/QueryBuilderPropertyEditorState.ts
```diff
@@ -241,7 +359,12 @@ export class QueryBuilderDerivedPropertyExpressionState {
       if (paramValue instanceof InstanceValue) {
         const isRequired = paramValue.multiplicity.lowerBound >= 1;
         // required and no values provided
-        if (isRequired && !paramValue.values.length) {
+        if (
+          isRequired &&
+          paramValue.genericType?.value.rawType.name !==
```
akphi, line 364: Let's add a comment about the strange property of `%latest` here.

packages/legend-query/src/stores/QueryBuilderPropertyEditorState.ts
```diff
@@ -321,6 +444,20 @@ export class QueryBuilderPropertyExpressionState {
       ) {
         requiresExistsHandling = true;
       }
+      if (
```
akphi, line 447: Let's add a comment here to explain that we're checking if the property is milestoned

packages/legend-query/src/stores/QueryBuilderPropertyEditorState.ts
```diff
@@ -115,68 +125,173 @@ export const getPropertyPath = (
...
+      );
+    }
   });
+  if (
+    temporalSource === MILESTONING_STEROTYPE.PROCESSING_TEMPORAL &&
+    temporalTarget === MILESTONING_STEROTYPE.BITEMPORAL &&
+    parameterValues.length
+  ) {
+    propertyArguments = [
+      guaranteeNonNullable(
```
akphi, line 288: why don't we use `querySetupState.processingDate` here?

packages/legend-query/src/stores/QueryBuilderPropertyEditorState.ts
```diff
@@ -115,68 +125,173 @@ export const getPropertyPath = (
...
+        derivedPropertyExpressionState,
+        temporalTarget,
+        idx,
+        isDatePropagationSupported,
+      );
+      if (milestoningParameter) {
+        propertyArguments.push(milestoningParameter);
+      }
+    } else {
+      let argument: ValueSpecification | undefined;
```
akphi, line 226: let's rebase, I think @YannanGao-gs changed this a bit

packages/legend-graph/src/models/metamodels/pure/valueSpecification/INTERNAL__PropagatedValue.ts
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+
+import {
+  ValueSpecification,
+  type ValueSpecificationVisitor,
+} from './ValueSpecification';
+
+export class INTERNAL__PropagatedValue extends ValueSpecification {
+  isDefaultDatePropagationSupported!: boolean;
```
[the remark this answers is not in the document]
akphi: After the discussion, I think we can safely remove this flag now. We can check if we can `pop` or `remove` the propagated value based on its context rather than some flags embedded in it. This will keep this type of value spec generic. 

Now in the case of milestoning, when we cleanup the bubbles as we build the value spec from the state, we can have a centralized logic to determine if we can pop that bubble, and the logic here follows the chart you kindly provided. The logic within that function is something we will focus on reviewing.

packages/legend-query/src/components/QueryBuilderSetupPanel.tsx
```diff
@@ -79,21 +79,21 @@ const generateClassLabel = (
   if (milestoneStereotype) {
     let milestoningParameterValues;
     switch (milestoneStereotype) {
-      case MILESTONING_STEROTYPES.BUSINESS_TEMPORAL:
+      case MILESTONING_STEROTYPE.BUSINESS_TEMPORAL:
         milestoningParameterValues = `Business Date: ${getParameterValue(
-          queryBuilderState.querySetupState.classMilestoningTemporalValues[0],
+          queryBuilderState.querySetupState._businessDate,
```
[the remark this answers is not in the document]
gayathrir11: We show that icon only when the class is milestoned

packages/legend-query/src/stores/QueryBuilderPropertyEditorState.ts
```diff
@@ -178,23 +187,94 @@ export const generateValueSpecificationForParameter = (
...
 };
 
-const fillDerivedPropertyArguments = (
+export const fillDerivedPropertyArguments = (
   derivedPropertyExpressionState: QueryBuilderDerivedPropertyExpressionState,
-  queryBuilderState: QueryBuilderState,
+  graph: PureModel,
 ): void => {
-  const propertyArguments: ValueSpecification[] =
+  let propertyArguments: ValueSpecification[] =
```
akphi, line 194: As mentioned in the comment about `fillMilestonedDerivedPropertyArguments`, let's try to better isolate this logic.

packages/legend-query/src/stores/QueryBuilderValueSpecificationCloneHelper.ts
```diff
@@ -0,0 +1,196 @@
...
+  RuntimeInstanceValue,
+  SimpleFunctionExpression,
+  TYPICAL_MULTIPLICITY_TYPE,
+  ValueSpecification,
+  ValueSpecificationVisitor,
+  VariableExpression,
+} from '@finos/legend-graph';
+import { UnsupportedOperationError } from '@finos/legend-shared';
+
+export class QueryBuilderValueSpecificionCloneHelper
```
akphi, line 47: This is fair, but we don't need to do a full clone right? If we do, we'd better just turn it into `V1_protocol` and then build it back rather than having to do this

packages/legend-query/src/stores/QueryBuilderMilestoningHelper.ts
```diff
@@ -0,0 +1,641 @@
...
+  DEFAULT_BUSINESS_DATE_MILESTONING_PARAMETER_NAME,
+} from '@finos/legend-graph';
+import {
+  fillDerivedPropertyArguments,
+  type QueryBuilderPropertyExpressionState,
+  type QueryBuilderDerivedPropertyExpressionState,
+} from './QueryBuilderPropertyEditorState';
+import { functionExpression_setParametersValues } from './QueryBuilderValueSpecificationModifierHelper';
+
+export const valueSpecifiation_isEqual = (
```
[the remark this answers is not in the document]
akphi: Hm, let's

packages/legend-query/src/stores/QueryBuilderMilestoningHelper.ts
```diff
@@ -0,0 +1,447 @@
...
+import type { QueryBuilderSetupState } from './QueryBuilderSetupState';
+import {
+  functionExpression_setParametersValues,
+  propertyExpression_setParametersValue,
+} from './QueryBuilderValueSpecificationModifierHelper';
+import type { QueryBuilderState } from './QueryBuilderState';
+import { QueryBuilderValueSpecificionCloneHelper } from './QueryBuilderValueSpecificationCloneHelper';
+
+export const valueSpecifiation_isEqual = (
+  param1: ValueSpecification | undefined,
```
akphi, line 50: Let's reason about the usefulness of this method in the final round of review. For now it can be left here.

akphi (changes requested):

packages/legend-graph/src/models/metamodels/pure/valueSpecification/INTERNAL__PropagatedValue.ts
```diff
@@ -0,0 +1,31 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {
+  ValueSpecification,
+  type ValueSpecificationVisitor,
+} from './ValueSpecification';
+
+// Used to specify the type of default parameterValues studio propagates for milestoned properties.
```
akphi: Let's use block comment here

```ts
/**
 * This value specification maintains a reference to another value specification
 * 
 * @internal This type is specific to Studio only, not a standard, recognizeable in Pure/engine.
 */
```

akphi (changes requested):
sorry, looks like this is not ready for review yet

packages/legend-query/src/stores/QueryBuilderLambdaBuilder.ts
```diff
@@ -209,6 +221,66 @@ export const processPostFilterOnLambda = (
   return lambda;
 };
 
+export const buildPropertyExpressionChain = (
```
akphi, line 224: It's safe to make the param `propertyExpression` of type `AbstractPropertyExpression`. And as such, we can make this method return `AbstractPropertyExpression` and eliminate the awkward check and `undefined` return.

packages/legend-query/src/stores/QueryBuilderLambdaBuilder.ts
```diff
@@ -209,6 +221,66 @@ export const processPostFilterOnLambda = (
...
+  propertyExpression: ValueSpecification,
+  queryBuilderState: QueryBuilderState,
+): ValueSpecification | undefined => {
+  if (propertyExpression instanceof AbstractPropertyExpression) {
+    const newPropertyExpression =
+      propertyExpression.accept_ValueSpecificationVisitor(
+        new QueryBuilderValueSpecificionCloneHelper(
```
akphi, line 231: As discussed, let's use a cheap copy strategy here. Since you we know this is just the property expression chain. My take is we can just traverse the chain and copy anything but property expression over. Property expressions are the only thing you need to `new` here. But I might be wrong as the situation can get complicated with `subtyping`, I'm not 100% sure about that.

packages/legend-query/src/stores/QueryBuilderMilestoningHelper.ts
akphi: We don't need to `export` this

packages/legend-query/src/stores/QueryBuilderValueSpecificationCloneHelper.ts
akphi: As you said, there's a loophole here with the roundtrip approach since we don't deal with `INTERNAL__PropagatedValue` there. But I still think we should investigate if there's a less overkill way to clone.

packages/legend-query/src/stores/QueryBuilderPropertyEditorState.ts
```diff
@@ -178,22 +285,71 @@ export const generateValueSpecificationForParameter = (
...
+    temporalTarget = getMilestoneTemporalStereotype(
+      derivedPropertyExpressionState.propertyExpression.func.genericType.value
+        .rawType,
+      derivedPropertyExpressionState.queryBuilderState.graphManagerState.graph,
+    );
+  }
   derivedPropertyExpressionState.parameters.forEach((parameter, idx) => {
     if (idx < derivedPropertyExpressionState.parameterValues.length) {
+      if (temporalTarget) {
+        const milestoningParameter = fillMilestonedDerivedPropertyArguments(
```
akphi, line 310: Let's cleanup all of these milestoning-related logic and put them in `fillMilestonedDerivedPropertyArguments`. My thinking here is given a derived property parameter, we produce a value, so I wold expect something like


```ts
const parameterValue = generateMilestonedPropertyParameterValue(...)
  ?? enerateValueSpecificationForParameter(
      parameter,
        derivedPropertyExpressionState.queryBuilderState.graphManagerState
          .graph,
    );
```

packages/legend-query/src/stores/QueryBuilderPropertyEditorState.ts
```diff
@@ -178,22 +285,71 @@ export const generateValueSpecificationForParameter = (
   );
 };
 
-const fillDerivedPropertyArguments = (
+export const fillDerivedPropertyArguments = (
```
akphi, line 288: Let's rename this method to `fillDerivedPropertyParameterValues`

packages/legend-query/src/stores/QueryBuilderPropertyEditorState.ts
```diff
@@ -178,22 +285,71 @@ export const generateValueSpecificationForParameter = (
...
   derivedPropertyExpressionState: QueryBuilderDerivedPropertyExpressionState,
-  queryBuilderState: QueryBuilderState,
+  graph: PureModel,
 ): void => {
   const propertyArguments: ValueSpecification[] =
```
akphi, line 292: --> `parameterValues`

packages/legend-query/src/stores/QueryBuilderPropertyEditorState.ts
```diff
@@ -39,21 +40,126 @@ import {
...
+} from './QueryBuilderMilestoningHelper';
 import { functionExpression_setParametersValues } from './QueryBuilderValueSpecificationModifierHelper';
 
 export const prettyPropertyName = (value: string): string =>
   isCamelCase(value) ? prettyCamelCase(value) : prettyCONSTName(value);
 
+/**
+ * Propagates default dates for milestoned properties when date propagation is not supported by engine.
+ */
+export const fillMilestonedDerivedPropertyArguments = (
```
akphi, line 71: --> `generateMilestonedPropertyParameterValue`

packages/legend-query/src/stores/QueryBuilderMilestoningHelper.ts
```diff
@@ -0,0 +1,207 @@
...
+      } else {
+        return false;
+      }
+    default:
+  }
+  return false;
+};
+
+/**
+ * sourceStereotype: Stereotype of source class of current property expression.
```
akphi, line 162: the name of the method speaks for itself really, but I think we should move these comment to where we create these variables instead. As this is useful for code readability, not as documentation of the function's usage.

packages/legend-query/src/stores/QueryBuilderMilestoningHelper.ts
```diff
@@ -0,0 +1,207 @@
...
+ * targetStereotype: Stereotype (if exists) of genericType of current property expression.
+ */
+export const isDefaultDatePropagationSupported = (
+  propertyExpression: AbstractPropertyExpression,
```
akphi, line 166: --> `currentPropertyExpression`

packages/legend-query/src/stores/QueryBuilderMilestoningHelper.ts
```diff
@@ -0,0 +1,207 @@
...
+  queryBuilderState: QueryBuilderState,
```
akphi, line 167: Let's reorder the params, `queryBuilderState` can go last

packages/legend-query/src/stores/QueryBuilderMilestoningHelper.ts
```diff
@@ -0,0 +1,207 @@
+/**
+ * Copyright (c) 2020-present, Goldman Sachs
```
akphi, line 2: I think we can eliminate this file and move all of these methods to `QueryBuilderPropertyEditorState`

packages/legend-query/src/stores/QueryBuilderMilestoningHelper.ts
akphi: I think it's safe to inline this method in `checkForMilestoningParameterEquality` and rename it to `checkIfEquivalent`

packages/legend-query/src/stores/QueryBuilderPropertyEditorState.ts
```diff
@@ -178,22 +285,71 @@ export const generateValueSpecificationForParameter = (
   );
 };
 
-const fillDerivedPropertyArguments = (
+export const fillDerivedPropertyArguments = (
   derivedPropertyExpressionState: QueryBuilderDerivedPropertyExpressionState,
-  queryBuilderState: QueryBuilderState,
+  graph: PureModel,
```
akphi, line 290: if we don't need `graph` anymore, let's remove it.

packages/legend-query/src/stores/QueryBuilderMilestoningHelper.ts
```diff
@@ -0,0 +1,207 @@
...
+      return getMilestoneTemporalStereotype(
+        guaranteeType(ownerClass, Class),
+        graph,
+      );
+    }
+  }
+  return undefined;
+};
+
+export const checkForOverridingMilestoningQuery = (
```
akphi, line 85: Let's document this method

packages/legend-query/src/stores/QueryBuilderMilestoningHelper.ts
```diff
@@ -0,0 +1,207 @@
...
+      param2.genericType.value.rawType.name
+    ) {
+      return param1.values[0] === param2.values[0];
+    }
+    return false;
+  }
+  return false;
+};
+
+export const getSourceTemporalStereotype = (
```
akphi, line 65: why not just call it `getDerivedPropertyMilestoningSteoreotype`?

packages/legend-query/src/stores/QueryBuilderMilestoningHelper.ts
```diff
@@ -0,0 +1,207 @@
...
+    derivedPropertyExpressionState.parameterValues.length === 1
+  ) {
+    return true;
+  } else {
+    return false;
+  }
+};
+
+/**
+ * Checks if the parameter of the current milestoned property is same as `businessDate` or `processingDate`
```
akphi, line 105: --> 
```ts
/**
 * Check if the parameter value of the milestoned property is
 * the same as those specified in global scope, so that we can
 * potentially replace them with propgated value.
 */
```

packages/legend-query/src/stores/QueryBuilderPropertyEditorState.ts
```diff
@@ -178,22 +285,71 @@ export const generateValueSpecificationForParameter = (
...
   derivedPropertyExpressionState.parameters.forEach((parameter, idx) => {
     if (idx < derivedPropertyExpressionState.parameterValues.length) {
+      if (temporalTarget) {
+        const milestoningParameter = fillMilestonedDerivedPropertyArguments(
+          derivedPropertyExpressionState,
+          temporalTarget,
+          idx,
+        );
+        // Changes the `valueSpecification` to `INTERNAL__PropagatedValue` only if the parameter value
+        // matches with the corresponding `businessDate` or `processingDate`
```
akphi, line 316: Let's also add a `NOTE: ` here to say that this will rewrite provided query as in if people explicitly specify the values, we will remove them.

packages/legend-query/src/stores/QueryBuilderValueSpecificationCloneHelper.ts
akphi: We can use a while loop to walk the property expression chain. Remember, we can also rewrite a visitor to a flat method :D, especially when we default out most of the methods of the visitor. Also, we got a bug here, since we have to deal with `subtyping`, i.e. the `SimpleFunctionExpression` case. So my take is we really have to do the `while` loop to walk the property expression chain.

akphi (changes requested):

packages/legend-query/src/stores/QueryBuilderMilestoningHelper.ts
gayathrir11: third parameter si optional so cannot reorder

packages/legend-query/src/stores/QueryBuilderLambdaBuilder.ts
```diff
@@ -323,12 +443,13 @@ export const buildLambdaFunction = (
...
+              projectionColumnState.propertyExpressionState.propertyExpression,
+              projectionColumnState.propertyExpressionState.queryBuilderState,
+            );
             columnLambda = buildGenericLambdaFunctionInstanceValue(
               projectionColumnState.lambdaParameterName,
-              [
-                projectionColumnState.propertyExpressionState
-                  .propertyExpression,
-              ],
+              [guaranteeNonNullable(propertyExpression)],
```
akphi, line 452: we don't need to call `guaranteeNonNullable` anymore.

packages/legend-query/src/stores/QueryBuilderLambdaBuilder.ts
```diff
@@ -428,12 +549,13 @@ export const buildLambdaFunction = (
...
```
akphi, line 558: we don't need to call `guaranteeNonNullable` anymore

packages/legend-query/src/stores/QueryBuilderPropertyEditorState.ts
```diff
@@ -39,21 +40,339 @@ import {
...
+    return sourceStereotype === targetStereotype;
+  }
+  return false;
+};
 
 export const prettyPropertyName = (value: string): string =>
   isCamelCase(value) ? prettyCamelCase(value) : prettyCONSTName(value);
 
+/**
+ * Propagates default dates for milestoned properties when date propagation is not supported by engine.
```
akphi, line 231: Rewrite this comment

packages/legend-query/src/stores/QueryBuilderPropertyEditorState.ts
```diff
@@ -39,21 +40,339 @@ import {
...
+        return true;
+      } else {
+        return false;
+      }
+    default:
+  }
+  return false;
+};
+
+export const isDefaultDatePropagationSupported = (
```
akphi, line 180: Document this method

akphi (approved):
Left some minor comments for cleaning up, majorly about documentation of things. I will merge this now. Thanks @gayathrir11 for tackling this painful challenge 🎉

MauricioUyaguari (approved):

packages/legend-query/src/components/QueryBuilderPropertyExpressionEditor.tsx
```diff
@@ -84,6 +86,14 @@ const DerivedPropertyParameterValueEditor = observer(
       [handleDrop],
     );
     const resetParameterValue = (): void => {
+      if (
```
akphi, line 89: Looking back, this seems wrong to me, this is the reset button, so it should only reset 1 parameter value rather than resetting all of them as you call `fillDerivedPropertyParameterValues`. I think what we should do instead is to use the similar logic we have in `fillDerivedPropertyParameterValues` `forEach` block

packages/legend-query/src/stores/QueryBuilderPropertyEditorState.ts
akphi, line 180: Let's document this method. And I meant a serious doc since this method seems to capture a fairly complex logic which you use diagram to explain in your PR. It might be a good idea to try to chart this out similar to what we did in `ChangeDetectionState`

packages/legend-query/src/stores/QueryBuilderPropertyEditorState.ts
```diff
@@ -39,21 +40,339 @@ import {
...
+      return false;
+    }
+  }
+  if (property.genericType.value.rawType instanceof Class) {
+    // sourceStereotype: Stereotype of source class of current property expression.
+    const sourceStereotype =
+      property instanceof DerivedProperty
+        ? getDerivedPropertyMilestoningSteoreotype(property, graph)
+        : undefined;
+    if (sourceStereotype === MILESTONING_STEREOTYPE.BITEMPORAL) {
```
akphi, line 213: Could we explain why we do this shortcut?

packages/legend-query/src/stores/QueryBuilderPropertyEditorState.ts
```diff
@@ -39,21 +40,339 @@ import {
...
+};
+
+export const isDefaultDatePropagationSupported = (
+  currentPropertyExpression: AbstractPropertyExpression,
+  queryBuilderState: QueryBuilderState,
+  prevPropertyExpression?: AbstractPropertyExpression | undefined,
+): boolean => {
+  const property = currentPropertyExpression.func;
+  const graph = queryBuilderState.graphManagerState.graph;
+  if (
```
akphi, line 187: Let's add a comment to explain the purpose of this block

packages/legend-query/src/stores/QueryBuilderPropertyEditorState.ts
```diff
@@ -39,21 +40,339 @@ import {
...
+    if (sourceStereotype === MILESTONING_STEREOTYPE.BITEMPORAL) {
+      return true;
+    }
+
+    // targetStereotype: Stereotype (if exists) of genericType of current property expression.
+    const targetStereotype = getMilestoneTemporalStereotype(
+      property.genericType.value.rawType,
+      graph,
+    );
+    return sourceStereotype === targetStereotype;
```
akphi, line 222: Let's add a comment here to explain this

akphi (commented):
@gayathrir11 I attempted to do some cleanup in https://github.com/finos/legend-studio/pull/1123

Please test and let me know if I messed up anything, the main thing I would anticipate any breakage is `resetParameterValue` method in `DerivedPropertyParameterValueEditor`.

I need you to do a little more docs, especially for some of the logic like `generateMilestonedPropertyParameterValue` and `isDefaultDatePropagationSupported`. I think it would be great if we could document the logic in detail in the code itself rather than having to jump to some PR to get the idea. Be creative, like drawing tables/diagram in ASCII.

Thanks!

packages/legend-query/src/stores/QueryBuilderPropertyEditorState.ts
akphi, line 222: I think there might a bug here also since we could have `sourceStereotype === undefined && targetStereotype === undefined` and that returns `true`

PR #909, opened 2022-03-03
Add support for DateTime valuespecification editor

### LINKED ISSUE #842: Bug: DateTime value specification editor produces StrictDate type causing class cast exception
### Similar issues

### How are you using Studio?

Legend Studio

A property of type `DateTime` should produce a text input for users to type out the dateTime. This is both for filter and parameter instance editor
Currently we show a `StrictDate` value editor for properties of `DateTime`. This causes a class exception when executing. 

### Steps to reproduce

_No response_

### Model data

_No response_

### Environment

```text
Chrome
```

_No response_

### Contribution

gayathrir11:
## Summary

Closes #842 

- [ ] Test(s) added

### DIFF
--- a/.changeset/slimy-spiders-hope.md
+++ b/.changeset/slimy-spiders-hope.md
@@ -1,0 +1,5 @@
+---
+"@finos/legend-query": patch
+---
+
+Add support for `DateTime` valueSpecification editor
--- a/packages/legend-query/src/components/QueryBuilderValueSpecificationEditor.tsx
+++ b/packages/legend-query/src/components/QueryBuilderValueSpecificationEditor.tsx
@@ -250,6 +250,31 @@ export const DatePrimitiveInstanceValueEditor = observer(
   },
 );
 
+export const DateTimePrimitiveInstanceValueEditor = observer(
+  (props: {
+    valueSpecification: PrimitiveInstanceValue;
+    className?: string | undefined;
+  }) => {
+    const { valueSpecification, className } = props;
+    const value = valueSpecification.values[0] as string;
+    const changeValue: React.ChangeEventHandler<HTMLInputElement> = (event) => {
+      valueSpecification.changeValue(event.target.value, 0);
+    };
+
+    return (
+      <div className={clsx('query-builder-value-spec-editor', className)}>
+        <input
+          className="panel__content__form__section__input query-builder-value-spec-editor__input"
+          type="datetime-local"
+          spellCheck={false}
+          value={value}
+          onChange={changeValue}
+        />
+      </div>
+    );
+  },
+);
+
 const EnumValueInstanceValueEditor = observer(
   (props: {
     valueSpecification: EnumValueInstanceValue;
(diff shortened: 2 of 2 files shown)


packages/legend-query/src/components/QueryBuilderValueSpecificationEditor.tsx
```diff
@@ -250,6 +250,31 @@ export const DatePrimitiveInstanceValueEditor = observer(
...
+    const value = valueSpecification.values[0] as string;
+    const changeValue: React.ChangeEventHandler<HTMLInputElement> = (event) => {
+      valueSpecification.changeValue(event.target.value, 0);
+    };
+
+    return (
+      <div className={clsx('query-builder-value-spec-editor', className)}>
+        <input
+          className="panel__content__form__section__input query-builder-value-spec-editor__input"
+          type="datetime-local"
```
akphi, line 268: @MauricioUyaguari what should it be here? `local` or `UTC`?

.changeset/slimy-spiders-hope.md
```diff
@@ -0,0 +1,5 @@
+---
+"@finos/legend-query": patch
+---
+
+Add support for `DateTime` valueSpecification editor
```
akphi, line 5:
```suggestion
Add support for `DateTime` value specification editor ([#842](https://github.com/finos/legend-studio/issues/842))
```

akphi (changes requested):

akphi:
@gayathrir11 Let's add a quick video for this. Thanks!

gayathrir11:
> @gayathrir11 Let's add a quick video for

> @gayathrir11 Let's add a quick video for this. Thanks!

Screenshot 2022-03-03 172911
 It looks like this. I don't like it.  Thinking of a better design

akphi:
> > @gayathrir11 Let's add a quick video for this. Thanks!
> 
> Screenshot 2022-03-03 172911
> 
> It looks like this. I don't like it. Looking for a better design

It's fine 😛 that's the default from browser. We don't need to ask for more now.

packages/legend-query/src/components/QueryBuilderValueSpecificationEditor.tsx
akphi: I think we should do `UTC`

akphi:
@gayathrir11 let's add video to the PR description about setting this and going in/out of text mode (kinda a roundtrip)

gayathrir11:
> @gayathrir11 let's add video to the PR description about setting this and going in/out of text mode (kinda a roundtrip)

Cool i will add video but the recording wouldn't show the editor actually and it's in the execute panel so i don't think there will be any changes to the query

akphi:
> > @gayathrir11 let's add video to the PR description about setting this and going in/out of text mode (kinda a roundtrip)
> 
> Cool i will add video but the recording wouldn't show the editor actually and it's in the execute panel so i don't think there will be any changes to the query so nothing will be actually visisble in text mode

Why though? We could show a video when you use this in filter?

packages/legend-query/src/components/QueryBuilderValueSpecificationEditor.tsx
akphi: Could we try this https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local#setting_timezones
so we set the timezone to UTC?

akphi (approved):
I'll test out this change myself, if it does not suffice, I can roll back.

packages/legend-query/src/components/QueryBuilderValueSpecificationEditor.tsx
akphi: Turns out `datetime-local` already assume `UTC` 😆 so we're good I guess

akphi:
Turns out the `datetime-local` actually worked with `UTC` by default - see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local#setting_timezones

We're good folks!

PR #911, opened 2022-03-04
Sort dependency project versions in descending order

YannanGao-gs:
## Summary
Sort dependency project versions in descending order

- [ ] Test(s) added

demo

### DIFF
--- a/.changeset/clever-forks-remember.md
+++ b/.changeset/clever-forks-remember.md
@@ -1,0 +1,5 @@
+---
+'@finos/legend-server-depot': patch
+---
+
+Add a utility function to compare semver versions.
--- a/.changeset/clever-forks-remember.md
+++ b/.changeset/clever-forks-remember.md
@@ -2,4 +2,4 @@
 '@finos/legend-server-depot': patch
 ---
 
-Add a utility function to compare semver versions.
+Add a utility function to compare `semver` versions.
--- a/packages/legend-studio/src/components/editor/edit-panel/project-configuration-editor/ProjectConfigurationEditor.tsx
+++ b/packages/legend-studio/src/components/editor/edit-panel/project-configuration-editor/ProjectConfigurationEditor.tsx
@@ -47,7 +47,7 @@ import {
   useApplicationStore,
 } from '@finos/legend-application';
 import { LEGEND_STUDIO_LOG_EVENT_TYPE } from '../../../../stores/LegendStudioLogEvent';
-import type { ProjectData } from '@finos/legend-server-depot';
+import { type ProjectData, compareVersions } from '@finos/legend-server-depot';
 
 interface VersionOption {
   label: string;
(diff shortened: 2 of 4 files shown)


packages/legend-studio/src/components/editor/edit-panel/project-configuration-editor/ProjectConfigurationEditor.tsx
```diff
@@ -230,7 +230,20 @@ const ProjectDependencyEditor = observer(
     // version
     const version = projectDependency.versionId;
     const versions = selectedProject?.versions ?? [];
-    const versionOptions = versions.map((v) => ({ value: v, label: v }));
+    const versionOptions = versions
```
akphi, line 233: could we look into what options we have with `npm`? This doesn't seem like the proper comparison to me. Also, if we cannot go with 3rd party library, perhaps, let's go with an utility function that we can unit test, I would be more confident then.

That being said, why don't we use what the backend return and just revert it, don't they sorta ensure the order for us in some way?

YannanGao-gs: this is what backend returns.
Screen Shot 2022-03-04 at 3 35 46 PM
e.g.
1.0 -> 11.2.0 -> 12.0.0 ->  2.1.0

akphi: funny how `2.0.0` shows up in the end there.
Anyway, I think we can use this https://www.npmjs.com/package/compare-versions

package.json
```diff
@@ -80,6 +80,9 @@
   "resolutions": {
     "glob-parent": "^6.0.1"
   },
+  "dependencies": {
```
akphi, line 83: this should be in `@finos/legend-studio`

packages/legend-studio/src/components/editor/edit-panel/project-configuration-editor/ProjectConfigurationEditor.tsx
```diff
@@ -230,7 +231,9 @@ const ProjectDependencyEditor = observer(
...
+      .map((v) => ({ value: v, label: v }))
+      .sort((v1, v2) => compareVersions(v2.value, v1.value));
```
akphi, line 236: probably can do 
```suggestion
      .sort(compareVersions)
      .map((v) => ({ value: v, label: v }))
```

packages/legend-studio/src/components/editor/edit-panel/project-configuration-editor/ProjectConfigurationEditor.tsx
akphi: Or we can do this

https://stackoverflow.com/questions/55466274/simplify-semver-version-compare-logic/55466325#55466325

https://github.com/omichelsen/compare-versions/issues/45

That's not bad. We can publish this as part of utilities in `DepotUtils`, we can set a note there linking to this library and this issue.

I can merge this and do that myself if you'd rather me doing that. Let me know!

akphi (changes requested):

.changeset/clever-forks-remember.md
```diff
@@ -0,0 +1,5 @@
+---
+'@finos/legend-server-depot': patch
+---
+
+Add a utility function to compare semver versions.
```
akphi, line 5:
```suggestion
Add a utility function to compare `semver` versions.
```

akphi (approved):

PR #963, opened 2022-03-17
support opening dependency project in configuration editor

MauricioUyaguari:

- [ ] Test(s) added


packages/legend-graph/src/models/protocols/pure/v1/engine/V1_EngineServerClient.ts
```diff
@@ -51,7 +51,7 @@ import type { EXECUTION_SERIALIZATION_FORMAT } from '../../../../../graphManager
 import type { V1_ExternalFormatDescription } from './externalFormat/V1_ExternalFormatDescription';
 import type { V1_ExternalFormatModelGenerationInput } from './externalFormat/V1_ExternalFormatModelGeneration';
 
-enum CORE_ENGINE_TRACER_SPAN {
+export enum V1_ENGINE_CLIENT_ACTION {
```
akphi, line 54: uhm, the responsibility is a bit different here though, one is in the form of `tracer span: transform Pure code to protocol` vs `log event: TRANSFORM_PURE_CODE_TO_PROTOCOL`. Let's revert this change for now

MauricioUyaguari: @akphi  you think it would be better to add a ENGINE_LOG_EVENT to that file instead ?

akphi: yeah, let's do that. We should not use this, the tracer really should only be here in the network client for now.

akphi (approved):

PR #975, opened 2022-03-22
Query setup

MauricioUyaguari:

- [ ] Test(s) added

### DIFF
--- a/.changeset/moody-plums-approve.md
+++ b/.changeset/moody-plums-approve.md
@@ -1,0 +1,5 @@
+---
+"@finos/legend-studio": patch
+---
+
+Enable `lambdaEditor` and allow DnD for returnType in FunctionEditor
--- a/packages/legend-studio/src/components/editor/edit-panel/FunctionEditor.tsx
+++ b/packages/legend-studio/src/components/editor/edit-panel/FunctionEditor.tsx
@@ -511,6 +511,27 @@ export const FunctionMainEditor = observer(
       }),
       [handleDropParameter],
     );
+    const handleDropReturnType = useCallback(
+      (item: UMLEditorElementDropTarget): void => {
+        if (!isReadOnly && item.data.packageableElement instanceof Type) {
+          functionElement.setReturnType(item.data.packageableElement);
+        }
+      },
+      [functionElement, isReadOnly],
+    );
+    const [{ isReturnTypeDragOver }, dropReturnTypeRef] = useDrop(
+      () => ({
+        accept: [
+          CORE_DND_TYPE.PROJECT_EXPLORER_CLASS,
+          CORE_DND_TYPE.PROJECT_EXPLORER_ENUMERATION,
+        ],
+        drop: (item: ElementDragSource): void => handleDropReturnType(item),
+        collect: (monitor): { isReturnTypeDragOver: boolean } => ({
+          isReturnTypeDragOver: monitor.isOver({ shallow: true }),
+        }),
+      }),
+      [handleDropReturnType],
+    );
 
     return (
       <div className="panel__content function-editor__element">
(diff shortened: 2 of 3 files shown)


packages/legend-graph/src/graphManager/AbstractPureGraphManager.ts
```diff
@@ -336,6 +336,21 @@ export abstract class AbstractPureGraphManager {
   abstract updateQuery(query: Query, graph: PureModel): Promise<Query>;
   abstract deleteQuery(queryId: string): Promise<void>;
 
+  //TODO: These methods will be removed once we have a way to `lazy` load elements into the graph
+  abstract buildCreateQueryLightGraph(
```
akphi, line 340: As discussed, I think the better way to do this is not to expose these kinds of customizations on the interface, but rather to create a method like `buildLightGraph` which allows the passing in a function to further build the graph after building the light graph. The `light graph` refers to a graph where only indexing is done.

packages/legend-graph/src/graphManager/GraphManagerLogEvent.ts
```diff
@@ -53,4 +53,7 @@ export enum GRAPH_MANAGER_LOG_EVENT {
   EXECUTION_FAILURE = 'EXECUTION_FAILURE',
   COMPILATION_FAILURE = 'COMPILATION_FAILURE',
   PARSING_FAILURE = 'PARSING_FAILURE',
+
+  QUERY_LIGHT_GRAPH_ENTITIES_WITH_DEPENDENCY_PARSE = 'QUERY_LIGHT_GRAPH_ENTITIES_WITH_DEPENDENCY_PARSE',
```
akphi, line 57: Not sure if we would need these anymore

packages/legend-graph/src/models/protocols/pure/v1/V1_PureGraphManager.ts
```diff
@@ -2119,6 +2134,243 @@ export class V1_PureGraphManager extends AbstractPureGraphManager {
     );
   }
 
+  async indexEntitiesWithDependencyIntoGraph<T extends V1_PackageableElement>(
+    graph: PureModel,
```
akphi, line 2138: I think we should pre-filter the elements to improve performance here.

As discussed, We could even tinker with the entities (JSON objects) to lessen the load for deserialization, maybe could you add a small comment about that? It's clearly premature optimization, but worth noting.

packages/legend-query/src/stores/LegendQueryStore.ts
```diff
@@ -491,33 +491,32 @@ export class LegendQueryStore {
 
     try {
       this.editorInitState.inProgress();
-      let queryInfoState: ServiceQueryInfoState;
+      let project: ProjectData;
```
akphi, line 494: --> `projectData`?

packages/legend-query/src/stores/LegendQueryStore.ts
```diff
@@ -771,7 +782,79 @@ export class LegendQueryStore {
...
+      this.applicationStore.log.error(
+        LogEvent.create(LEGEND_QUERY_LOG_EVENT_TYPE.QUERY_PROBLEM),
+        error,
+      );
+      this.applicationStore.notifyError(error);
+      this.buildGraphState.fail();
+    }
+  }
+
+  *buildLightCreateGraph(
```
akphi, line 797: --> `buildLightGraphForCreateQuerySetup`

akphi (changes requested):
One thing I would like us to check is memory leak, just base on the fact that we switch the graph in and out a few times. I even starting to think we could swap the graph out for light graph version. Could we do a quick check?

PR #976, opened 2022-03-23
Add Extra ModelLoader studio plugin

emilia-sokol-gs:
## Summary

Add extra model loader studio plugin to allow new model loader types to be added in more modularized fashion

The change will be tested when extra model loader plugin will be included

- [ ] Test(s) added
- [x]  No testing (please provide an explanation)

### DIFF
--- a/packages/legend-studio/src/components/editor/edit-panel/ModelLoader.tsx
+++ b/packages/legend-studio/src/components/editor/edit-panel/ModelLoader.tsx
@@ -39,6 +39,7 @@ import {
   EDITOR_LANGUAGE,
 } from '@finos/legend-application';
 import { StudioTextInputEditor } from '../../shared/StudioTextInputEditor';
+import type { ExtraModelLoaderRendererConfiguration } from '../../../stores/LegendStudioPlugin';
 
 export const ModelLoader = observer(() => {
   const editorStore = useEditorStore();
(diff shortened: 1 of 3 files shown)


packages/legend-studio/src/components/editor/edit-panel/ModelLoader.tsx
```diff
@@ -187,25 +214,35 @@ export const ModelLoader = observer(() => {
...
-            className="btn--dark model-loader__header__load-btn"
-            onClick={loadModel}
-            disabled={modelLoaderState.isLoadingModel}
-            tabIndex={-1}
-            title="Load model"
-          >
-            Load
-          </button>
-        </div>
+        {!modelLoaderState.currentExtraNativeInputType && (
```
MauricioUyaguari, line 217: Might make sense to add this as part of the extension definition. So we could add a `hardReplaceOption?: boolean | undefined` property on `ExtraModelLoaderRendererConfiguration` to see if we render this or not.

packages/legend-studio/src/stores/LegendStudioPlugin.ts
```diff
@@ -52,6 +53,12 @@ export type ClassPreviewRenderer = (
   _class: Class,
 ) => React.ReactNode | undefined;
 
+export type ExtraModelLoaderRendererConfiguration = {
+  key: string;
+  defaultModelText: string;
```
MauricioUyaguari, line 58: `ExtraModelLoaderRendererConfiguration` -> `ExtraModelLoaderExtensionsConfiguration`
We probably don't need `defaultModelText` as that is specific to external and native types. 
Also it probably make sense to add a `label?: string` property to this as we may not always want to use the key in the render option.

packages/legend-studio/src/components/editor/edit-panel/ModelLoader.tsx
```diff
@@ -187,25 +214,35 @@ export const ModelLoader = observer(() => {
...
+          <div className="model-loader__header__action">
+            <button
+              className="btn--dark model-loader__header__load-btn"
+              onClick={loadModel}
```
MauricioUyaguari, line 221: I think we also need an extension when user clicks `load`. Your extension will handle that click as load/import your model to studio.

MauricioUyaguari:
Lets also add a changset

akphi:
@emilia-sokol-gs I think the pipelines are not running for you. I don't know the exact reasons but try to force push again to see if it was a transient issue, otherwise, the only explanation I can think of is that since you haven't worked on `finos/legend-studio` in a while, `Github` auto-disable github actions on your fork

emilia-sokol-gs:
@akphi I pushed those changes again from another branch (I was using master so far) and this fixed this issue
@MauricioUyaguari I applied your feedback here: https://github.com/finos/legend-studio/pull/980 so I will close this PR

emilia-sokol-gs:
PR is currently moved here: https://github.com/finos/legend-studio/pull/980

PR #983, opened 2022-03-24
optimize query setup page

MauricioUyaguari:

- [ ] Test(s) added

### DIFF
--- a/packages/legend-graph/src/graphManager/AbstractPureGraphManager.ts
+++ b/packages/legend-graph/src/graphManager/AbstractPureGraphManager.ts
@@ -336,6 +336,19 @@ export abstract class AbstractPureGraphManager {
   abstract updateQuery(query: Query, graph: PureModel): Promise<Query>;
   abstract deleteQuery(queryId: string): Promise<void>;
 
+  // ------------------------------------------- Legend Query -------------------------------------
+  abstract buildGraphForCreateQuerySetup(
+    graph: PureModel,
+    entities: Entity[],
+    dependencyEntitiesMap: Map<string, Entity[]>,
+  ): Promise<void>;
+
+  abstract buildGraphForServiceQuerySetup(
+    graph: PureModel,
+    entities: Entity[],
+    dependencyEntitiesMap: Map<string, Entity[]>,
+  ): Promise<void>;
+
   // ------------------------------------------- Utilities -------------------------------------------
 
   abstract buildDatabase(
--- a/packages/legend-graph/src/models/protocols/pure/v1/V1_PureGraphManager.ts
+++ b/packages/legend-graph/src/models/protocols/pure/v1/V1_PureGraphManager.ts
@@ -39,6 +39,7 @@ import {
   promisify,
   type ActionState,
   StopWatch,
+  assertNonEmptyString,
 } from '@finos/legend-shared';
 import type { TEMPORARY__AbstractEngineConfig } from '../../../../graphManager/action/TEMPORARY__AbstractEngineConfig';
 import {
@@ -47,8 +48,11 @@ import {
   type GraphBuilderOptions,
   type ExecutionOptions,
 } from '../../../../graphManager/AbstractPureGraphManager';
-import type { Mapping } from '../../../metamodels/pure/packageableElements/mapping/Mapping';
-import type { Runtime } from '../../../metamodels/pure/packageableElements/runtime/Runtime';
+import { Mapping } from '../../../metamodels/pure/packageableElements/mapping/Mapping';
+import {
+  type Runtime,
+  EngineRuntime,
+} from '../../../metamodels/pure/packageableElements/runtime/Runtime';
 import type {
   ImportConfigurationDescription,
   ImportMode,
@@ -76,7 +80,11 @@ import type { ExecutionResult } from '../../../../graphManager/action/execution/
 import type { GenerationOutput } from '../../../../graphManager/action/generation/GenerationOutput';
 import type { ValueSpecification } from '../../../metamodels/pure/valueSpecification/ValueSpecification';
 import { ServiceExecutionMode } from '../../../../graphManager/action/service/ServiceExecutionMode';
-import { PureSingleExecution } from '../../../metamodels/pure/packageableElements/service/ServiceExecution';
+import {
+  KeyedExecutionParameter,
+  PureMultiExecution,
+  PureSingleExecution,
+} from '../../../metamodels/pure/packageableElements/service/ServiceExecution';
 import {
   V1_deserializeRawValueSpecification,
   V1_serializeRawValueSpecification,
@@ -162,6 +170,7 @@ import type { V1_Multiplicity } from './model/packageableElements/domain/V1_Mult
 import type { V1_RawVariable } from './model/rawValueSpecification/V1_RawVariable';
 import { V1_setupDatabaseSerialization } from './transformation/pureProtocol/serializationHelpers/V1_DatabaseSerializationHelper';
 import {
+  V1_PACKAGEABLE_RUNTIME_ELEMENT_PROTOCOL_TYPE,
   V1_setupEngineRuntimeSerialization,
   V1_setupLegacyRuntimeSerialization,
 } from './transformation/pureProtocol/serializationHelpers/V1_RuntimeSerializationHelper';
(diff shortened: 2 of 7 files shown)


packages/legend-graph/src/graphManager/AbstractPureGraphManager.ts
```diff
@@ -336,6 +336,19 @@ export abstract class AbstractPureGraphManager {
   abstract updateQuery(query: Query, graph: PureModel): Promise<Query>;
   abstract deleteQuery(queryId: string): Promise<void>;
 
+  //TODO: These methods will be removed once we have a way to `lazy` load elements into the graph
```
akphi, line 339: Let's remove this comment, also perhaps let's section this out by comment

```js
  // ------------------------------------------- Legend Query -------------------------------------------
```

packages/legend-query/src/stores/LegendQueryStore.ts
```diff
@@ -805,7 +816,80 @@ export class LegendQueryStore {
...
+    options?: { quiet?: boolean },
+  ): GeneratorFn<void> {
+    try {
+      const stopWatch = new StopWatch();
+      this.buildGraphState.inProgress();
+      const [entities, dependencyMap] = (yield flowResult(
+        this.fetchProjectEntities(project, versionId, stopWatch, options),
+      )) as [Entity[], Map<string, Entity[]>];
+      this.graphManagerState.resetGraph();
+      // // build dependencies
```
akphi, line 880: --> `// //`, also the above pls

packages/legend-graph/src/graphManager/AbstractPureGraphManager.ts
```diff
@@ -336,6 +336,19 @@ export abstract class AbstractPureGraphManager {
...
+  abstract buildCreateQueryLightGraph(
```
akphi, line 340: --> `buildGraphForCreateQuerySetup()`

packages/legend-graph/src/graphManager/AbstractPureGraphManager.ts
```diff
@@ -336,6 +336,19 @@ export abstract class AbstractPureGraphManager {
...
+    graph: PureModel,
+    entities: Entity[],
+    dependencyEntitiesMap: Map<string, Entity[]>,
+  ): Promise<void>;
+
+  abstract buildServiceQueryLightGraph(
```
akphi, line 346: --> `buildGraphForServiceQuerySetup()`

packages/legend-graph/src/models/protocols/pure/v1/V1_PureGraphManager.ts
```diff
@@ -2016,6 +2036,237 @@ export class V1_PureGraphManager extends AbstractPureGraphManager {
     );
   }
 
+  async indexEntitiesWithDependencyIntoGraph<T extends V1_PackageableElement>(
```
akphi, line 2039: So, my thinking is we could potentially rename this method to build light graph with just indexing (similar to what you have here) but return a slightly more complicated object instead: separating the `graph` and the `dependencies` part. That way we could reuse for the future `buildLightGraph` method which we would expose at interface level, i.e. `AbstractPureGraphManager`.

packages/legend-graph/src/models/protocols/pure/v1/V1_PureGraphManager.ts
```diff
@@ -2016,6 +2036,237 @@ export class V1_PureGraphManager extends AbstractPureGraphManager {
...
+    graph: PureModel,
+    _entities: Entity[],
+    dependencyEntities: Map<string, Entity[]>,
+    filteredElementClazzes: GenericClazz<T>[],
+    // TODO we don't need this once protocol type is properly modularized
+    filteredProtocolTypes: string[],
+  ): Promise<V1_GraphBuilderInput[]> {
+    // remove unnecessary entities
+    Array.from(dependencyEntities.entries()).forEach(
```
akphi, line 2048: To keep this method generic, I think the best way is not to filter entities here but in the caller and so we could eliminte `filteredElementClazzes` and `filteredProtocolTypes`

packages/legend-graph/src/models/protocols/pure/v1/V1_PureGraphManager.ts
```diff
@@ -2016,6 +2036,237 @@ export class V1_PureGraphManager extends AbstractPureGraphManager {
...
+      {
+        model: graph,
+        data: indexPureModelContextData(report, data, this.extensions),
+      },
+    ];
+    const graphBuilderInput = [
+      ...dependencyGraphBuilderInput,
+      ...mainGraphBuilderInput,
+    ];
+    graphBuilderInput.forEach((e) => {
```
akphi, line 2101: if we pre-filter, do we still need this?

packages/legend-graph/src/models/protocols/pure/v1/V1_PureGraphManager.ts
akphi: I'm not suggesting we have to do it now, but we should put a `TODO` if we don't.

packages/legend-graph/src/models/protocols/pure/v1/V1_PureGraphManager.ts
```diff
@@ -2016,6 +2036,237 @@ export class V1_PureGraphManager extends AbstractPureGraphManager {
...
+      e.data.elements = e.data.elements.filter((el: V1_PackageableElement) =>
+        filteredElementClazzes.find((_clazz) => el instanceof _clazz),
+      );
+    });
+    await flowResult(this.initializeAndIndexElements(graph, graphBuilderInput));
+    return graphBuilderInput;
+  }
+
+  async buildCreateQueryLightGraph(
```
akphi, line 2110: Could we put a comment to keep my suggestion about optimizing this further, i.e. to tinker with the entities to omit parts we don't need? Thanks!

packages/legend-graph/src/models/protocols/pure/v1/V1_PureGraphManager.ts
```diff
@@ -2016,6 +2036,237 @@ export class V1_PureGraphManager extends AbstractPureGraphManager {
...
+       * Wrap all error with `GraphBuilderError`, as we throw a lot of assertion error in the graph builder
+       * But we might want to rethink this decision in the future and throw appropriate type of error
+       */
+      throw error instanceof GraphBuilderError
+        ? error
+        : new GraphBuilderError(error);
+    }
+  }
+
+  async buildServiceQueryLightGraph(
```
akphi, line 2197: Could we put a comment to keep my suggestion about optimizing this further, i.e. to tinker with the entities to omit parts we don't need? Thanks!

packages/legend-query/src/stores/LegendQueryStore.ts
```diff
@@ -721,6 +723,45 @@ export class LegendQueryStore {
...
+      entities = (yield this.depotServerClient.getVersionEntities(
+        project.groupId,
+        project.artifactId,
+        versionId === LATEST_VERSION_ALIAS ? project.latestVersion : versionId,
+      )) as Entity[];
+    }
+    this.buildGraphState.setMessage(undefined);
+    stopWatch.record(GRAPH_MANAGER_EVENT.GRAPH_ENTITIES_FETCHED);
+
+    //fetch dependencies
```
akphi, line 750: please revert to the previous formatting.

packages/legend-query/src/stores/LegendQueryStore.ts
```diff
@@ -721,6 +723,45 @@ export class LegendQueryStore {
     }
   }
 
+  *fetchProjectEntities(
```
akphi, line 726: I have mixed-feelings about this. It's decent, but it tears away the analytics we have using `stopWatch`. I believe this might screw up `getRecord(...)` call later when we build the report for graph building. All in all, I think we should revert this.

akphi (changes requested):

packages/legend-graph/src/models/protocols/pure/v1/V1_PureGraphManager.ts
MauricioUyaguari: @akphi not sure i understand what you mean here. whats the purpose of returning this more complicated object ?

akphi (approved):

akphi:
@cla-bot[bot] check

PR #1009, opened 2022-03-30
Fix viewer mode showing incorrect grammar when graph error occurs

### LINKED ISSUE #868: Bug: Viewer mode fails to show correct grammar text when graph error occurs
### Similar issues

### How are you using Studio?

Legend Studio

During a graph building error, viewer mode stays in form mode and reports graph building error. When user tries to investigate and goes to text mode, the grammar is shown from the state of the pure model at the time of the error. Therefore viewer mode is showing an incorrect grammar. It should show the grammar using the original entities when loading. 

### Steps to reproduce

Have a project with the following grammar (consider using model loader). Create a review and merge this. Then try to view the project in viewer mode, and toggle `text-mode`. The grammar is wrong.

### Model data

```
Class model::NewClass extends model::NewClass2
{
  prop1: String[1];
}


###Mapping
Mapping model::NewMapping
(
  model::NewClass: Pure
  {
  }
)
```

### Environment

```text
Chrome
```

_No response_

### Contribution

gayathrir11:
## Summary

Closes #868 

- [ ] Test(s) added

***Before the fix***
pic2

***After the fix***
pic1

### DIFF
--- a/packages/legend-studio/src/stores/EditorStore.ts
+++ b/packages/legend-studio/src/stores/EditorStore.ts
@@ -1281,10 +1281,24 @@ export class EditorStore {
         showLoading: true,
       });
       try {
-        const graphGrammar =
-          (yield this.graphManagerState.graphManager.graphToPureCode(
-            this.graphManagerState.graph,
-          )) as string;
+        let graphGrammar;
+        if (
+          this.isInViewerMode &&
+          (!this.graphManagerState.graph.buildState.hasSucceeded ||
+            !this.graphManagerState.graph.dependencyManager.buildState
+              .hasSucceeded)
+        ) {
+          graphGrammar =
+            (yield this.graphManagerState.graphManager.entitiesToPureCode(
+              this.changeDetectionState.workspaceLocalLatestRevisionState
+                .entities,
+            )) as string;
+        } else {
+          graphGrammar =
+            (yield this.graphManagerState.graphManager.graphToPureCode(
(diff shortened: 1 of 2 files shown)


packages/legend-studio/src/stores/EditorStore.ts
```diff
@@ -1281,6 +1281,12 @@ export class EditorStore {
         showLoading: true,
       });
       try {
+        if (this.isInViewerMode) {
+          this.graphState.buildGraph(
```
MauricioUyaguari, line 1285: How does this fix the problem. 
I would assume that the `buildGraph` would still break with the given entities and form an incomplete graph and then give incomplete text.

akphi:
@gayathrir11 why don't we see the red underline in the second screenshot?

MauricioUyaguari:
Seems like we may need to force a compilation after entering text mode to show the error

gayathrir11:
> Seems like we may need to force a compilation after entering text mode to show the error

yeah. Looking into the code to understand how we do it for a workspace

akphi:
Thanks @gayathrir11 and @MauricioUyaguari after discussing, I realized we are aiming at the wrong target here. What we should do is not focusing on toggling to text-mode, but try to do something similar to editor mode in viewer mode where we properly reveal compilation error in `text-mode`. I have opened https://github.com/finos/legend-studio/pull/1199 to address this, so I'll be closing this PR here.

PR #1020, opened 2022-04-01
move actions to legend-studio part 3

MauricioUyaguari:

- [ ] Test(s) added

### DIFF
--- a/packages/legend-graph/src/helpers/MappingResolutionHelper.ts
+++ b/packages/legend-graph/src/helpers/MappingResolutionHelper.ts
@@ -42,12 +42,14 @@ export const updateRootSetImplementationOnCreate = (
     setImp.class.value,
   ).filter((si) => si !== setImp);
   if (classMappingsWithSimilarTarget.length) {
-    setImp.setRoot(false);
+    setImp.root.setValue(false);
     if (classMappingsWithSimilarTarget.length === 1) {
-      (classMappingsWithSimilarTarget[0] as SetImplementation).setRoot(false);
+      (classMappingsWithSimilarTarget[0] as SetImplementation).root.setValue(
+        false,
+      );
     }
   } else {
-    setImp.setRoot(true);
+    setImp.root.setValue(true);
   }
 };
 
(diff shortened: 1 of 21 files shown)


packages/legend-graph/src/helpers/MappingResolutionHelper.ts
```diff
@@ -42,12 +42,14 @@ export const updateRootSetImplementationOnCreate = (
     setImp.class.value,
   ).filter((si) => si !== setImp);
   if (classMappingsWithSimilarTarget.length) {
-    setImp.setRoot(false);
```
akphi, line 45: looking at these, I wonder if these methods should be here or `legend-studio`

packages/legend-graph/src/models/protocols/pure/v1/transformation/pureGraph/to/helpers/V1_MappingBuilderHelper.ts
```diff
@@ -189,7 +190,8 @@ export const V1_buildMappingInclude = (
     context.resolveMapping(mappingInclude.includedMappingPath),
   );
   if (mappingInclude.sourceDatabasePath && mappingInclude.targetDatabasePath) {
-    includedMapping.addStoreSubstitution(
+    addUniqueEntry(
```
akphi, line 193: I think we can spare the `addUnqueEntry` here since we're adding new instances anyway.

packages/legend-graph/src/helpers/MappingResolutionHelper.ts
```diff
@@ -42,12 +42,14 @@ export const updateRootSetImplementationOnCreate = (
...
+    setImp.root.setValue(false);
```
akphi, line 45: Are we tackling this `setValue` later?

packages/legend-graph/src/helpers/MappingResolutionHelper.ts
akphi: In particular, as I just checked, let's move
`updateRootSetImplementationOnCreate`, `updateRootSetImplementationOnDelete`, and `nominateRootSetImplementation` to `legend-studio` and leave the rest of them in `legend-graph`, also it's safe to move them to say `MappingHelper`. Also, we could eliminate `classMappingsWithSimilarTarget[0].setRoot(true);` I don't even trust the logic we have in `findRootSetImplementation` anymore.

akphi (approved):

Issue #628, opened 2021-10-29
Feature Request: support more date/time capabilities in query filter


akphi:
### Similar requests

- [X] I have searched and found no existing similar requests

### What problems are you trying to solve?

There are more date capabilities to support in query builder filter:
- [ ] Functions
  - [ ] Now: now()
  - [ ] Today: today()
  - [ ] ... (To be added)
- [ ] Absolute date (we probably can support time here as well)
- [ ] Custom date: allow building a date by specifying 4 measures `Number  | Unit | Before/After | When` (when refers to a special date created by a function, like `today()`, `now()`, etc.

### Contribution

- [ ] I would like to work on this feature

PR #1021, opened 2022-04-01
Support more date/time capabilities in query builder filter

YannanGao-gs:
## Summary

Fixes https://github.com/finos/legend-studio/issues/628

- [ ] Test(s) added

- Date Filter

https://user-images.githubusercontent.com/73408381/166477477-09946fb9-e0ab-4ecf-9308-ddbba2d10c96.mov

- DateTime Filter

https://user-images.githubusercontent.com/73408381/166478478-7c4d1156-18c9-4785-9b8b-531bafcd7829.mov

- Date and DateTime Post Filter

https://user-images.githubusercontent.com/73408381/166482231-c3c05c47-fa77-4843-8515-f30d0835c23b.mov

- Date and DateTime Derived Properties

https://user-images.githubusercontent.com/73408381/166479269-aaa2762c-5808-4384-9975-960e7500480a.mov

- Date and DateTime Parameters

https://user-images.githubusercontent.com/73408381/166481805-9e9dfd0d-5e0f-41f7-956b-a4d4c7be2b11.mov

Model used in the video
```
Enum model::postFilter::MyEnum
{
  Enum1,
  Enum2
}

Class model::postFilter::Person
{
  fullName: String[1];
  myOptionalBoolean: Boolean[0..1];
  myNumber: Number[1];
  age: Integer[1];
  myFloat: Float[1];
  myDecimal: Decimal[1];
  myStrictDate: StrictDate[1];
  myDateTime: DateTime[1];
  myDate: Date[1];
  myEnum: model::postFilter::MyEnum[1];
  myBinary: Binary[1];
  myOptionalEnum: model::postFilter::MyEnum[0..1];
  testDerived(date: Date[1],dateTime: DateTime[1],strictDate: StrictDate[1]) {''}: String[1];
}


###Mapping
Mapping model::postFilter::M2MMapping
(
  *model::postFilter::Person: Pure
  {
    ~src model::postFilter::Person
  }
)


###Connection
JsonModelConnection model::postFilter::MyConnection
{
  class: model::postFilter::Person;
  url: 'data:application/json,%7B%7D';
}


###Runtime
Runtime model::postFilter::MyRuntime
{
  mappings:
  [
    model::postFilter::M2MMapping
  ];
  connections:
  [
    ModelStore:
    [
      connection_1: model::postFilter::MyConnection
    ]
  ];
}

```


packages/legend-query/src/stores/QueryBuilderOperatorsHelper.ts
```diff
@@ -82,6 +82,10 @@ export const getNonCollectionValueSpecificationType = (
     return guaranteeNonNullable(valueSpecification.values[0]).value.owner;
   } else if (valueSpecification instanceof VariableExpression) {
     return valueSpecification.genericType?.value.rawType;
+  } else if (valueSpecification instanceof SimpleFunctionExpression) {
+    return valueSpecification.genericType
```
MauricioUyaguari, line 86:
```suggestion
valueSpecification.genericType?.value.rawType
```

packages/legend-query/src/components/QueryBuilderValueSpecificationEditor.tsx
```diff
@@ -605,11 +1159,22 @@ export const DateInstanceValueEditor = observer(
 
 export const QueryBuilderValueSpecificationEditor: React.FC<{
   valueSpecification: ValueSpecification;
+  queryBuilderState: QueryBuilderState;
```
MauricioUyaguari, line 1162: I dont like `queryBuilderState` being added as a parameter here. 
`graph` should be enough to do everything you need for the editor. 
Stuff like `buildPrimitiveInstanceValue` can be refactored just to take a `PureModel`

packages/legend-query/src/components/QueryBuilderValueSpecificationEditor.tsx
```diff
@@ -605,11 +1159,22 @@ export const DateInstanceValueEditor = observer(
 
 export const QueryBuilderValueSpecificationEditor: React.FC<{
```
MauricioUyaguari, line 1160: I think the reason this editor is a bit chunky to code in is because we are editing value specification directly.
I think what we should do is built a ValueSpecificationEditorState which is build in the lambda processing and mirrors the rest of query builder app. 
@akphi thoughts?

packages/legend-query/src/components/QueryBuilderFilterPanel.tsx
```diff
@@ -262,6 +263,10 @@ const QueryBuilderFilterConditionEditor = observer(
               )}
               <QueryBuilderValueSpecificationEditor
                 valueSpecification={node.condition.value}
+                queryBuilderState={node.condition.filterState.queryBuilderState}
+                setValueSpecification={(
```
MauricioUyaguari, line 267: `updateValueSpecification`

packages/legend-query/src/components/QueryBuilderPropertyExpressionEditor.tsx
```diff
@@ -69,7 +69,19 @@ const DerivedPropertyParameterEditor = observer(
           item: QueryBuilderParameterDragSource,
           monitor: DropTargetMonitor,
         ): void => {
-          if (!monitor.didDrop()) {
+          if (
+            !monitor.didDrop() &&
+            (derivedPropertyExpressionState.parameters[
```
MauricioUyaguari, line 74: this code is kind of hard to read. Could we maybe abstract this out and name it exactly what we are doing. We checking compatiblity of types right?

packages/legend-query/src/stores/filterOperators/QueryBuilderFilterOperatorHelper.ts
```diff
@@ -316,8 +317,13 @@ const buildFilterConditionStateWithExists = (
         existsLambdaExpressions.push(mainFilterExpression.parametersValues[0]);
       }
     }
+
+    const reservedDateTimeOperatorFunctionFullPath = Array.from(
```
MauricioUyaguari, line 321: why are doing this. What case would this result from ?

.changeset/funny-pianos-teach.md
```diff
@@ -0,0 +1,5 @@
+---
+'@finos/legend-query': patch
```
akphi, line 2:
```suggestion
'@finos/legend-query': minor
```

.changeset/funny-pianos-teach.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Support more date capabilites in query builder.
```
akphi, line 5:
```suggestion
Support more date/time capabilities in filter.
```

packages/legend-query/src/QueryBuilder_Const.ts
```diff
@@ -57,6 +57,20 @@ export enum SUPPORTED_FUNCTIONS {
   LESS_THAN = 'meta::pure::functions::lang::tests::lessThan',
   LESS_THAN_EQUAL = 'meta::pure::functions::lang::tests::lessThanEqual',
   STARTS_WITH = 'meta::pure::functions::string::startsWith',
+  ADJUST = 'meta::pure::functions::date::adjust',
```
akphi, line 60: let's add a newline here to separate these and the operations, also please remove the empty line `56`

packages/legend-query/src/stores/QueryParametersState.ts
```diff
@@ -18,6 +18,7 @@ import addDays from 'date-fns/addDays';
 import {
   type InstanceValue,
   type Type,
+  type ValueSpecification,
```
akphi, line 21: This change look weird. I think it might throw off linter

packages/legend-query/src/stores/QueryBuilderFilterState.ts
```diff
@@ -58,7 +58,7 @@ import {
 export abstract class QueryBuilderFilterOperator {
   uuid = uuid();
 
-  abstract getLabel(filterConditionState: FilterConditionState): string;
```
akphi, line 61: why do we change this? we could use the state in the future you know?

packages/legend-query/src/components/QueryBuilderParameterPanel.tsx
```diff
@@ -98,8 +100,13 @@ const ParameterValuesEditor = observer(
                   {paramState.value && (
                     <QueryBuilderValueSpecificationEditor
                       valueSpecification={paramState.value}
+                      queryBuilderState={queryBuilderState}
+                      setValueSpecification={(
+                        val: ValueSpecification | undefined,
```
akphi, line 105: why don't make this `val: ValueInstance`

packages/legend-query/src/components/QueryBuilderValueSpecificationEditor.tsx
akphi: That's fair point, we could make this generic `ValueSpecEditor` if we can remove `queryBuilderState`

packages/legend-query/src/components/QueryBuilderValueSpecificationEditor.tsx
```diff
@@ -617,11 +1180,22 @@ export const DateInstanceValueEditor = observer(
  */
 export const QueryBuilderValueSpecificationEditor: React.FC<{
   valueSpecification: ValueSpecification;
+  queryBuilderState: QueryBuilderState;
   graph: PureModel;
   expectedType: Type;
   className?: string | undefined;
+  setValueSpecification: (val: ValueSpecification | undefined) => void;
+  matchType?: boolean | undefined;
```
akphi, line 1188: I don't think I get the point of having `matchType`. We should document it, and perhaps remove it altogether. Let's discuss about that. Anyway, we should make `expectedType` optional as well.

packages/legend-query/src/components/QueryBuilderValueSpecificationEditor.tsx
```diff
@@ -617,11 +1180,22 @@ export const DateInstanceValueEditor = observer(
  */
 export const QueryBuilderValueSpecificationEditor: React.FC<{
   valueSpecification: ValueSpecification;
+  queryBuilderState: QueryBuilderState;
   graph: PureModel;
   expectedType: Type;
   className?: string | undefined;
+  setValueSpecification: (val: ValueSpecification | undefined) => void;
```
akphi, line 1187: --> `updateValueSpecification`

packages/legend-query/src/components/QueryBuilderValueSpecificationEditor.tsx
akphi: As discussed, `matchedType` behavior can be controlled using `expectedType`, for example, when we're dealing with filter, we can say if the type of the field is `DateTime`, we can say expected type is `Date`, for example, thanks to stuff like `IS_ON_DAY`.

packages/legend-query/src/stores/filterOperators/QueryBuilderFilterOperatorHelper.ts
```diff
@@ -443,22 +449,38 @@ export const buildFilterConditionState = (
   // This is used for post-build checks (useful when this expression is nested inside a longer
   // chain of expression, e.g. ->exists($x|x == 'something'), etc.)
   let mainExpressionWithOperator: SimpleFunctionExpression | undefined;
-
-  if (matchFunctionName(expression.functionName, operatorFunctionFullPath)) {
+  const reservedDateTimeOperatorFunctionFullPath = Array.from(
```
akphi, line 452: Same as Mauricio, I think we shouldn't expose those function at this level. This is an utility and when we introduce `QUERY_BUILDER_RESERVED_DATETIME_OPERATOR_FUNCTION_TO_LABEL` here, we are adding unecessary logic which could have been added to the operator function itself. When you put this logic here, we also lose some nuances of the logic: we know for example, `IS_ON_DAY` only applies to `EQUAL` operator.

packages/legend-query/src/stores/QueryBuilderDateCapabilityHelper.ts
```diff
@@ -0,0 +1,588 @@
...
+} from '../QueryBuilder_Const';
+import { buildPrimitiveInstanceValue } from './QueryBuilderOperatorsHelper';
+import type { QueryBuilderState } from './QueryBuilderState';
+
+export type CustomDateInputOption = {
+  label: string;
+  value: string;
+};
+
+export type DateOption = {
```
akphi, line 47: --> `CustomDateOption`

packages/legend-query/src/stores/QueryBuilderDateCapabilityHelper.ts
```diff
@@ -0,0 +1,588 @@
...
+}
+
+export enum QUERY_BUILDER_DATE_DURATION_UNIT_LABELS {
+  DAYS = 'Day(s)',
+  WEEKS = 'Week(s)',
+  MONTHS = 'Month(s)',
+  YEARS = 'Year(s)',
+}
+
+export enum QUERY_BUILDER_DATE_DURATION_UNITS {
```
akphi, line 85: just our convention for `enum`s, don't do plural, --> `QUERY_BUILDER_DATE_DURATION_UNIT`

packages/legend-query/src/stores/QueryBuilderDateCapabilityHelper.ts
```diff
@@ -0,0 +1,588 @@
...
+  ONE_YEAR_AGO = 'One Year Ago',
+  ONE_MONTH_AGO = 'One Month Ago',
+  ONE_WEEK_AGO = 'One Week Ago',
+  CUSTOM_DATE = 'Custom Date',
+  ABSOLUTE_DATE = 'Absolute Date',
+  PREVIOUS_DAY_OF_WEEK = 'Previous ... of Week',
+  START_DAY_OF_DATE = 'Start of...',
+}
+
+export enum QUERY_BUILDER_DATE_DURATION_UNIT_LABELS {
```
akphi, line 78: just our convention for `enum`s, don't do plural, --> `QUERY_BUILDER_DATE_DURATION_UNIT_LABEL`

packages/legend-query/src/stores/QueryBuilderDateCapabilityHelper.ts
```diff
@@ -0,0 +1,588 @@
...
+  SATURDAY = 'Saturday',
+  SUNDAY = 'Sunday',
+}
+
+export enum QUERY_BUILDER_DATE_ADJUSTMENT {
+  BEFORE = 'Before',
+  AFTER = 'After',
+}
+
+export enum QUERY_BUILDER_CUSTOM_DATE_START_DATE_OPTIONS {
```
akphi, line 107: --> `QUERY_BUILDER_CUSTOM_DATE_START_DATE_OPTION`

packages/legend-query/src/stores/QueryBuilderDateCapabilityHelper.ts
```diff
@@ -0,0 +1,588 @@
+/**
```
akphi, line 1: Let's rename this file to `QueryBuilderDateTimeHelper`

packages/legend-query/src/stores/QueryBuilderDateCapabilityHelper.ts
```diff
@@ -0,0 +1,588 @@
...
+    }
+  } else {
+    return {
+      label: valueSpecification.values[0] as string,
+      value: QUERY_BUILDER_DATE_OPTONS.ABSOLUTE_DATE,
+    };
+  }
+};
+
+export const chunkArray = (arr: unknown[], stepSize: number): unknown[][] => {
```
akphi, line 582: What do we use this for?

akphi, line 582: We should move this as part of componentizing `radio group` in `legend-art`

packages/legend-query/src/components/QueryBuilderValueSpecificationEditor.tsx
```diff
@@ -617,11 +1172,22 @@ export const DateInstanceValueEditor = observer(
...
```
akphi, line 1179: Would be nice to have an optional parameter `resetValueSpecification` to enable the reset behavior though. If people don't specify it, we can make use of the default reset behavior (i.e. use the mock value for primitive type utility), else if they set it to `null`, we disable `reset`. Something like that. If you don't do it, leave this comment here so we can scrape it next time.

packages/legend-query/src/components/QueryBuilderValueSpecificationEditor.tsx
```diff
@@ -53,20 +60,68 @@ import {
   PRIMITIVE_TYPE,
   TYPICAL_MULTIPLICITY_TYPE,
   VariableExpression,
+  SimpleFunctionExpression,
 } from '@finos/legend-graph';
 import { getMultiplicityDescription } from './shared/QueryBuilderUtils';
+import { FormControlLabel, Radio, RadioGroup } from '@mui/material';
```
akphi, line 66: As mentioned before, let's expose these via `@finos/legend-art` in a make-sense way.

packages/legend-query/src/stores/filterOperators/QueryBuilderFilterOperator_Equal.ts
```diff
@@ -144,22 +164,33 @@ export class QueryBuilderFilterOperator_Equal extends QueryBuilderFilterOperator
...
+          }'`,
         );
     }
   }
 
   buildFilterConditionExpression(
     filterConditionState: FilterConditionState,
   ): ValueSpecification {
+    if (
+      filterConditionState.propertyExpressionState.propertyExpression.func
```
MauricioUyaguari, line 184: this is only applied to `DateTime` not date?

packages/legend-query/src/stores/QueryBuilderDateTimeHelper.ts
```diff
@@ -0,0 +1,597 @@
...
+      value:
+        valueSpecification.genericType.value.rawType.path ===
+        PRIMITIVE_TYPE.DATETIME
+          ? QUERY_BUILDER_DATE_OPTION.ABSOLUTE_TIME
+          : QUERY_BUILDER_DATE_OPTION.ABSOLUTE_DATE,
+    };
+  }
+};
+
+export const chunkArray = (arr: unknown[], stepSize: number): unknown[][] => {
```
MauricioUyaguari, line 591: should probably be in `legend-shared`

packages/legend-query/src/components/QueryBuilderPropertyExpressionEditor.tsx
```diff
@@ -114,16 +127,39 @@ const DerivedPropertyParameterValueEditor = observer(
...
-                .value.rawType
+            valueSpecification={
+              derivedPropertyExpressionState.parameterValues[
+                idx
+              ] as ValueSpecification
             }
+            updateValueSpecification={(
+              val: ValueSpecification | undefined,
+            ): void => {
+              if (val) {
```
MauricioUyaguari, line 138: lets not clutter the Component with this logic. Could we move this up to a declared function. Looks a bit cleaner.

packages/legend-query/src/components/QueryBuilderValueSpecificationEditor.tsx
```diff
@@ -533,75 +587,596 @@ export const LatestDatePrimitiveInstanceValueEditor: React.FC = () => (
...
+    } = props;
     const strictDate = graph.getPrimitiveType(PRIMITIVE_TYPE.STRICTDATE);
     const date = graph.getPrimitiveType(PRIMITIVE_TYPE.DATE);
     const dateTime = graph.getPrimitiveType(PRIMITIVE_TYPE.DATETIME);
     const latestDate = graph.getPrimitiveType(PRIMITIVE_TYPE.LATESTDATE);
-    const changeType = (val: PackageableElementOption<Type>): void => {
-      if (variableType !== val.value) {
-        genericType_setRawType(valueSpecification.genericType.value, val.value);
-      }
+    const targetQueryBuilderDateOptionsEnum = typeCheckOption.match
```
MauricioUyaguari, line 1028: why do we need a specific input of `match` (probably not a good name as i dont know what it is intended for) ? 
Seems like it only depends on the type and whether it is of type `DateTime`.

packages/legend-query/src/components/QueryBuilderValueSpecificationEditor.tsx
```diff
@@ -533,75 +587,596 @@ export const LatestDatePrimitiveInstanceValueEditor: React.FC = () => (
...
+              }}
+            >
+              <BaseRadioGroup
+                value={dateOptionValue.value}
+                onChange={handleDateOptionChange}
+                row={true}
+                options={targetQueryBuilderDateOptionsEnum}
+                size={2}
+              />
+              {dateOptionValue.value ===
```
MauricioUyaguari, line 1125: Could we leverage case statements for this rendering. Becomes very unclear when we have multiply && statements like this as to what is being rendered in the screen

packages/legend-graph/src/MetaModelUtils.ts
```diff
@@ -88,3 +89,13 @@ export const hashLambda = (
     parameters ? hashObjectWithoutSourceInformation(parameters) : '',
     body ? hashObjectWithoutSourceInformation(body) : '',
   ]);
+
+export const isCompatibleWithType = (
+  type: Type | undefined,
```
MauricioUyaguari, line 94: any reason to allow the `type` for this function to be undefined? seems to specific to its usage in query builder than the actual function `name`. Returning true when its undefined seems wrong.

MauricioUyaguari (changes requested):

packages/legend-query/src/components/QueryBuilderValueSpecificationEditor.tsx
YannanGao-gs: For some cases, we need to filter out some options for DateTime because value of date/strictdate type can't be assigned to DateTime anymore. E.g. the dateTime parameters of derivedProperty. We pass in the value directly without calling isOnDay, isAfterDay.
Comments were added in the file too.

packages/legend-query/src/stores/filterOperators/QueryBuilderFilterOperator_Equal.ts
YannanGao-gs: Since Date is the supertype, RHS could be date/datetime/strictdate,  etc.
For DateTime, if RHS is date/stricdate, isOnDay() needs to be called.

packages/legend-query/src/stores/postFilterOperators/QueryBuilderPostFilterOperator_Equal.ts
```diff
@@ -50,7 +53,37 @@ export class QueryBuilderPostFilterOperator_Equal extends QueryBuilderPostFilter
...
+    let innerExpression = expression;
+    if (matchFunctionName(expression.functionName, SUPPORTED_FUNCTIONS.NOT)) {
+      innerExpression = guaranteeNonNullable(unwrapNotExpression(expression));
+    }
+    if (
+      innerExpression.parametersValues[0] instanceof
+        AbstractPropertyExpression &&
+      innerExpression.parametersValues[0].func.genericType.value.rawType
+        .path === PRIMITIVE_TYPE.DATETIME &&
+      innerExpression.parametersValues[1]?.genericType?.value.rawType.path !==
```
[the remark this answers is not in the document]
YannanGao-gs: The high-level idea here is the same as what we do to the filter.
Since Date is the supertype, RHS could be date/datetime/strictdate, etc.
For DateTime, if RHS is date/stricdate, isOnDay() needs to be called.

packages/legend-query/src/stores/postFilterOperators/QueryBuilderPostFilterOperator_Equal.ts
```diff
@@ -90,8 +123,22 @@ export class QueryBuilderPostFilterOperator_Equal extends QueryBuilderPostFilter
...
+        PRIMITIVE_TYPE.DATETIME,
+        PRIMITIVE_TYPE.STRICTDATE,
+        PRIMITIVE_TYPE.LATESTDATE,
+      ] as string[];
+
       // When changing the return type for LHS, the RHS value should be adjusted accordingly.
       // Numeric value is handled loosely because execution still works if a float (RHS) is assigned to an Integer property(LHS), etc.
+      if (
+        lhsType.path === PRIMITIVE_TYPE.DATETIME &&
+        type !== undefined &&
```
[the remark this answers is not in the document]
YannanGao-gs: `datePrimitiveTypes.includes(type?.path)` will trigger an error, Type 'undefined' is not assignable to type 'string'.
  
`When LHS is of type DateTime, RHS is handled loosely since the operator could be changed to another pure function.  e.g. is -> isOnDay()`.  comments were added.

packages/legend-query/src/stores/QueryBuilderDateTimeHelper.ts
YannanGao-gs: This should be removed. I moved this function into BaseRadioGroup.tsx, and renamed it to `transformToMatrix`.

packages/legend-graph/src/MetaModelUtils.ts
YannanGao-gs: I removed this, and created a query builder specific type-checking function.

packages/legend-query/src/components/QueryBuilderParameterPanel.tsx
```diff
@@ -98,8 +101,18 @@ const ParameterValuesEditor = observer(
                   {paramState.value && (
                     <QueryBuilderValueSpecificationEditor
                       valueSpecification={paramState.value}
-                      graph={queryBuilderState.graphManagerState.graph}
-                      expectedType={variableType}
+                      updateValueSpecification={(
+                        val: ValueSpecification | undefined,
+                      ): void =>
+                        paramState.setValue(val ? (val as InstanceValue) : val)
```
akphi: Might as well do a type check here:

```ts
if (val instanceof InstanceValue) ...
```

packages/legend-query/src/components/QueryBuilderPropertyExpressionEditor.tsx
```diff
@@ -64,14 +74,27 @@ const DerivedPropertyParameterValueEditor = observer(
       },
       [derivedPropertyExpressionState, idx],
     );
+    const isDraggedTypeCompatibleWithTarget = (
```
akphi, line 77: I think we can inline this instead of creating this function.

packages/legend-query/src/stores/QueryBuilderDateTimeHelper.ts
```diff
@@ -0,0 +1,589 @@
...
+  value: string;
+  durationRange?: string | undefined;
+  unit?: string | undefined;
+  adjustment?: string | undefined;
+  startDate?: string | undefined;
+};
+
+export type TypeCheckOption = {
+  expectedType: Type;
+  match?: boolean | undefined;
```
akphi, line 58: can we document this flag?

packages/legend-query/src/models/protocols/pure/v1/V1_QueryBuilder_FunctionExpressionBuilder.ts
```diff
@@ -47,8 +47,64 @@ import {
   AbstractPropertyExpression,
   TYPICAL_MULTIPLICITY_TYPE,
   VariableExpression,
+  PRIMITIVE_TYPE,
 } from '@finos/legend-graph';
 
+const generateGenericTypeForBaseSimpleFunctionExpression = (
```
akphi, line 53: I think this is the wrong place to do this, let's leverage this to `V1_getExtraFunctionExpressionBuilders` level

packages/legend-query/src/QueryBuilder_Const.ts
```diff
@@ -70,6 +84,7 @@ export enum SUPPORTED_FUNCTIONS {
   STD_DEV_POPULATION = 'meta::pure::functions::math::stdDevPopulation',
   STD_DEV_SAMPLE = 'meta::pure::functions::math::stdDevSample',
   SUM = 'meta::pure::functions::math::sum',
+  MINUS = 'meta::pure::functions::math::minus',
```
akphi, line 87: `MINUS` is not part of `aggregation` so let's move `MINUS` and `ADJUST` to their separate group for now,

```ts
...

// adjust time
ADJUST = 'meta::pure::functions::date::adjust',
MINUS = 'meta::pure::functions::math::minus',

...
```

packages/legend-query/src/stores/QueryBuilderDateTimeHelper.ts
akphi: Also, if this is used in `ValueSpecEditor`, let's move it there.

packages/legend-query/src/stores/QueryBuilderDateTimeHelper.ts
```diff
@@ -0,0 +1,589 @@
...
+  UnsupportedOperationError,
+} from '@finos/legend-shared';
+import {
+  DAY_OF_WEEK,
+  DURATION_UNIT,
+  SUPPORTED_FUNCTIONS,
+} from '../QueryBuilder_Const';
+import { buildPrimitiveInstanceValue } from './QueryBuilderOperatorsHelper';
+
+export type CustomDateInputOption = {
```
akphi, line 42: -> move this to `ValueSpecEditor`

packages/legend-query/src/stores/QueryBuilderDateTimeHelper.ts
```diff
@@ -0,0 +1,589 @@
...
+      return previousFridaySFE;
+    }
+    default:
+      throw new UnsupportedOperationError(
+        `${dateOptionValue.value} is not supported yet.`,
+      );
+  }
+};
+
+export const getDurationUnitEnumFromString = (
```
akphi, line 261: We don't need to export this right?

packages/legend-query/src/stores/QueryBuilderDateTimeHelper.ts
```diff
@@ -0,0 +1,589 @@
...
+        durationNumberParam.functionName,
+        SUPPORTED_FUNCTIONS.MINUS,
+      )
+    ? durationNumberParam.parametersValues[0] instanceof PrimitiveInstanceValue
+      ? (durationNumberParam.parametersValues[0].values[0] as number).toString()
+      : ''
+    : '';
+};
+
+export const getAdjustmentValue = (
```
akphi, line 365: We don't need to export this

packages/legend-query/src/stores/QueryBuilderDateTimeHelper.ts
```diff
@@ -0,0 +1,589 @@
...
+      )
+      .map(
+        (key) =>
+          QUERY_BUILDER_DATE_DURATION_UNIT_LABEL[
+            key as keyof typeof QUERY_BUILDER_DATE_DURATION_UNIT_LABEL
+          ],
+      )[0],
+  );
+
+export const getStartDateValue = (
```
akphi, line 396: we don't need to export this

packages/legend-query/src/stores/QueryBuilderDateTimeHelper.ts
```diff
@@ -0,0 +1,589 @@
...
+    case SUPPORTED_FUNCTIONS.FIRST_DAY_OF_MONTH:
+      return QUERY_BUILDER_CUSTOM_DATE_START_DATE_OPTION.START_OF_MONTH;
+    case SUPPORTED_FUNCTIONS.FIRST_DAY_OF_WEEK:
+      return QUERY_BUILDER_CUSTOM_DATE_START_DATE_OPTION.START_OF_WEEK;
+    default:
+      return `${funcName} is not supported`;
+  }
+};
+
+export const generateDateOptionForAdjustFunction = (
```
akphi, line 420: we don't need to export this

packages/legend-query/src/stores/QueryBuilderDateTimeHelper.ts
```diff
@@ -0,0 +1,589 @@
...
+      return {
+        label: QUERY_BUILDER_CUSTOM_DATE_START_DATE_OPTION.START_OF_YEAR,
+        value: QUERY_BUILDER_CUSTOM_DATE_START_DATE_OPTION.START_OF_YEAR,
+      };
+    default:
+      return { label: '', value: '' };
+  }
+};
+
+export const generateDateOption = (
```
akphi, line 521: can you check for all the functions that you don't need to export and remove the `export` keyword. Let's not pollute the import namespace. Also, could we document all of these functions?

packages/legend-query/src/stores/QueryBuilderPostFilterOperator.ts
```diff
@@ -34,7 +34,13 @@ export abstract class QueryBuilderPostFilterOperator {
...
-  abstract getPureFunction(): SUPPORTED_FUNCTIONS | undefined;
+  abstract getPureFunction(
+    filterConditionState: PostFilterConditionState,
+  ): SUPPORTED_FUNCTIONS | undefined;
+
+  abstract getPureFunctionFromExpression(
+    expression: FunctionExpression,
+  ): SUPPORTED_FUNCTIONS | undefined;
 
   getTdsColumnGetter(): TDS_COLUMN_GETTERS | undefined {
```
akphi, line 45: Can you help me rename the enum to `TDS_COLUMN_GETTER` instead. Thanks!

packages/legend-query/src/stores/filterOperators/QueryBuilderFilterOperator_LessThan.ts
```diff
@@ -119,6 +137,17 @@ export class QueryBuilderFilterOperator_LessThan extends QueryBuilderFilterOpera
   buildFilterConditionExpression(
     filterConditionState: FilterConditionState,
   ): ValueSpecification {
+    if (
```
akphi, line 140: Let's inline this logic. Could you also do this for all of them?

```ts
return buildFilterConditionExpression(
        filterConditionState,
        filterConditionState.propertyExpressionState.propertyExpression.func
        .genericType.value.rawType.path === PRIMITIVE_TYPE.DATETIME &&
      filterConditionState.value?.genericType?.value.rawType.path !==
        PRIMITIVE_TYPE.DATETIME
        ? SUPPORTED_FUNCTIONS.IS_BEFORE_DAY
        : SUPPORTED_FUNCTIONS.LESS_THAN,
      );
```

packages/legend-query/src/stores/QueryBuilderPostFilterOperator.ts
```diff
@@ -34,7 +34,13 @@ export abstract class QueryBuilderPostFilterOperator {
 
   abstract getLabel(): string;
 
-  abstract getPureFunction(): SUPPORTED_FUNCTIONS | undefined;
+  abstract getPureFunction(
```
akphi, line 37: These 2 functions `getPureFunction` and `getPureFunctionFromExpression` nolonger makes sense, let's remove them and do it the same way we do `QueryBuilderFilterOperator`. Thanks!

packages/legend-query/src/components/QueryBuilderValueSpecificationEditor.tsx
```diff
@@ -14,7 +14,13 @@
  * limitations under the License.
```
akphi, line 14: I will review this file last. Please update the video to the latest so we can make some discussion on styling as well. Thanks!

akphi (changes requested):

packages/legend-query/src/stores/filterOperators/QueryBuilderFilterOperator_Equal.ts
```diff
@@ -92,8 +97,24 @@ export class QueryBuilderFilterOperator_Equal extends QueryBuilderFilterOperator
       PRIMITIVE_TYPE.FLOAT,
     ] as string[];
 
+    const datePrimitiveTypes = [
```
akphi, line 100: let's rename this to `DATE_PRIMITIVE_TYPES` and apply to other places. Thanks!

packages/legend-query/src/components/QueryBuilderPropertyExpressionEditor.tsx
YannanGao-gs: See the previous comment from Mauricio: `this code is kind of hard to read. Could we maybe abstract this out and name it exactly what we are doing. We checking compatiblity of types right?`
Let me know if we need to inline this

packages/legend-graph/src/MetaModelUtils.ts
```diff
@@ -88,3 +89,13 @@ export const hashLambda = (
     parameters ? hashObjectWithoutSourceInformation(parameters) : '',
     body ? hashObjectWithoutSourceInformation(body) : '',
   ]);
+
+export const isCompatibleWithType = (
```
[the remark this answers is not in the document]
YannanGao-gs: This file is outdated. The function has been removed

packages/legend-query/src/components/QueryBuilderPropertyExpressionEditor.tsx
akphi: yeah, @MauricioUyaguari suggested one way of making the code more readable, I would argue that abstracting it with this minute piece of logic make code equally hard to read. Instead, what we should do here is to inline and write comment for that condition.

packages/legend-art/style/reset/_mui.scss
```diff
@@ -64,3 +64,24 @@
 .MuiTooltip-tooltipPlacementTop {
   margin: 0.5rem 0;
 }
+
+.mui-radio-group {
```
akphi, line 68: Let's move this block up to the top of the file, this section belongs to overriding global MUI styles

packages/legend-query/src/components/QueryBuilderPropertyExpressionEditor.tsx
```diff
@@ -97,11 +113,23 @@ const DerivedPropertyParameterValueEditor = observer(
...
-        generateValueSpecificationForParameter(
-          variable,
-          derivedPropertyExpressionState.queryBuilderState.graphManagerState
-            .graph,
-        ),
+        generateValueSpecificationForParameter(variable, graph),
+        derivedPropertyExpressionState.queryBuilderState.observableContext,
+      );
+    };
+    const updateValueSpecification = (val: ValueSpecification): void => {
```
akphi, line 120: Please rebase, I have corrected this.

packages/legend-query/src/components/QueryBuilderPropertyExpressionEditor.tsx
```diff
@@ -124,16 +152,19 @@ const DerivedPropertyParameterValueEditor = observer(
...
-            valueSpecification={parameterValue as ValueSpecification}
-            graph={
-              derivedPropertyExpressionState.queryBuilderState.graphManagerState
-                .graph
-            }
-            expectedType={
-              parameterValue?.genericType?.value.rawType ??
-              derivedPropertyExpressionState.propertyExpression.func.genericType
-                .value.rawType
+            valueSpecification={
```
akphi, line 155: might as well do `guaranteeNonNullable` here

packages/legend-query/src/models/protocols/pure/QueryBuilder_PureProtocolProcessorPlugin.ts
```diff
@@ -105,6 +108,69 @@ export class QueryBuilder_PureProtocolProcessorPlugin extends PureProtocolProces
...
             processingContext,
           );
+        } else if (
+          matchFunctionName(functionName, SUPPORTED_FUNCTIONS.TODAY) ||
+          matchFunctionName(
+            functionName,
+            SUPPORTED_FUNCTIONS.FIRST_DAY_OF_QUARTER,
+          )
+        ) {
+          const SFEWithValueSpecifications = V1_buildGenericFunctionExpression(
```
akphi, line 118: Sorry, let's rename all `SFEWithValueSpecifications` --> `expression`

packages/legend-query/src/stores/filterOperators/QueryBuilderFilterOperator_Equal.ts
```diff
@@ -46,7 +47,7 @@ import {
 } from '../QueryBuilderOperatorsHelper';
 
 export class QueryBuilderFilterOperator_Equal extends QueryBuilderFilterOperator {
-  getLabel(filterConditionState: FilterConditionState): string {
+  getLabel(): string {
```
akphi, line 50: Let's revert this

packages/legend-query/src/stores/postFilterOperators/QueryBuilderPostFilterOperatorHelper.ts
```diff
@@ -66,6 +66,7 @@ export const getColumnMultiplicity = (
 export const buildPostFilterConditionExpression = (
   filterConditionState: PostFilterConditionState,
   operator: QueryBuilderPostFilterOperator,
+  operatorFunctionFullPath?: string,
```
akphi, line 69: --> `operatorFunctionFullPath: string | undefined,`

Let's document this flag as well.

```ts
/**
 * If provided, this will be used to construct the simple
 * function expression for the function with the specified
 * name. If not provided, we will fall back to use the TDS column getter function expression. This is the case because with TDS, we are provided some filter-like operators, e.g. IS_NULL, IS_NOT_NULL, etc.
 */
```

packages/legend-query/src/stores/postFilterOperators/QueryBuilderPostFilterOperator_Equal.ts
```diff
@@ -90,8 +123,24 @@ export class QueryBuilderPostFilterOperator_Equal extends QueryBuilderPostFilter
...
+      ] as string[];
+
       // When changing the return type for LHS, the RHS value should be adjusted accordingly.
       // Numeric value is handled loosely because execution still works if a float (RHS) is assigned to an Integer property(LHS), etc.
+      // When LHS is of type DateTime, RHS is handled loosely since the operator could be changed to another pure function.
+      // e.g. is -> isOnDay()
+      if (
+        lhsType.path === PRIMITIVE_TYPE.DATETIME &&
+        type !== undefined &&
+        DATE_PRIMITIVE_TYPES.includes(type.path)
```
akphi, line 140: Please make sure you update this the way we discussed in `filterOperator`s

packages/legend-query/src/QueryBuilder_Const.ts
```diff
@@ -53,11 +53,28 @@ export enum SUPPORTED_FUNCTIONS {
   GREATER_THAN_EQUAL = 'meta::pure::functions::lang::tests::greaterThanEqual',
   IN = 'meta::pure::functions::collection::in',
   IS_EMPTY = 'meta::pure::functions::collection::isEmpty',
-
   LESS_THAN = 'meta::pure::functions::lang::tests::lessThan',
   LESS_THAN_EQUAL = 'meta::pure::functions::lang::tests::lessThanEqual',
   STARTS_WITH = 'meta::pure::functions::string::startsWith',
 
+  // operations
```
akphi, line 60: --> `// date-time value helper functions`

packages/legend-query/src/components/QueryBuilderValueSpecificationEditor.tsx
```diff
@@ -54,20 +61,78 @@ import {
...
+
+type TypeCheckOption = {
+  expectedType: Type;
+  // The match flag represents if a strict type match is forced or not.
+  // If match is true, it means that options in the date-capability-dropdown
+  // that are not of the same type or of child types will be filtered out or not.
+  match?: boolean;
+};
+
+type CustomDateInputOption = {
```
akphi, line 97: Let's inline and remove this type

packages/legend-query/src/components/QueryBuilderValueSpecificationEditor.tsx
```diff
@@ -54,20 +61,78 @@ import {
...
+  reservedCustomDates,
+  generateDateOptionForStartDayOfDateOption,
+  QUERY_BUILDER_DATE_DURATION_UNIT,
+  generateDateAdjustFunctionFromDateOption,
+  QUERY_BUILDER_DAY_OF_WEEK,
+} from '../stores/QueryBuilderDateTimeHelper';
+
+type TypeCheckOption = {
+  expectedType: Type;
+  // The match flag represents if a strict type match is forced or not.
```
akphi, line 91: Please use `block comment` so this shows up properly in `IDE`

packages/legend-query/src/components/QueryBuilderValueSpecificationEditor.tsx
```diff
@@ -54,20 +61,78 @@ import {
...
+        // disable transition
+        // NOTE: somehow, this is the only workaround we have, if for example
+        // we set `appear = true`, the tooltip will jump out of position
+        timeout: 0,
+      }}
+      title={
+        <div className="query-builder__tooltip__content">
+          <div className="query-builder__tooltip__item">
+            <div className="query-builder__tooltip__item__value">
+              More date options will be available by clicking the left button
```
akphi, line 126: I think it's better to make this an `edit` button, you can click on to trigger the popover, also, when you hover on it, you give users hint. We can do this later

packages/legend-query/src/components/QueryBuilderValueSpecificationEditor.tsx
```diff
@@ -534,75 +599,601 @@ export const LatestDatePrimitiveInstanceValueEditor: React.FC = () => (
   </div>
 );
 
+export const AbsoluteDateValueSpecificationEditor: React.FC<{
```
akphi, line 602: Could you do me a favor and remove `export` from everything in this file except for `QueryBuilderValueSpecificationEditor`?

packages/legend-query/src/components/QueryBuilderValueSpecificationEditor.tsx
```diff
@@ -534,75 +599,601 @@ export const LatestDatePrimitiveInstanceValueEditor: React.FC = () => (
...
+              graph={graph}
+              customDateAdjustOptionValue={generateDateOption(
+                valueSpecification,
+              )}
+              updateValueSpecification={updateValueSpecification}
+              setDateOptionValue={setDateOptionValue}
+            />
+          );
+        default:
+          return <></>;
```
akphi, line 1145: --> return `null`

packages/legend-query/src/components/QueryBuilderValueSpecificationEditor.tsx
```diff
@@ -534,75 +599,601 @@ export const LatestDatePrimitiveInstanceValueEditor: React.FC = () => (
...
+      </div>
+    </div>
+  );
+};
+
+export const CustomStartDayOfDateValueSpecificationEditor: React.FC<{
+  graph: PureModel;
+  customDateAdjustOptionValue: CustomDateOption;
+  updateValueSpecification: (val: ValueSpecification) => void;
+  setDateOptionValue: Dispatch<SetStateAction<CustomDateOption>>;
```
akphi, line 905: Let's not use `Dispatch` since it's getting deprecated

packages/legend-query/src/components/QueryBuilderValueSpecificationEditor.tsx
```diff
@@ -534,75 +599,601 @@ export const LatestDatePrimitiveInstanceValueEditor: React.FC = () => (
...
+          darkMode={true}
+        />
+      </div>
+    </div>
+  );
+};
+
+export const CustomPreviousDayOfWeekValueSpecificationEditor: React.FC<{
+  graph: PureModel;
+  customDateAdjustOptionValue: CustomDateOption;
```
akphi, line 964: let's reorder `customDateAdjustOptionValue` to be the first param. Sorry for this nitpickiness :pray:

packages/legend-query/src/components/QueryBuilderValueSpecificationEditor.tsx
```diff
@@ -534,75 +599,601 @@ export const LatestDatePrimitiveInstanceValueEditor: React.FC = () => (
...
+          darkMode={true}
+        />
+      </div>
+    </div>
+  );
+};
+
+export const CustomStartDayOfDateValueSpecificationEditor: React.FC<{
+  graph: PureModel;
+  customDateAdjustOptionValue: CustomDateOption;
```
akphi, line 903: let's reorder `customDateAdjustOptionValue` to be the first param. Sorry for this nitpickiness :pray:

packages/legend-query/src/components/QueryBuilderValueSpecificationEditor.tsx
```diff
@@ -534,75 +599,601 @@ export const LatestDatePrimitiveInstanceValueEditor: React.FC = () => (
...
+  ) => {
+    if (valueSpecification instanceof SimpleFunctionExpression) {
+      updateValueSpecification(
+        buildPrimitiveInstanceValue(
+          graph,
+          PRIMITIVE_TYPE.STRICTDATE,
+          event.target.value,
+        ),
+      );
+    } else {
```
akphi, line 629: Let's do `else if (valueSpecification instanceof InstanceValue)` in these cases

packages/legend-query/src/stores/QueryBuilderDateTimeHelper.ts
```diff
@@ -0,0 +1,587 @@
...
+  UnsupportedOperationError,
+} from '@finos/legend-shared';
+import {
+  DAY_OF_WEEK,
+  DURATION_UNIT,
+  SUPPORTED_FUNCTIONS,
+} from '../QueryBuilder_Const';
+import { buildPrimitiveInstanceValue } from './QueryBuilderOperatorsHelper';
+
+/**
```
akphi, line 41: Let's put the comment directly above each attribute.

Also, let's restructure this slightly:

```ts
export type CustomDateOption = {
  label: string;
  value: string;
  unit?: string | undefined;
  adjustment?: {
    durationRange: string;
    adjustment: string;
    startDate: string;
  } | undefined;
};
```

packages/legend-query/src/stores/QueryBuilderDateTimeHelper.ts
```diff
@@ -0,0 +1,587 @@
...
+  NOW = 'Now',
+  YESTERDAY = 'Yesterday',
+  ONE_YEAR_AGO = 'One Year Ago',
+  ONE_MONTH_AGO = 'One Month Ago',
+  ONE_WEEK_AGO = 'One Week Ago',
+  ABSOLUTE_DATE = 'Absolute Date',
+  ABSOLUTE_TIME = 'Absolute Time',
+  CUSTOM_DATE = 'Custom Date',
+  PREVIOUS_DAY_OF_WEEK = 'Previous ... of Week',
+  START_DAY_OF_DATE = 'Start of...',
```
akphi, line 76: --> `START_DAY_OF`

packages/legend-query/src/stores/QueryBuilderDateTimeHelper.ts
```diff
@@ -0,0 +1,587 @@
...
+      return QUERY_BUILDER_CUSTOM_DATE_START_DATE_OPTION.START_OF_WEEK;
+    default:
+      return `${funcName} is not supported`;
+  }
+};
+
+// Generate CustomDateOption based on the pure date adjust() function.
+// Transform CustomDateOption if it matches any preserved custom adjust date functions. e.g. One Month Ago..
+const generateDateOptionForAdjustFunction = (
+  pureDateAjustFunction: SimpleFunctionExpression,
```
akphi, line 425: Fix the typo `pureDateAdjustFunction`

packages/legend-query/src/stores/QueryBuilderDateTimeHelper.ts
```diff
@@ -0,0 +1,587 @@
...
+    label: '',
+    value: '',
+    durationRange: '',
+    unit: '',
+    adjustment: '',
+    startDate: '',
+  };
+};
+
+export const generateDateOptionForStartDayOfDateOption = (
```
akphi, line 491: --> `generateDateOptionForFirstDate`

packages/legend-query/src/stores/QueryBuilderDateTimeHelper.ts
```diff
@@ -0,0 +1,587 @@
...
+  if (
+    valueSpecification instanceof SimpleFunctionExpression &&
+    matchFunctionName(
+      valueSpecification.functionName,
+      SUPPORTED_FUNCTIONS.ADJUST,
+    )
+  ) {
+    return generateDateOptionForAdjustFunction(valueSpecification);
+  }
+  return {
```
akphi, line 481: Why do we do this?

packages/legend-query/src/stores/QueryBuilderDateTimeHelper.ts
```diff
@@ -0,0 +1,587 @@
...
+        label: QUERY_BUILDER_CUSTOM_DATE_START_DATE_OPTION.START_OF_QUARTER,
+        value: QUERY_BUILDER_CUSTOM_DATE_START_DATE_OPTION.START_OF_QUARTER,
+      };
+    case QUERY_BUILDER_DATE_DURATION_UNIT.YEAR:
+      return {
+        label: QUERY_BUILDER_CUSTOM_DATE_START_DATE_OPTION.START_OF_YEAR,
+        value: QUERY_BUILDER_CUSTOM_DATE_START_DATE_OPTION.START_OF_YEAR,
+      };
+    default:
+      return { label: '', value: '' };
```
akphi, line 516: Does it hurt if instead of returning something like this, we return `undefined`?

packages/legend-query/src/stores/QueryBuilderDateTimeHelper.ts
akphi: Also, why is `durationRange: string`? And we probably should rename `durationRange` to `duration`

akphi: we need to discuss about `startDate`, it's slightly confusing to me

packages/legend-query/src/components/QueryBuilderValueSpecificationEditor.tsx
```diff
@@ -534,75 +599,601 @@ export const LatestDatePrimitiveInstanceValueEditor: React.FC = () => (
...
-        {expectedType === date && (
-          <div className="query-builder-value-spec-editor__dropdown">
-            <CustomSelectorInput
-              placeholder="Choose a type..."
-              options={typeOptions}
-              onChange={changeType}
-              value={selectedType}
-              darkMode={true}
-            />
+        {(typeCheckOption.expectedType === date ||
```
akphi, line 1160: Given the amount of custom logic to do, I think we should move this into its own file, I'm thinking of `QueryBuilderCustomDatePicker` and we can move all logic from `QueryBuilderDateTimeHelper` there.

packages/legend-query/src/stores/postFilterOperators/QueryBuilderPostFilterOperatorHelper.ts
```diff
@@ -66,6 +66,12 @@ export const getColumnMultiplicity = (
 export const buildPostFilterConditionExpression = (
   filterConditionState: PostFilterConditionState,
   operator: QueryBuilderPostFilterOperator,
+  /**
+   * If provided, this will be used to construct the simple
+   * function expression for the function with the specified
+   * name. If not provided, we will fall back to use the TDS column getter function expression. This is the case because with TDS, we are provided some filter-like operators, e.g. IS_NULL, IS_NOT_NULL, etc.
```
akphi, line 72: chop down this comment pls

packages/legend-query/src/components/QueryBuilderCustomDatePicker.tsx
```diff
@@ -0,0 +1,1267 @@
...
+enum QUERY_BUILDER_CUSTOM_DATE_OPTION_REFERENCE_MOMENT {
+  TODAY = 'Today',
+  NOW = 'Now',
+  FIRST_DAY_OF_YEAR = 'Start of Year',
+  FIRST_DAY_OF_QUARTER = 'Start of Quarter',
+  FIRST_DAY_OF_MONTH = 'Start of Month',
+  FIRST_DAY_OF_WEEK = 'Start of Week',
+}
+
+/**
```
akphi, line 117: Update this comment

packages/legend-query/src/components/QueryBuilderCustomDatePicker.tsx
```diff
@@ -0,0 +1,1267 @@
...
+      | undefined,
+  ) {
+    super(label, value);
+    this.duration = duration;
+    this.unit = unit;
+    this.direction = direction;
+    this.referenceMoment = referenceMoment;
+  }
+
+  generateDisplayableLabel(): string {
```
akphi, line 157: --> `DisplayLabel`

packages/legend-query/src/components/QueryBuilderCustomDatePicker.tsx
```diff
@@ -0,0 +1,1267 @@
...
+
+class CustomFirstDayOfOption extends DatePickerOption {
+  unit: QUERY_BUILDER_FIRST_DAY_OF_UNIT | undefined;
+
+  constructor(
+    label: string,
+    value: string,
+    unit: QUERY_BUILDER_FIRST_DAY_OF_UNIT | undefined,
+  ) {
+    super(label, value);
```
akphi, line 179: I don't think we need to pass in the value here (maybe the label?) since it is always the same

packages/legend-query/src/components/QueryBuilderCustomDatePicker.tsx
```diff
@@ -0,0 +1,1267 @@
...
+    this.unit = unit;
+  }
+}
+
+class CustomPreviousDayOfWeekOption extends DatePickerOption {
+  day: QUERY_BUILDER_DAY_OF_WEEK;
+
+  constructor(label: string, value: string, day: QUERY_BUILDER_DAY_OF_WEEK) {
+    super(label, value);
```
akphi, line 188: I don't think we need to pass in the value here (maybe the label?) since it is always the same

packages/legend-query/src/components/QueryBuilderCustomDatePicker.tsx
```diff
@@ -0,0 +1,1267 @@
...
+  graph: PureModel,
+  datePickerOption: DatePickerOption,
+): SimpleFunctionExpression => {
+  const multiplicityOne = graph.getTypicalMultiplicity(
+    TYPICAL_MULTIPLICITY_TYPE.ONE,
+  );
+  const strictDate = graph.getPrimitiveType(PRIMITIVE_TYPE.STRICTDATE);
+  const date = graph.getPrimitiveType(PRIMITIVE_TYPE.DATE);
+  const dateTime = graph.getPrimitiveType(PRIMITIVE_TYPE.DATETIME);
+  switch (datePickerOption.value) {
```
akphi, line 241: Here I think we can use the `instanceof` check so we don't end up with the `default` case and have to throw error.

packages/legend-query/src/components/QueryBuilderCustomDatePicker.tsx
```diff
@@ -0,0 +1,1267 @@
...
+      throw new UnsupportedOperationError(
+        `${datePickerOption.value} is not supported yet.`,
+      );
+  }
+};
+
+/**
+ * Generate the enum value of type Pure Enum, DURATION_UNIT, based on the input string.
+ */
+const generatePureDurationEnumValueFromString = (
```
akphi, line 389: To make it easier to analyze and read the code, let's stick to the naming convention:

`valueSpec -> UI state` -> process...
`UI state -> valueSpec` -> build...

packages/legend-query/src/components/QueryBuilderCustomDatePicker.tsx
```diff
@@ -0,0 +1,1267 @@
...
+    case SUPPORTED_FUNCTIONS.FIRST_DAY_OF_YEAR:
+      return QUERY_BUILDER_CUSTOM_DATE_OPTION_REFERENCE_MOMENT.FIRST_DAY_OF_YEAR;
+    case SUPPORTED_FUNCTIONS.FIRST_DAY_OF_QUARTER:
+      return QUERY_BUILDER_CUSTOM_DATE_OPTION_REFERENCE_MOMENT.FIRST_DAY_OF_QUARTER;
+    case SUPPORTED_FUNCTIONS.FIRST_DAY_OF_MONTH:
+      return QUERY_BUILDER_CUSTOM_DATE_OPTION_REFERENCE_MOMENT.FIRST_DAY_OF_MONTH;
+    case SUPPORTED_FUNCTIONS.FIRST_DAY_OF_WEEK:
+      return QUERY_BUILDER_CUSTOM_DATE_OPTION_REFERENCE_MOMENT.FIRST_DAY_OF_WEEK;
+    default:
+      throw new UnsupportedOperationError(`${funcName}is not supported`);
```
akphi, line 559: miss a space

packages/legend-query/src/components/QueryBuilderCustomDatePicker.tsx
```diff
@@ -0,0 +1,1267 @@
...
+    : Object.values(QUERY_BUILDER_CUSTOM_DATE_PICKER_OPTION);
+  const [datePickerOption, setDatePickerOption] = useState(
+    generateDatePickerOption(valueSpecification),
+  );
+  const [anchorEl, setAnchorEl] = useState<HTMLButtonElement | null>(null);
+  const openCustomDatePickerPopover = (
+    event: React.MouseEvent<HTMLButtonElement>,
+  ): void => {
+    setAnchorEl(event.currentTarget);
+    setDatePickerOption(generateDatePickerOption(valueSpecification));
```
akphi, line 1124: Let's move this call to the `onEnter` event handler of the `Popover`

packages/legend-query/src/components/QueryBuilderCustomDatePicker.tsx
```diff
@@ -0,0 +1,1250 @@
...
+  FIRST_DAY_OF_MONTH = 'Start of Month',
+  FIRST_DAY_OF_WEEK = 'Start of Week',
+}
+
+/**
+ * DatePickerOption is the base class being used to display and generate the corresponding pure date function.
+ */
+class DatePickerOption {
+  /**
+   * @label is the text that shows up in the valueSpecification box.
```
akphi, line 122: TODO: remove the `@` and comment on all of them form the subclasses

packages/legend-query/src/components/QueryBuilderCustomDatePicker.tsx
```diff
@@ -0,0 +1,1250 @@
...
+        );
+        firstDayOfWeekSFE.genericType = GenericTypeExplicitReference.create(
+          new GenericType(date),
+        );
+        return firstDayOfWeekSFE;
+      }
+    }
+  } else {
+    if (
+      datePickerOption.value === QUERY_BUILDER_CUSTOM_DATE_PICKER_OPTION.TODAY
```
akphi, line 314: Let's use `switch...case`

akphi (dismissed):

Issue #1037, opened 2022-04-07
Bug: Graph modifiers not working in editors post `mobx` refactoring


akphi:
### Similar issues

- [X] I have searched and found no existing similar issues

### How are you using Studio?

Legend Studio, Legend Query

### Current and expected behavior

This is expected and caused by #1000 
We're dealing with the spots here

- [x] Query Builder
  - [x] Parameter
  - [x] Value Specification Editor
  - [x] Add projection, etc.
  - [x] Filter, etc.
  - [x] Post-filter
  - [x] Derivation
  - [x] Graph Fetch
  - [x] SubType
  - [x] exists()
- [x] UML editor
  - [x] Profile
  - [x] Enumeration
  - [x] Class
  - [x] Association
  - [x] Function
- [x] Text
- [x] Diagram editor
  - [x] DnD
- [x] File Generation
- [x] Generation Spec
- [x] Binding/Schema Set
- [x] Connection editor
  - [x] Model Connection
  - [x] FlatData
  - [x] Relational
- [x] Runtime editor
- [x] Service editor
  - [x] Execution
  - [x] Test
- [x] Mapping editor
  - [x] Operation
  - [x] Pure
  - [x] Relational
  - [x] FlatData
  - [x] Enumeration
  - [x] Mapping test
- [x] Studio app
  - [x] Create element
  - [x] Create package
  - [x] Delete element
  - [x] Rename element
- [x] Database
- [x] FlatData Store
- [x] Dataspace

### Environment

```text
N/A
```

### Contribution

- [ ] I would like to work on the fix for this issue

YannanGao-gs:
Bugs expected and caused by https://github.com/finos/legend-studio/pull/1000 in @finos/legend-studio@5.1.4 : https://github.com/finos/legend-studio/issues/1050

akphi:
> Bugs related to this found in query builder
> 
> * [ ]  Query builder's Derived Property name can't be reset by clicking the 'reset icon'
> 
> <img alt="bug1" width="1463" src="https://user-images.githubusercontent.com/73408381/162840457-689ada2f-4804-46ae-8b77-4c1dd26486dd.png">
> 
> * [ ]  Query builder's parameter can't be switched to a new type by selecting an option in the dropdown
> 
> <img alt="bug2" width="600" src="https://user-images.githubusercontent.com/73408381/162840578-b19cbd10-215c-4dcd-9694-b6a2b9ca77bd.png">
> 
[...]

PR #1049, opened 2022-04-11
Observe ValueSpecification for Query Builder editors

YannanGao-gs:
## Summary

A bunch of fixes to address #1037 
Fixes https://github.com/finos/legend-studio/issues/1008

- [ ] Test(s) added

https://user-images.githubusercontent.com/73408381/162839540-6e97e1e8-d800-417c-94f7-7ef96928e71f.mov

### DIFF
--- a/.changeset/rotten-flies-tie.md
+++ b/.changeset/rotten-flies-tie.md
@@ -1,0 +1,5 @@
+---
+'@finos/legend-query': patch
+---
+
+Observe `ValueSpecification` for query builder editors.
--- a/.changeset/rotten-flies-tie.md
+++ b/.changeset/rotten-flies-tie.md
@@ -1,5 +1,3 @@
 ---
 '@finos/legend-query': patch
 ---
-
-Observe `ValueSpecification` for query builder editors.
--- a/packages/legend-graph/src/graphManager/action/changeDetection/ValueSpecificationObserver.ts
+++ b/packages/legend-graph/src/graphManager/action/changeDetection/ValueSpecificationObserver.ts
@@ -14,8 +14,14 @@
  * limitations under the License.
  */
 
+import { isType, Pair } from '@finos/legend-shared';
 import { computed, makeObservable, observable } from 'mobx';
-import type { AlloySerializationConfigInstanceValue } from '../../../models/metamodels/pure/valueSpecification/AlloySerializationConfig';
+import { PackageableElementReference } from '../../../models/metamodels/pure/packageableElements/PackageableElementReference';
+import { Runtime } from '../../../models/metamodels/pure/packageableElements/runtime/Runtime';
+import {
+  type AlloySerializationConfigInstanceValue,
+  AlloySerializationConfig,
+} from '../../../models/metamodels/pure/valueSpecification/AlloySerializationConfig';
 import {
   type RootGraphFetchTreeInstanceValue,
   type PropertyGraphFetchTreeInstanceValue,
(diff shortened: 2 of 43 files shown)


packages/legend-query/src/stores/QueryParametersState.ts
```diff
@@ -114,7 +115,7 @@ export class QueryParameterState {
       mockParameterValue: action,
     });
     this.queryParameterState = queryParameterState;
-    this.parameter = variableExpression;
+    this.parameter = observe_VariableExpression(variableExpression);
```
MauricioUyaguari, line 118: Ideally the observable should already be done before calling the constructor. I think it might be better to do.
new QueryParameterState(queryParameterState, observe_VariableExpression(variableExpression)), instead of adding the logic in the constructor.

packages/legend-query/src/stores/QueryBuilderPropertyEditorState.ts
```diff
@@ -205,6 +203,10 @@ export class QueryBuilderDerivedPropertyExpressionState {
   ) {
     this.path = getPropertyPath(propertyExpression);
     this.title = getPropertyChainName(propertyExpression, true);
+    observe_AbstractPropertyExpression(
```
MauricioUyaguari, line 206: Same comment, ideally we should call this observable before new QueryBuilderDerivedPropertyExpressionState...

packages/legend-query/src/components/QueryBuilderFilterPanel.tsx
```diff
@@ -261,7 +262,11 @@ const QueryBuilderFilterConditionEditor = observer(
                 </div>
               )}
               <QueryBuilderValueSpecificationEditor
-                valueSpecification={node.condition.value}
+                valueSpecification={observe_ValueSpecification(
```
MauricioUyaguari, line 265: I think this works but ideally the way I see it is.
For value specification coming from the orginal `RawLambda`, those should be captured in the  observable after building the query state i.e QueryBuilderState `initialize` function. 
For value specifications build in the editor we should trace when we create new observable elements and wrap them with observables upon creation. 
For example here, we should wrap the observable when the FilterConditionState is created as this is where we create the new `AbstractPropertyExpression` in `buildPropertyExpressionFromExplorerTreeNodeData`

MauricioUyaguari (changes requested):

.changeset/rotten-flies-tie.md
```diff
@@ -0,0 +1,5 @@
+---
+'@finos/legend-query': patch
+---
+
+Observe `ValueSpecification` for query builder editors.
```
akphi, line 5: Let's get rid of this, this comes as fixing problem coming from https://github.com/finos/legend-studio/issues/1037

packages/legend-graph/src/graphManager/action/changeDetection/ValueSpecificationObserver.ts
```diff
@@ -60,6 +61,7 @@ import {
 const observe_Abstract_ValueSpecification = (
   metamodel: ValueSpecification,
 ): void => {
+  observe_Multiplicity(metamodel.multiplicity);
```
akphi, line 64: Let's move this line down after the `makeObservable(...)` block. Thanks!

packages/legend-graph/src/graphManager/action/changeDetection/ValueSpecificationObserver.ts
```diff
@@ -68,6 +70,9 @@ const observe_Abstract_ValueSpecification = (
 export const observe_VariableExpression = skipObserved(
   (metamodel: VariableExpression): VariableExpression => {
     observe_Abstract_ValueSpecification(metamodel);
+    if (metamodel.genericType) {
```
akphi, line 73: Why not do this in `observe_Abstract_ValueSpecification`?

MauricioUyaguari (approved):

PR #1065, opened 2022-04-14
Update filter and post filter RHS value when changing LHS type

YannanGao-gs:
## Summary
Fixes https://github.com/finos/legend-studio/issues/1094

- [ ] Test(s) added

https://user-images.githubusercontent.com/73408381/163410420-96f63b06-513f-4835-868e-21d00e87004b.mov

### DIFF
--- a/.changeset/twenty-flies-smell.md
+++ b/.changeset/twenty-flies-smell.md
@@ -1,0 +1,5 @@
+---
+'@finos/legend-query': patch
+---
+
+Fix a problem with adjusting (post) filter's RHS value when changing type for LHS
--- a/packages/legend-query/src/stores/filterOperators/QueryBuilderFilterOperator_Equal.ts
+++ b/packages/legend-query/src/stores/filterOperators/QueryBuilderFilterOperator_Equal.ts
@@ -84,22 +84,22 @@ export class QueryBuilderFilterOperator_Equal extends QueryBuilderFilterOperator
     const type = filterConditionState.value
       ? getNonCollectionValueSpecificationType(filterConditionState.value)
       : undefined;
+
+    const numericPrimitiveTypes = [
+      PRIMITIVE_TYPE.NUMBER,
+      PRIMITIVE_TYPE.INTEGER,
+      PRIMITIVE_TYPE.DECIMAL,
+      PRIMITIVE_TYPE.FLOAT,
+    ] as string[];
+
+    // When changing the return type for LHS, the RHS value should be adjusted accordingly.
+    // Numeric value is handled loosely because execution still works if a float (RHS) is assigned to an Integer property(LHS), etc.
     return (
       type !== undefined &&
-      ((
-        [
-          PRIMITIVE_TYPE.STRING,
-          PRIMITIVE_TYPE.BOOLEAN,
-          PRIMITIVE_TYPE.NUMBER,
-          PRIMITIVE_TYPE.INTEGER,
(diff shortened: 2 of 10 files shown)


packages/legend-query/src/stores/filterOperators/QueryBuilderFilterOperator_Equal.ts
```diff
@@ -84,22 +84,20 @@ export class QueryBuilderFilterOperator_Equal extends QueryBuilderFilterOperator
...
-          PRIMITIVE_TYPE.INTEGER,
-          PRIMITIVE_TYPE.DECIMAL,
-          PRIMITIVE_TYPE.FLOAT,
-          PRIMITIVE_TYPE.DATE,
-          PRIMITIVE_TYPE.STRICTDATE,
-          PRIMITIVE_TYPE.DATETIME,
-        ] as string[]
-      ).includes(type.path) ||
-        type === propertyType)
+      ((numericPrimitiveTypes.includes(type.path) &&
```
akphi, line 97: I thought we can use `isSuperType` for number types as well?

akphi: Also, what about other operators? Why only `equal`?

.changeset/twenty-flies-smell.md
```diff
@@ -0,0 +1,5 @@
+---
+'@finos/legend-query': patch
+---
+
+Update filter and post filter RHS value when changing LHS type.
```
akphi, line 5:
```suggestion
Fix a problem with adjusting (post) filter's RHS value when changing type for LHS.
```

akphi (changes requested):

packages/legend-query/src/stores/filterOperators/QueryBuilderFilterOperator_Equal.ts
YannanGao-gs: For number, i don't think we could because we could assign integer to float, etc.

akphi: you mean the LHS can be narrower than the RHS, i.e. `someIntegerProperty == 123.123`? Did you execute to confirm this?

akphi: Like said, let's check. I don't think the following could work:

```
LHS (Integer) == RHS (Float)
```

Can you check?

akphi: Ok, so this is not straight-forward behavior at all. Let's document this in a comment here about why we do this.

packages/legend-query/src/stores/filterOperators/QueryBuilderFilterOperator_Equal.ts
```diff
@@ -84,22 +84,22 @@ export class QueryBuilderFilterOperator_Equal extends QueryBuilderFilterOperator
     const type = filterConditionState.value
       ? getNonCollectionValueSpecificationType(filterConditionState.value)
       : undefined;
+
+    const numericPrimitiveTypes = [
```
akphi, line 88: Let's capitalize this. Also probably should rephrase the comment a bit

akphi (approved):

PR #1101, opened 2022-04-23
include query parameter when generating service test data

MauricioUyaguari:
## Summary

Include Query parameter data when generating service data. 
Add support to anonymize data for service test. 

- [ ] Test(s) added

image

### DIFF
--- a/.changeset/fuzzy-shrimps-double.md
+++ b/.changeset/fuzzy-shrimps-double.md
@@ -1,0 +1,5 @@
+---
+'@finos/legend-graph': patch
+---
+
+Fix a problem with generating test data by properly including query parameter. Also, we now make use of the `test-data` generation's feature to `anonymize` sensitive data if needed.
--- a/.changeset/nothing-too-small.md
+++ b/.changeset/nothing-too-small.md
@@ -1,0 +1,5 @@
+---
+'@finos/legend-art': patch
+---
+
+Add `MaskIcon`.
--- a/.changeset/nothing-too-small.md
+++ b/.changeset/nothing-too-small.md
@@ -1,5 +1,3 @@
 ---
 '@finos/legend-art': patch
 ---
-
-Add `MaskIcon`.
--- a/packages/legend-graph/src/graphManager/AbstractPureGraphManager.ts
+++ b/packages/legend-graph/src/graphManager/AbstractPureGraphManager.ts
@@ -315,6 +315,11 @@ export abstract class AbstractPureGraphManager {
     lambda: RawLambda,
     runtime: Runtime,
     clientVersion: string,
+    parameters: (string | number | boolean)[],
+    options?: {
+      // Anonymizes data by hashing any string values in the generated data
+      anonymizeGeneratedData?: boolean;
+    },
   ): Promise<string>;
 
   // ------------------------------------------- Service -------------------------------------------
--- a/packages/legend-graph/src/models/protocols/pure/v1/engine/execution/V1_ExecuteInput.ts
+++ b/packages/legend-graph/src/models/protocols/pure/v1/engine/execution/V1_ExecuteInput.ts
@@ -17,6 +17,7 @@
 import {
   createModelSchema,
   custom,
+  list,
   object,
   optional,
   primitive,
@@ -52,6 +53,28 @@ export class V1_ExecuteInput {
         () => SKIP,
       ),
       context: usingModelSchema(V1_rawBaseExecutionContextModelSchema),
+    }),
+  );
+}
+
+export class V1_TestDataGenerationExecutionInput extends V1_ExecuteInput {
+  parameters: (string | number | boolean)[] = [];
+  hashStrings = false;
+
+  // TODO: rename to serialization once `SerializationFactory` is rewritten to handle polymorphism
+  static _serialization = new SerializationFactory(
+    createModelSchema(V1_TestDataGenerationExecutionInput, {
+      clientVersion: optional(primitive()),
+      function: usingModelSchema(V1_rawLambdaModelSchema),
+      mapping: primitive(),
+      model: object(V1_PureModelContextData),
+      runtime: custom(
+        (val) => V1_serializeRuntime(val),
+        () => SKIP,
+      ),
+      context: usingModelSchema(V1_rawBaseExecutionContextModelSchema),
+      hashStrings: primitive(),
+      parameters: list(primitive()),
     }),
   );
 }
--- a/packages/legend-studio/src/components/editor/edit-panel/service-editor/ServiceTestEditor.tsx
+++ b/packages/legend-studio/src/components/editor/edit-panel/service-editor/ServiceTestEditor.tsx
@@ -47,6 +47,7 @@ import {
   InfoCircleIcon,
   MenuContentItem,
   MenuContent,
+  MaskIcon,
 } from '@finos/legend-art';
 import {
   type TestContainerState,
@@ -656,6 +657,10 @@ export const ServiceTestEditor = observer(
     const serviceState = editorStore.getCurrentEditorState(ServiceEditorState);
     const isReadOnly = serviceState.isReadOnly;
     // test data
+    const anonymizeGeneratedData = selectedTestState.anonymizeGeneratedData;
+    const toggleAnonymizeGeneratedData = (): void => {
+      selectedTestState.setAnonymizeGeneratedData(!anonymizeGeneratedData);
+    };
     const updateTestData = (val: string): void =>
       singleExecTest_setData(selectedTest, val);
     const generateTestData = (): void => {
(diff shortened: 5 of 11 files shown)


packages/legend-studio/src/stores/editor-state/element-editor-state/service/ServiceTestState.ts
```diff
@@ -497,39 +513,60 @@ export class SingleExecutionTestState {
...
-      }
-    }
-    if (generatedTestData) {
+    try {
+      this.isGeneratingTestData = true;
+      // NOTE: here, we attempt to use engine to generate test data.
+      // Once all types of generate data are supported we will move to just using engine
+      const executionInput =
+        this.serviceEditorState.executionState.serviceExecutionParameters;
+      assertNonNullable(
```
akphi, line 522: why not `guaranteeNonNullable` instead?

packages/legend-studio/src/stores/editor-state/element-editor-state/service/ServiceTestState.ts
```diff
@@ -497,39 +513,60 @@ export class SingleExecutionTestState {
...
-        `Can't auto-generate test data for service`,
-      );
+      this.editorStore.applicationStore.notifyError(error);
+    } finally {
+      this.isGeneratingTestData = false;
     }
-    this.isGeneratingTestData = false;
+  }
+
+  buildTestDataParameters(rawLambda: RawLambda): (string | number | boolean)[] {
```
akphi, line 546: Let's just make this an utility method instead, exposing through the class is dangerous as it can be mistakenly used by somebody else.

packages/legend-graph/src/graphManager/AbstractPureGraphManager.ts
```diff
@@ -314,6 +314,8 @@ export abstract class AbstractPureGraphManager {
     lambda: RawLambda,
     runtime: Runtime,
     clientVersion: string,
+    parameters: (string | number | boolean)[],
+    hashStrings: boolean,
```
akphi, line 318: Let's put `hashStrings` in an `options` field? and also let's document that a bit since the feature doesn't get reflected very clearly with the name now

packages/legend-graph/src/models/protocols/pure/v1/engine/execution/V1_ExecuteInput.ts
```diff
@@ -55,3 +56,24 @@ export class V1_ExecuteInput {
     }),
   );
 }
+
+export class V1_TestDataGenerationExecutionInput extends V1_ExecuteInput {
+  parameters: (string | number | boolean)[] = [];
+  hashStrings = false;
+
+  static _serialization = new SerializationFactory(
```
akphi, line 64: I believe we can do `static override serialization` here

.changeset/fuzzy-shrimps-double.md
```diff
@@ -0,0 +1,6 @@
+---
+'@finos/legend-graph': patch
+'@finos/legend-studio': patch
+---
+
+Include query parameter when generating service test data. Support hashing test data for service tests.
```
akphi, line 6: Let's split this changeset:

1. `@finos/legend-graph`: patch
```
Fix a problem with generating test data by properly including query parameter. Also, we now make use of the `test-data` generation's feature to `anonymize` sensitive data if needed.
```

2. `@finos/legend-studio`: patch
```
(on message)
```

packages/legend-studio/src/stores/editor-state/element-editor-state/service/ServiceTestState.ts
```diff
@@ -403,6 +413,7 @@ export class SingleExecutionTestState {
   selectedTestContainerState?: TestContainerState | undefined;
   isRunningAllTests = false;
   isGeneratingTestData = false;
+  hashGenerateTestData = true;
```
akphi, line 416: --> `anonymizeGeneratedTestData`

packages/legend-studio/src/components/editor/edit-panel/service-editor/ServiceTestEditor.tsx
```diff
@@ -683,13 +689,45 @@ export const ServiceTestEditor = observer(
...
+                      'panel__content__form__section__toggler',
+                      'service-execution-editor__test-data__hash-btn',
+                      {
+                        'panel__content__form__section__toggler--disabled':
+                          isReadOnly,
+                      },
+                    )}
+                    onClick={toggleHashTestData}
+                  >
+                    Hash Data
```
akphi, line 703: For this, we probably can be subtle and use a toggler icon instead of a full button with text, can we find some icon that means `hash/anonymize/security/password` or something?

packages/legend-studio/src/components/editor/edit-panel/service-editor/ServiceTestEditor.tsx
```diff
@@ -683,13 +689,45 @@ export const ServiceTestEditor = observer(
...
+                      className={clsx(
+                        'panel__content__form__section__toggler__btn',
+                        {
+                          'panel__content__form__section__toggler__btn--toggled':
+                            hashTestData,
+                        },
+                      )}
+                      disabled={isReadOnly}
+                      tabIndex={-1}
+                      title="Hash Test Data"
```
akphi, line 714: Let's also call this `anonymize` instead of `hash`

packages/legend-graph/src/graphManager/AbstractPureGraphManager.ts
akphi: Also, I think at this level, we should call this `anonymizeGeneratedData`

akphi (changes requested):

akphi:
@MauricioUyaguari can you include a screenshot?

packages/legend-studio/src/components/editor/edit-panel/service-editor/ServiceTestEditor.tsx
MauricioUyaguari: No, I think it makes it clearer is we just say what is does.

packages/legend-graph/src/models/protocols/pure/v1/engine/execution/V1_ExecuteInput.ts
MauricioUyaguari: Nope because of

// NOTE: due to contra-variance and variance, this class will not work for polymorphism
// hence, use `BasicSerializationFactory`


We could leverage  `BasicSerializationFactory`
but that would make redefined toJson, fromJson on every V1_ExecutionInput which is more used. Figured this is a cleaner approach. Maybe a note ?

packages/legend-graph/src/graphManager/AbstractPureGraphManager.ts
```diff
@@ -315,6 +315,9 @@ export abstract class AbstractPureGraphManager {
     lambda: RawLambda,
     runtime: Runtime,
     clientVersion: string,
+    parameters: (string | number | boolean)[],
+    // Anonymizes data by hashing any string values in the generated data
+    anonymizeGeneratedData?: boolean,
```
akphi, line 320: as said, this is a very specific feature of the data generation endpoint, we should put it in `options`

packages/legend-graph/src/models/protocols/pure/v1/engine/execution/V1_ExecuteInput.ts
akphi: sure, a note should be fine

packages/legend-studio/src/components/editor/edit-panel/service-editor/ServiceTestEditor.tsx
```diff
@@ -683,13 +689,48 @@ export const ServiceTestEditor = observer(
                   </div>
                 </div>
                 <div className="panel__header__actions">
+                  <div
```
akphi, line 692: This is a tiny feature people may not even need to know about them, why do we need to make it so apparent? Consider the mapping test/execution runner panel and this

akphi (changes requested):

packages/legend-graph/src/models/protocols/pure/v1/engine/execution/V1_ExecuteInput.ts
akphi: I just tried, looks like we can change `BasicSerializationFactory` to `SerializationFactory` for `V1_ExecuteInput`

akphi: Thanks to you I discovered some TS quirks I never faced before. If we rewrote `SerializationFactory` to

```ts

export class SerializationFactory<T> extends BasicSerializationFactory<T> {
  toJson(val: T): PlainObject<T> {
    return serialize(this.schema, val);
  }
  fromJson(val: PlainObject<T>): T {
    return deserialize(this.schema, val);
  }
}

```

we're good. Anyway, leave this as is, I'll deal with it

packages/legend-studio/src/components/editor/edit-panel/service-editor/ServiceTestEditor.tsx
MauricioUyaguari: We don't know this is a tiny feature. There is no clear cut Icon to make the distinction. I would rather make this apparent.

.changeset/nothing-too-small.md
```diff
@@ -0,0 +1,5 @@
+---
+'@finos/legend-art': patch
+---
+
+Add `MaskIcon`.
```
akphi, line 5:
```suggestion
```

akphi (approved):

PR #1212, opened 2022-06-13
Support property search in query builder

### LINKED ISSUE #782: Feature request: Query Builder Property Search
### Similar requests

### How are you using Studio?

Legend Studio

There are times when the class being queried has multiple properties. This can make it difficult to find a property as this problem can trickle down to other properties.

The goal is to provide users with a searchable UI to search for properties in the class graph and drag and drop directly into the fetch structure.

Property Search 
  - [ ] Search with depth to prevent infinite chain
  - [ ] Can search using humanized name or property name (have a checkbox)
  - [ ] Search with settings (multiplicity/type/etc.)

_No response_

_No response_

### Contribution

gayathrir11:
## Summary

Closes #782 

- [x] Test(s) added

https://user-images.githubusercontent.com/87973126/177756013-67f44e1c-27e4-452d-9e6d-0b030d1a077c.mp4


packages/legend-query/src/stores/QueryBuilderPropertySearchPanelState.ts
```diff
@@ -0,0 +1,398 @@
...
+
+  updateClassIncludedMappedProperties(): void {
+    this.isClassIncluded
+      ? this.setClassMappedPropertyNodes(this.searchedMappedPropertyNodes)
+      : this.removeClassPropertyNodes();
+  }
+
+  updateStringIncludedMappedProperties(): void {
+    this.isStringIncluded
+      ? this.setStringMappedPropertyNodes(this.searchedMappedPropertyNodes)
```
MauricioUyaguari, line 235: I find this code a bit confusing. You are setting all the `searchedMappedPropertyNodes` as the stringMappedPropertyNodes. You do this for types of nodes. This causes the method `filteredPropertyNodes` to be a bit confusing to read.
Couldn't these all be `computed` from the seachMappedPropertyNodes? 
You can then just use the flag of whether the user has checked off that type to decide whether to show it or not.

packages/legend-query/src/stores/QueryBuilderPropertySearchPanelState.ts
```diff
@@ -0,0 +1,398 @@
...
+        .forEach((node) => {
+          if (node.mappingData.mapped && !node.isPartOfDerivedPropertyBranch) {
+            currentLevelPropertyNodes.push(node);
+            this.allMappedPropertyNodes.push(node);
+          }
+        });
+    }
+    let currentDepth = 1;
+    const maxDepth = 15;
+    while (currentLevelPropertyNodes.length && currentDepth <= maxDepth) {
```
MauricioUyaguari, line 284: Do we handle loops in here. Like Firm -> Person -> Firm?

MauricioUyaguari (changes requested):
Will keep reviewing tomorrow but left some comments. 
2 additional thoughts based on the video. 
The `only` word is not clear that it is clickable. Maybe underline the word when hovering over it. Want to let users know that it is clickable. 
Also the results could use a bit of padding on the top. Seems too close to the search input.

packages/legend-query/src/stores/QueryBuilderPropertySearchPanelState.ts
gayathrir11: No, we don't handle loops in here. I think we should fix handling of loops in `mappingAlgo`

packages/legend-query/src/stores/QueryBuilderExplorerState.ts
```diff
@@ -447,7 +459,11 @@ export const getQueryBuilderPropertyNodeData = (
     propertyNode.childrenIds =
       generateExplorerTreeClassNodeChildrenIDs(propertyNode);
   }
-  return propertyNode;
+  if (!checkMappedness || propertyNode.mappingData.mapped) {
```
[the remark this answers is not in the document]
gayathrir11: Removing it

packages/legend-query/src/components/QueryBuilderPropertySearchPanel.tsx
```diff
@@ -0,0 +1,717 @@
...
+      propertySearchPanelState.setNumberIncluded(
+        !propertySearchPanelState.isNumberIncluded,
+      );
+    };
+    const toggleDateIncluded = (): void => {
+      propertySearchPanelState.setDateIncluded(
+        !propertySearchPanelState.isDateIncluded,
+      );
+    };
+    const onlyClassIncluded = (): void => {
```
MauricioUyaguari, line 405: maybe `setOnlyClassIncluded`

packages/legend-query/src/components/QueryBuilderPropertySearchPanel.tsx
```diff
@@ -0,0 +1,717 @@
...
+              subclass,
+              node,
+            );
+            treeData.nodes.set(subTypeTreeNodeData.id, subTypeTreeNodeData);
+          });
+        }
+      }
+      explorerState.refreshTree();
+    };
+    const showSelectedNodeInTree = (): void => {
```
MauricioUyaguari, line 184: should we close the search panel once the user selects to see the node in the tree?

MauricioUyaguari (changes requested):
could we add integration tests for this change. Your example in the video might make a good example for testing.

packages/legend-query/src/components/QueryBuilderPropertySearchPanel.tsx
```diff
@@ -0,0 +1,717 @@
...
+          guaranteeNonNullable(ancestors.pop()).id,
+        );
+        if (currentAncestor) {
+          onNodeSelect(currentAncestor);
+        }
+      }
+      const treeNode = treeData.nodes.get(node.id);
+      treeNode?.setIsSelected(true);
+      const nodeElement = document.getElementById(
+        `query-builder-tree-node-${node.id}`,
```
MauricioUyaguari, line 214: - `query-builder-tree-node` -> CONST
- lets add a comment on scrollIntoView.

packages/legend-query/src/components/QueryBuilderProjectionPanel.tsx
```diff
@@ -568,6 +568,14 @@ export const QueryBuilderProjectionPanel = observer(
           default:
             break;
         }
+        if (
```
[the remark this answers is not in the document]
gayathrir11: yes we do turn it on during `handleDrop`

packages/legend-query/src/components/QueryBuilderPropertySearchPanel.tsx
gayathrir11: we do close it once user selects to see the node in the tree

packages/legend-query/src/stores/QueryBuilderPropertySearchPanelState.ts
```diff
@@ -0,0 +1,310 @@
...
+        .slice(1)
+        .forEach((node) => {
+          if (node.mappingData.mapped && !node.isPartOfDerivedPropertyBranch) {
+            currentLevelPropertyNodes.push(node);
+            this.allMappedPropertyNodes.push(node);
+          }
+        });
+    }
+    let currentDepth = 1;
+    const maxDepth = 15;
```
akphi, line 231: let's make this a constant and put it in `QueryBuilder_Const`: `QUERY_BUILDER_PROPERTY_SEARCH_MAX_DEPTH = 15`

packages/legend-query/src/components/QueryBuilderProjectionPanel.tsx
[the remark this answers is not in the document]
akphi: I still don't get it why do we open search panel when we DnD something in this panel.

I thought the interaction goes like this:
1. User opens search property panel
2. User searches the property
3. User drags and drops the property to the projection panel

Why in (3) do we want to open the search panel again, hasn't it already been opened?

---

**EDITED** Check out my comment on usage of `useDrag` in `QueryBuilderPropertySearchPanel`

packages/legend-query/src/components/QueryBuilderGraphFetchTreePanel.tsx
```diff
@@ -219,8 +219,16 @@ export const QueryBuilderGraphFetchTreePanel = observer(
     const handleDrop = useCallback(
       (item: QueryBuilderExplorerTreeDragSource): void => {
         graphFetchState.addProperty(item.node);
+        if (
```
akphi, line 222: same comment as in projection panel

packages/legend-query/src/components/QueryBuilderFilterPanel.tsx
```diff
@@ -729,8 +729,16 @@ export const QueryBuilderFilterPanel = observer(
           ),
           undefined,
         );
+        if (
```
akphi, line 732: same comment as in projection panel

packages/legend-query/src/components/QueryBuilderPropertySearchPanel.tsx
```diff
@@ -0,0 +1,641 @@
...
+                  </button>
+                </>
+              )}
+            </div>
+            <div className="query-builder-property-search-panel__content">
+              <ResizablePanelGroup orientation="vertical">
+                <ResizablePanel size={150}>
+                  <div className="query-builder-property-search-panel__form__section">
+                    <div className="query-builder-property-search-panel__form__section__header__label">
+                      One many rows
```
akphi, line 436: --> `Multiple` because we change the text, let's add a hint icon to explain what this is about. I don't think `One-many rows` is descriptive enough

packages/legend-query/src/components/QueryBuilderPropertySearchPanel.tsx
```diff
@@ -0,0 +1,641 @@
...
+                        )}
+                      </button>
+                      <div className="query-builder-property-search-panel__form__section__toggler__prompt">
+                        Class
+                      </div>
+                      <div
+                        className="query-builder-property-search-panel__only__element"
+                        onClick={setOnlyClassIncluded}
+                      >
+                        only
```
akphi, line 498: I think this is a niche feature that we don't need for now. I would be more comfortable not having this implemented in the first round to reduce code complexity. If this proves to be a highly desirable feature, let's add it then. Sorry about this !

packages/legend-query/src/components/QueryBuilderPropertySearchPanel.tsx
```diff
@@ -0,0 +1,641 @@
...
+    .filter((p) => p !== '')
+    .map((p) => prettyPropertyName(p));
+  let propName = '';
+  propNameArray.forEach((p) => {
+    propName = `${propName}@${p}`;
+  });
+  return propName;
+};
+
+export const QueryBuilderTreeNodeViewer = observer(
```
akphi, line 105: why do we `export` this?

packages/legend-query/src/components/QueryBuilderPropertySearchPanel.tsx
```diff
@@ -0,0 +1,641 @@
...
+              ? QUERY_BUILDER_EXPLORER_TREE_DND_TYPE.ENUM_PROPERTY
+              : node.type instanceof Class
+              ? QUERY_BUILDER_EXPLORER_TREE_DND_TYPE.CLASS_PROPERTY
+              : QUERY_BUILDER_EXPLORER_TREE_DND_TYPE.PRIMITIVE_PROPERTY
+            : QUERY_BUILDER_EXPLORER_TREE_DND_TYPE.ROOT,
+        item: (): { node?: QueryBuilderExplorerTreePropertyNodeData } =>
+          node instanceof QueryBuilderExplorerTreePropertyNodeData
+            ? { node }
+            : {},
+        collect: (monitor): { isDragging: boolean } => ({
```
akphi, line 130: I think I understand why you do what you did in the `onDrop` of other panels now, but I think you can checkout this doc

https://react-dnd.github.io/react-dnd/docs/api/use-drag

Basically, you can use `end()` hook to check when dragging ends instead

akphi (changes requested):

packages/legend-query/src/components/QueryBuilderPropertySearchPanel.tsx
```diff
@@ -0,0 +1,595 @@
...
+      (node instanceof QueryBuilderExplorerTreePropertyNodeData &&
+        (node.property.multiplicity.upperBound === undefined ||
+          node.property.multiplicity.upperBound > 1)) ||
+      (node instanceof QueryBuilderExplorerTreeSubTypeNodeData &&
+        (node.multiplicity.upperBound === undefined ||
+          node.multiplicity.upperBound > 1));
+    queryBuilderState.explorerState.propertySearchPanelState.setIsSearchPanelOpen(
+      !isDragging,
+    );
+    const [isExpandable, setIsExpandable] = useState(false);
```
akphi, line 149: Let's move the state to top of the component

packages/legend-query/src/components/QueryBuilderPropertySearchPanel.tsx
```diff
@@ -0,0 +1,595 @@
...
+        collect: (monitor): { isDragging: boolean } => ({
+          isDragging: monitor.isDragging(),
+        }),
+        end: (): void => {
+          propertySearchPanelState.setIsSearchPanelOpen(true);
+        },
+      }),
+      [node],
+    );
+    const isMultiple =
```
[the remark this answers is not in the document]
gayathrir11: I have added it to the list of filterByOptions

packages/legend-query/src/stores/QueryBuilderPropertySearchPanelState.ts
```diff
@@ -0,0 +1,277 @@
...
+} from './QueryBuilderExplorerState.js';
+import type { QueryBuilderState } from './QueryBuilderState.js';
+
+export class QueryBuilderPropertySearchPanelState {
+  queryBuilderState: QueryBuilderState;
+  allMappedPropertyNodes: QueryBuilderExplorerTreeNodeData[] = [];
+  searchedMappedPropertyNodes: QueryBuilderExplorerTreeNodeData[] = [];
+  isSearchPanelOpen = false;
+  searchText = '';
+  filterByTypes: boolean[];
```
akphi, line 43: Oh, I think you missed my point, I was expecting:
```ts
filterByMultiple: boolean;
typeFilters: PRIMITIVE_TYPE[] (or PrimitiveType[])
```

packages/legend-query/src/components/QueryBuilderPropertySearchPanel.tsx
```diff
@@ -0,0 +1,595 @@
...
+      (node instanceof QueryBuilderExplorerTreePropertyNodeData &&
+        (node.property.multiplicity.upperBound === undefined ||
+          node.property.multiplicity.upperBound > 1)) ||
+      (node instanceof QueryBuilderExplorerTreeSubTypeNodeData &&
+        (node.multiplicity.upperBound === undefined ||
+          node.multiplicity.upperBound > 1));
+    queryBuilderState.explorerState.propertySearchPanelState.setIsSearchPanelOpen(
```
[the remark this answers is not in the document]
akphi: Hm, could you investigate if there's some sort of drag start hook? maybe the HTML ondragstart event? This seems like a strange usage to me.

akphi (changes requested):

packages/legend-query/src/components/QueryBuilderPropertySearchPanel.tsx
```diff
@@ -0,0 +1,595 @@
...
+    return (
+      <div>
+        <div
+          className="tree-view__node__container"
+          ref={dragConnector}
+          style={{
+            paddingLeft: `${(level - 1) * stepPaddingInRem + 0.5}rem`,
+            display: 'flex',
+          }}
+          onClick={(): void => setIsExpandable(!isExpandable)}
```
akphi, line 201: I believe we can use `onDragStart` here to trigger hidding the panel while dragging.

akphi (approved):

MauricioUyaguari (approved):

PR #1215, opened 2022-06-14
Feature/master dependency

ijeh-i:

- [ ] Test(s) added


akphi:
@ijeh-i could you rebase and resolve the conflicts, maybe then we could get the CI to build

.changeset/shaggy-squids-jam.md
```diff
@@ -1,5 +1,5 @@
 ---
-"@finos/legend-graph": patch
+'@finos/legend-graph': patch
 ---
 
 Fix populating of `extendsClassMappingId` for class mappings ([#1181](https://github.com/finos/legend-studio/issues/1181)).
```
akphi, line 5: There's no reason to touch these existing changesets.

packages/legend-server-sdlc/src/models/configuration/ProjectDependency.ts
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { createModelSchema, primitive } from 'serializr';
+import { createModelSchema, primitive, alias } from 'serializr';
```
akphi, line 17: Why can't we just change `versionId: VersionId` to `version: string`? What does it cost? The reason why I'm not 100% through with this change is that I don't see @kevin-m-knight-gs making the same sort of change in https://github.com/finos/legend-sdlc/pull/442

akphi (changes requested):

packages/legend-server-sdlc/src/models/configuration/ProjectDependency.ts
ijeh-i: The reason I didn't change the `versionId: VersionId ` to `version: string` is because of the external packages which might be reading the `versionId` class from the project Dependency.

If I change the versionId to a string then external projects would also need to make the same changes when reading it from the legend-studio

And the new sdlc payload structure require a string in place of the class object value.

akphi: That's fair point, but I think we can prioritize keeping this consistent with sdlc model. We will release this change in a major version

ijeh-i: Ok.. So should I change the  the `versionId: VersionId` to `versionId: string` so the we'll be consistent with the sdlc model ?
Or leave it as I've implemented ?

akphi: let's do `versionId: string`

packages/legend-studio/src/components/editor/edit-panel/project-configuration-editor/ProjectConfigurationEditor.tsx
```diff
@@ -232,10 +234,16 @@ const ProjectDependencyEditor = observer(
...
     const version = projectDependency.versionId;
     const versions = selectedProject?.versions ?? [];
-    const versionOptions = versions
+    let versionOptions = versions
       .slice()
       .sort((v1, v2) => compareSemVerVersions(v2, v1))
       .map((v) => ({ value: v, label: v }));
+
+    versionOptions = [
+      { label: 'Head', value: MASTER_SNAPSHOT },
```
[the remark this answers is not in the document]
MauricioUyaguari: should stay `MASTER_SNAPSHOT `

PR #1221, opened 2022-06-16
Properly handle `sourceId` and `targetId` in property mappings

### LINKED ISSUE #1202: Feature request: Properly handle `sourceId` in `legend-studio`
### Similar requests

### How are you using Studio?

_No response_

Right now, we have to disable a number of grammar roundtrip test

https://github.com/finos/legend-studio/blob/b700e8765b4a7411b751dba3bb3a7f8df533d527/packages/legend-manual-tests/src/__tests__/roundtrip-grammar/RoundtripGrammar.test.ts#L79-L99

This is due to the fact that we don't properly model the field `source` right in metamodel. 

We actually should use `reference`, i.e.

```ts
export abstract class PropertyMapping implements Hashable {
  ...
  /**
   * NOTE: in case the holder of this property mapping is an embedded property mapping,
   * that embedded property mapping is considered the source otherwise, it is always
   * the top/root `InstanceSetImplementation` that is considered the source implementation
   *
   * TODO: change this to use `SetImplemenetationReference`
   */
  sourceSetImplementation: SetImplementation;
  /**
   * NOTE: in Pure, we actually only store `targetId` and `sourceId` instead of the
   * reference but for convenience and graph completeness validation purpose we will
   * resolve to the actual set implementations here
   *
   * TODO: change this to use `OptionalSetImplemenetationReference`
   */
  targetSetImplementation?: SetImplementation | undefined;
  ...
```

_No response_

_No response_

### Contribution

gayathrir11:
## Summary

Closes #1202 

- [ ] Test(s) added

### DIFF
--- a/packages/legend-graph/src/models/metamodels/pure/packageableElements/mapping/PropertyMapping.ts
+++ b/packages/legend-graph/src/models/metamodels/pure/packageableElements/mapping/PropertyMapping.ts
@@ -18,7 +18,6 @@ import { hashArray, type Hashable } from '@finos/legend-shared';
 import { CORE_HASH_STRUCTURE } from '../../../../../MetaModelConst.js';
 import type { PropertyReference } from '../domain/PropertyReference.js';
 import type { PropertyMappingsImplementation } from './PropertyMappingsImplementation.js';
-import type { SetImplementation } from './SetImplementation.js';
 import type { PurePropertyMapping } from '../store/modelToModel/mapping/PurePropertyMapping.js';
 import type { FlatDataPropertyMapping } from '../store/flatData/mapping/FlatDataPropertyMapping.js';
 import type { EmbeddedFlatDataPropertyMapping } from '../store/flatData/mapping/EmbeddedFlatDataPropertyMapping.js';
@@ -29,6 +28,10 @@ import type { InlineEmbeddedRelationalInstanceSetImplementation } from '../store
 import type { AggregationAwarePropertyMapping } from './aggregationAware/AggregationAwarePropertyMapping.js';
 import type { XStorePropertyMapping } from './xStore/XStorePropertyMapping.js';
 import type { LocalMappingPropertyInfo } from './LocalMappingPropertyInfo.js';
+import type {
+  OptionalSetImplementationReference,
+  SetImplementationReference,
+} from './SetImplementationReference.js';
 
 export interface PropertyMappingVisitor<T> {
   visit_PropertyMapping(propertyMapping: PropertyMapping): T;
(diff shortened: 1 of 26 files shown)


packages/legend-graph/src/models/metamodels/pure/packageableElements/mapping/PropertyMapping.ts
```diff
@@ -70,35 +73,37 @@ export abstract class PropertyMapping implements Hashable {
...
+  sourceSetImplementation: SetImplementationReference;
   /**
    * NOTE: in Pure, we actually only store `targetId` and `sourceId` instead of the
    * reference but for convenience and graph completeness validation purpose we will
    * resolve to the actual set implementations here
    *
    * TODO: change this to use `OptionalSetImplemenetationReference`
    */
-  targetSetImplementation?: SetImplementation | undefined;
+  targetSetImplementation?: OptionalSetImplementationReference;
```
akphi, line 84: Since you have already used `OptionalSetImplementationReference`, there's no need to do `?:`

```suggestion
  targetSetImplementation: OptionalSetImplementationReference;
```

akphi: And also, let's get rid of the comments 🙏

packages/legend-graph/src/models/metamodels/pure/packageableElements/mapping/SetImplementationReference.ts
```diff
@@ -57,24 +59,36 @@ export class SetImplementationExplicitReference extends SetImplementationReferen
...
   }
+
+  get valueForSerialization(): string | undefined {
+    return this.value.id.value;
+  }
 }
 
 export class SetImplementationImplicitReference extends SetImplementationReference {
   override readonly ownerReference: PackageableElementImplicitReference<Mapping>;
+  readonly input?: string | undefined;
```
akphi, line 70: I think for these references, we need to properly observe them as well

MauricioUyaguari (changes requested):
for these type of changes it is very important to do a sanity check on everywhere we use `targetSetImplementation` and `sourceSetImplementation` to ensure we are now using the .value. Typescript check would capture most of these but PLEASE make sure you check ALL places as it will not capture them all. 
Example .sourceSetImplementation === x;

packages/legend-studio/src/stores/editor-state/element-editor-state/mapping/FlatDataInstanceSetImplementationState.ts
```diff
@@ -255,15 +259,34 @@ export abstract class FlatDataInstanceSetImplementationState extends InstanceSet
       this.mappingElement,
       PropertyExplicitReference.create(property),
       rootInstanceSetImplementation,
-      this.mappingElement,
+      SetImplementationImplicitReference.create(
```
akphi, line 262: Hm, I might need to re-think about this, after talking to @MauricioUyaguari I realized that this just doesn't make a lot of sense. We must not create `ImplicitReference` anywhere in the editor code, that should be done in graph builder code only.

akphi:
@gayathrir11 there's a problem with some unit/integration tests. You might need to run them locally to find the culprit, but apparently one of your changes caused a mem-leak (or so it seems). I would try to run test by package to narrow down the tests causing the issue. If the mem-leak is built up overtime and running them all cause the mem-leak, it's harder to debug, unfortunately.

packages/legend-graph/src/models/metamodels/pure/packageableElements/mapping/SetImplementationReference.ts
```diff
@@ -108,26 +124,39 @@ export class OptionalSetImplementationExplicitReference extends OptionalSetImple
   ): OptionalSetImplementationExplicitReference {
     return new OptionalSetImplementationExplicitReference(value);
   }
+
+  get valueForSerialization(): string | undefined {
```
akphi, line 128: The `create` method should allow `value: SetImplementation | undefined`

packages/legend-graph/src/models/metamodels/pure/packageableElements/mapping/SetImplementationReference.ts
```diff
@@ -41,6 +41,8 @@ export abstract class SetImplementationReference extends ReferenceWithOwner {
     this.ownerReference = ownerReference;
     this.value = value;
   }
+
```
akphi, line 44: I don't think we have properly `observe` this reference?

akphi:
Closing in favor of https://github.com/finos/legend-studio/pull/1324

PR #1231, opened 2022-06-22
update on the project dependency version to contain HEAD as a dependency version option

ijeh-i:
…ency version option

- [ ] Test(s) added


MauricioUyaguari:
@ijeh-i this ready for review? 
your build is not passing

akphi:
@MauricioUyaguari I think it is, can you also review this since it's a breaking change that needs SDLC coordination?

packages/legend-studio/src/stores/EditorStore.ts
```diff
@@ -790,24 +791,39 @@ export class EditorStore {
...
-        ),
+        this.sdlcServerClient.getEntities(projectId, activeWorkspace),
+        this.sdlcServerClient.getConfiguration(projectId, activeWorkspace),
       ])) as [Entity[], PlainObject<ProjectConfiguration>];
       entities = result[0];
       projectConfiguration = result[1];
+
+      const projectConfigurationPayload = projectConfiguration;
+      const projectDependencyList =
+        projectConfiguration.projectDependencies as {
```
MauricioUyaguari, line 805: Do we need this?
My understanding was going forward legend-sdlc will always give us thew new format when we ask for the dependency list. 
Hence it will give us versionId: string. 
It will be on us to understand this versionId string

packages/legend-studio/src/stores/sidebar-state/WorkspaceReviewState.ts
```diff
@@ -262,20 +262,47 @@ export class WorkspaceReviewState {
...
-      );
+      const originalProjectDependencies =
+        this.editorStore.projectConfigurationEditorState
+          .originalProjectConfiguration?.projectDependencies;
+      const listLength = originalProjectDependencies?.length ?? 0;
+      let versionError = false;
+      let i = 0;
+
+      // eslint-disable-next-line
+      const VERSION_REGEX = new RegExp('((0|([1-9]d*)).){2}(0|[1-9]d*)');
```
MauricioUyaguari, line 273: Could we just do a forEach and check the regex of each dependency to have x.x.x and if they don't throw the error. 
This is already covered in a `try` ,`catch` with the error being notified to the user regardless.

packages/legend-studio/src/components/editor/edit-panel/project-configuration-editor/ProjectConfigurationEditor.tsx
```diff
@@ -232,12 +233,18 @@ const ProjectDependencyEditor = observer(
...
     const version = projectDependency.versionId;
     const versions = selectedProject?.versions ?? [];
-    const versionOptions = versions
+    let versionOptions = versions
       .slice()
       .sort((v1, v2) => compareSemVerVersions(v2, v1))
       .map((v) => ({ value: v, label: v }));
+
+    versionOptions = [
+      { label: 'HEAD', value: SNAPSHOT_VERSION_ALIAS },
```
MauricioUyaguari, line 242: Don't we have the word `HEAD` as a constant somewhere? If we don't should probably live in legend-sdlc and be used here instead of writing it directly.

packages/legend-studio/src/components/editor/edit-panel/project-configuration-editor/ProjectConfigurationEditor.tsx
```diff
@@ -263,13 +270,17 @@ const ProjectDependencyEditor = observer(
     };
     const openProject = (): void => {
       if (!projectDependency.isLegacyDependency) {
+        const projectDependencyVersionId =
```
MauricioUyaguari, line 273: same comment as above

MauricioUyaguari (changes requested):

packages/legend-studio/src/stores/EditorStore.ts
MauricioUyaguari: @akphi pretty sure my statement here is corret but could you verify

packages/legend-studio/src/stores/sidebar-state/WorkspaceReviewState.ts
MauricioUyaguari: Also please don't disable `eslint`

packages/legend-server-sdlc/src/models/configuration/ProjectDependency.ts
```diff
@@ -42,13 +58,13 @@ export class ProjectDependency implements Hashable {
     });
 
     this.projectId = projectId;
-    this.versionId = versionId ?? new VersionId();
+    this.versionId = versionId ?? '0.0.1';
```
akphi, line 61: --> `0.0.0`

packages/legend-studio/src/stores/EditorStore.ts
akphi: yeah, I'm looking at @kevin-m-knight-gs PR https://github.com/finos/legend-sdlc/pull/442/files - https://github.com/finos/legend-sdlc/pull/442/files#diff-33676c119eb36f75a50e53a37b8e1359783afd139a84f0c985900a26c768e2c9

and I could see now `ProjectDependency` returns a string instead

akphi: That means you could also remove `convertVersionObjectToString` method

packages/legend-studio/src/components/editor/edit-panel/project-configuration-editor/ProjectConfigurationEditor.tsx
akphi: We could use `SNAPSHOT_VERSION_ALIAS` from `legend-server-depot` here

packages/legend-studio/src/stores/sidebar-state/WorkspaceReviewState.ts
```diff
@@ -262,20 +262,47 @@ export class WorkspaceReviewState {
...
+              workspaceId: this.sdlcState.activeWorkspace.workspaceId,
+              title,
+              workspaceType: this.sdlcState.activeWorkspace.workspaceType,
+              description,
+            },
+          )) as PlainObject<Review>,
+        );
+      } else {
+        this.editorStore.applicationStore.notifyError(
+          `Project dependency verison should not be set to HEAD when creating a review.`,
```
akphi, line 303: Let's use the constant for `HEAD` in this error message.

packages/legend-studio/src/stores/sidebar-state/WorkspaceReviewState.ts
akphi: Like Mauricio said, this check here could be simplified quite a bit.

akphi (changes requested):
For the changeset error, you can locally run `yarn changeset` and it should produce the changeset for you automatically. Remember to split the changeset as needed to report the change properly. This is technically a breaking change in `@finos/legend-server-sdlc`

FYI, you can look at the pipeline log for the ChangeSet check log to figure out what's wrong (i.e. which packages changed, but not reported)

packages/legend-studio/src/stores/sidebar-state/WorkspaceReviewState.ts
ijeh-i: ok.. i'll make the required updates

packages/legend-server-sdlc/src/models/configuration/ProjectDependency.ts
ijeh-i: I used 0.0.1 because i noticed that the default value for the version in the versionId class was 0.0.1

packages/legend-studio/src/stores/EditorStore.ts
akphi: How did you test this out btw?

akphi: I suspected that could be from an older SDLC server. With the new code, I don't really see how the version could be `non-string`

akphi: @ijeh-i could you do more testing then? If you don't find any case, let's proceed assuming that it's string only

ijeh-i: ok ... I'll test again and confirm

packages/legend-server-sdlc/src/models/configuration/ProjectDependencyUtils.ts
```diff
@@ -0,0 +1,27 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { isString } from '@finos/legend-shared';
+
+import type { VersionId } from '../version/VersionId.js';
+
+export const buildVersionId = (version: string | VersionId): string => {
```
MauricioUyaguari, line 21: why do we need this method. We can expect the result to come back as a string right?

packages/legend-studio/src/stores/EditorStore.ts
```diff
@@ -790,24 +791,39 @@ export class EditorStore {
...
+          projectId: string;
+          versionId: string;
+        }[];
+      projectConfigurationPayload.projectDependencies =
+        projectDependencyList.map(
```
MauricioUyaguari, line 810: you shouldn't need to do this. you should be able to just call `fromJSON` on this.

akphi:
Thanks @ijeh-i ! @MauricioUyaguari has made some updates based on the review discussion in this PR. So we're closing this in favor of https://github.com/finos/legend-studio/pull/1241

Issue #1119, opened 2022-04-29
Check full property expression tree  when checking if isEmpty/isNotEmpty operator is supported


YannanGao-gs:
### Similar requests

- [X] I have searched and found no existing similar requests

### How are you using Studio?

Legend Query

### What problems are you trying to solve?

Currently, we check if a property is optional by checking if the current property type is optional. We need to check if the full property expression is optional. 
If the property belongs to a class that is optional then itself is optional even if the property is required. 

### Describe the solution you would like to see implemented

Check property expression tree to see if the property is optional

### Contribution

- [ ] I would like to work on this feature

YannanGao-gs:
model 
```
###Service
Service model::MyService
{
  pattern: '/d2c48a9c-70fa-46e3-8173-c355e774004f';
  documentation: '';
  autoActivateUpdates: true;
  execution: Single
  {
    query: |model::Firm.all()->project([x|$x.employeeSize, x|$x.legalName], ['Employee Size', 'Legal Name']);
    mapping: model::NewMapping;
    runtime: model::Runtime;
  }
  test: Single
  {
    data: '';
    asserts:
    [
    ];
  }
}


###Relational
Database model::Test
(
  Table FirmTable
  (
    id INTEGER PRIMARY KEY,
    Legal_name VARCHAR(200)
  )
  Table PersonTable
  (
    id INTEGER PRIMARY KEY,
    firm_id INTEGER,
    firstName VARCHAR(200),
    lastName VARCHAR(200)
  )
  Table AddressTable
  (
    id INTEGER PRIMARY KEY,
    firm_id INTEGER,
    name VARCHAR(200)
  )

  Join FirmPerson(PersonTable.firm_id = FirmTable.id)
  Join FirmAddress(AddressTable.firm_id = FirmTable.id)
)


###Pure
Profile model::MyExtension
{
  stereotypes: [important];
  tags: [doc];
}

Enum model::IncType
{
  Corp,
  LLC
}

Class model::LegalEntity
{
  legalName: String[1];
  address: model::Address[0..1];
}

Class model::Street extends model::Address
{
  streetName: String[1];
}

Class <<model::MyExtension.important>> {model::MyExtension.doc = 'This is a model of a firm'} model::Firm extends model::LegalEntity
[
  validName: $this.legalName->startsWith('_')
]
{
  employees: model::Person[*];
  incType: model::IncType[1];
  employeeSize() {$this.employees->count()}: Number[1];
}

Class model::Person
{
[...]

PR #1249, opened 2022-06-29
Check full property expression tree when checking if isEmpty/isNotEmpty operator is supported

YannanGao-gs:
## Summary
Fixes https://github.com/finos/legend-studio/issues/1119
`isEmpty/isNotEmpty` operator will be available if and only if LHS's property is optional. As long as there is an optional property in the chain, LHS is optional.

- [ ] Test(s) added

demo

Complex chain (with subtype)

demo1

### DIFF
--- a/.changeset/shaggy-olives-carry.md
+++ b/.changeset/shaggy-olives-carry.md
@@ -1,0 +1,5 @@
+---
+'@finos/legend-query': patch
+---
+
+Check full property expression tree when checking if `isEmpty/isNotEmpty` operator is supported ([#1119](https://github.com/finos/legend-studio/issues/1119)).
--- a/packages/legend-query/src/stores/filterOperators/QueryBuilderFilterOperator_IsEmpty.ts
+++ b/packages/legend-query/src/stores/filterOperators/QueryBuilderFilterOperator_IsEmpty.ts
@@ -20,11 +20,10 @@ import {
   type FilterConditionState,
 } from '../QueryBuilderFilterState.js';
 import {
-  PRIMITIVE_TYPE,
   type ValueSpecification,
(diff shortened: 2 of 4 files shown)


.changeset/shaggy-olives-carry.md
```diff
@@ -0,0 +1,5 @@
+---
+'@finos/legend-query': patch
+---
+
+Check full property expression tree when checking if isEmpty/isNotEmpty operator is supported ([#1119](https://github.com/finos/legend-studio/issues/1119)).
```
akphi, line 5:
```suggestion
Check full property expression tree when checking if `isEmpty/isNotEmpty` operator is supported ([#1119](https://github.com/finos/legend-studio/issues/1119)).
```

packages/legend-query/src/stores/filterOperators/QueryBuilderFilterOperator_IsEmpty.ts
```diff
@@ -46,11 +47,24 @@ export class QueryBuilderFilterOperator_IsEmpty extends QueryBuilderFilterOperat
     const propertyType =
       filterConditionState.propertyExpressionState.propertyExpression.func
         .genericType.value.rawType;
-    const multiplicity =
-      filterConditionState.propertyExpressionState.propertyExpression.func
-        .multiplicity;
+    let isOptional = false;
+    let propertyExpression: ValueSpecification | undefined =
```
akphi, line 51: I think we can abstract this to an utility function and put that in `QueryBuilderOperatorsHelper`. Then both post-filter and filter logic can reuse that.

akphi (changes requested):

akphi (approved):

PR #1250, opened 2022-06-29
Support updated persistence model.

davidsunglee:
## Summary

Related engine changes are here:
https://github.com/finos/legend-engine/pull/763

These should be released together.

Corresponding legend documentation change:
https://github.com/finos/legend/pull/516

The persistence model in engine is changing to clearly separate:

* persistence specification (e.g. a relational sink with database table definitions, reusable across env and runtime) from
* env configuration (e.g. a database connection, particular to an env) from
* backend-specific properties (e.g. DPU count for AWS Glue, particular to runtime)

- [x] Test(s) added


packages/legend-extension-dsl-persistence/src/components/studio/DSLPersistence_LegendStudioPlugin.tsx
```diff
@@ -59,19 +68,22 @@ export class DSLPersistence_LegendStudioPlugin
   override getExtraRequiredDocumentationKeys(): string[] {
     return [
       DSL_PERSISTENCE_LEGEND_STUDIO_DOCUMENTATION_KEY.GRAMMAR_ELEMENT_PERSISTENCE,
+      DSL_PERSISTENCE_LEGEND_STUDIO_DOCUMENTATION_KEY.GRAMMAR_ELEMENT_PERSISTENCE_CONTEXT,
```
akphi, line 71: When you do this, you would introduce a documentation key to be verified at runtime if the corresponding documentation is missing. Please open a corresponding PR on `finos/legend` to add the documentation key - see https://github.com/finos/legend/pull/515 for example

packages/legend-manual-tests/src/__tests__/roundtrip-grammar/cases/DSLPersistence-basic.pure
```diff
@@ -9,14 +9,6 @@ Persistence org::legend::ZooPersistence
     sink: ObjectStorage
     {
       binding: org::legend::ZooBinding;
-      connection:
```
akphi, line 12: This test doesn't seem to cover `PersistenceContext`. Should we add one for that use case? I'd prefer not to complicate this test file further, so let's add another file for that if appropriate. Thanks!

akphi (changes requested):
Just left some minor comments, let's revisit this when `engine` change is in. Please keep us posted! Thanks

packages/legend-manual-tests/src/__tests__/roundtrip-grammar/cases/DSLPersistence-basic.pure
davidsunglee: Good catch.

Unfortunately, I think the <code>PersistenceContext</code> will require a valid <code>Persistence</code> and friends to compile. But will try to split up and see.

packages/legend-extension-dsl-persistence/src/components/studio/DSLPersistence_LegendStudioPlugin.tsx
akphi: Thank you so much!

akphi:
@davidsunglee oh forgot to give you an update, we're doing a bunch of patch releases to stabilize after the latest release, we will review and merge your PR after the storm passes. Thank you!

PR #1258, opened 2022-07-01
Fix rendering of global test runner when new testable elements are added

gayathrir11:

- [ ] Test(s) added

***Before***

https://user-images.githubusercontent.com/87973126/176812668-64d468ec-0b9e-4ffa-b051-ea85355f27fe.mp4

***After***

https://user-images.githubusercontent.com/87973126/176812718-98adc381-8890-45d9-9f05-9c511a589ad2.mp4

### DIFF
--- a/packages/legend-studio/src/stores/sidebar-state/testable/GlobalTestRunnerState.ts
+++ b/packages/legend-studio/src/stores/sidebar-state/testable/GlobalTestRunnerState.ts
@@ -496,14 +496,11 @@ export class GlobalTestRunnerState {
       .filter(isNonNullable);
   }
 
-  init(force?: boolean): void {
-    if (!this.testableStates || force) {
-      const testables =
-        this.editorStore.graphManagerState.graph.allOwnTestables;
-      this.testableStates = testables.map(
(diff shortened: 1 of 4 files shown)


packages/legend-studio/src/stores/sidebar-state/testable/GlobalTestRunnerState.ts
```diff
@@ -496,14 +496,11 @@ export class GlobalTestRunnerState {
...
 
-  init(force?: boolean): void {
-    if (!this.testableStates || force) {
-      const testables =
-        this.editorStore.graphManagerState.graph.allOwnTestables;
-      this.testableStates = testables.map(
-        (testable) => new TestableState(this.editorStore, this, testable),
-      );
-    }
+  init(): void {
```
MauricioUyaguari, line 499: This `int` is called within a `useEffect`, this would cause the whole test runner to refresh upon each render. 
Might make sense to alter the useEffect logic rather than change this logic.

gayathrir11:
Issue addressed in #1267

Issue #1204, opened 2022-06-08
Feature request: Add Filter by on TDS Result


MauricioUyaguari:
### Similar requests

- [X] I have searched and found no existing similar requests

### What problems are you trying to solve?

Add filter by option to a TDS row value that will generate a post filter with that value. This will make creating queries more user friendly. 

### Describe the solution you would like to see implemented

Right click context menu on row value that will add a post filter condition using the column and row value. 

### Contribution

- [ ] I would like to work on this feature

PR #1270, opened 2022-07-06
Add Filter by/out on TDS Result to post filter 

YannanGao-gs:
## Summary
Fixes https://github.com/finos/legend-studio/issues/1204

- [ ] Test(s) added

https://user-images.githubusercontent.com/73408381/177859454-de9ee8ea-d6ac-40f5-8f4a-3870e9ca9c70.mov

### DIFF
--- a/packages/legend-query/src/components/QueryBuilderResultPanel.tsx
+++ b/packages/legend-query/src/components/QueryBuilderResultPanel.tsx
@@ -23,33 +23,283 @@ import {
   MenuContent,
   MenuContentItem,
   CaretDownIcon,
+  ContextMenu,
+  clsx,
 } from '@finos/legend-art';
 import { observer } from 'mobx-react-lite';
 import { flowResult } from 'mobx';
 import type { QueryBuilderState } from '../stores/QueryBuilderState.js';
 import {
   type ExecutionResult,
+  type Enumeration,
+  InstanceValue,
   extractExecutionResultValues,
   TdsExecutionResult,
   RawExecutionResult,
   EXECUTION_SERIALIZATION_FORMAT,
+  EnumValueInstanceValue,
+  EnumValueExplicitReference,
 } from '@finos/legend-graph';
 import {
   ActionAlertActionType,
   ActionAlertType,
   EDITOR_LANGUAGE,
   ExecutionPlanViewer,
+  instanceValue_changeValue,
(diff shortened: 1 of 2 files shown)


YannanGao-gs:
- [ ] Confirmed with John Pentelow, @MauricioUyaguari, and @akphi, we all agree to add a post-filter for the filter by/out operation.
- [ ] Range Selection on cell won't be supported since it is enterprise-level. https://www.ag-grid.com/react-data-grid/range-selection/

packages/legend-query/src/components/QueryBuilderResultPanel.tsx
```diff
@@ -39,17 +42,244 @@ import {
...
+    const theProjectionColumnState =
+      queryBuilderState.fetchStructureState.projectionState.columns
+        .filter((c) => c.columnName === event?.column.getColId())
+        .concat(
+          queryBuilderState.fetchStructureState.projectionState.aggregationState.columns
+            .filter((c) => c.columnName === event?.column.getColId())
+            .map((ag) => ag.projectionColumnState),
+        )[0];
+
+    const getExistingPostFilterNode = (
```
MauricioUyaguari, line 99: Your method return nodes not a singular node. We can either rename the method or return a singular node

packages/legend-query/src/components/QueryBuilderResultPanel.tsx
```diff
@@ -39,17 +42,244 @@ import {
...
+            isFilterBy ? postFilterInOperator : postFilterNotInOperator,
+          );
+          (conditionState.value as InstanceValue).values = [
+            currentValueSpecificaton,
+            newValueSpecification,
+          ];
+        }
+      } else {
+        if (
+          !(
```
MauricioUyaguari, line 204: when do you encounter this case ?

YannanGao-gs: existing post filter: id is 1, user action: filter by id = 1 
existing post filter:  id is in [1 ,2 ], user action : filter by id = 1

MauricioUyaguari: as discussed, lets cleanup the if statements to make it easier to read.

MauricioUyaguari (approved):

PR #1277, opened 2022-07-08
add form support for multi execution service

MauricioUyaguari:

- [ ] Test(s) added


akphi:
I'm doing the code review, but here's some comments just from watching the video:

- Let's make the key icon smaller for the list item
- Let's make the `input` for the execution key name span the whole width
- Let's make the `exchange` icon smaller (the ratio should roughly similar to that of `Execute` button). Also, `Change` for the button label is a bit cryptic, though the color is noticeable enough, I think we should reword
- In terms of flow, I think the only thing I would change is when we swap from `single` -> `multi` is not asking for the execution key value, let's default it to `execContext_1`, so as to simplify the flow a bit

MauricioUyaguari:
for the `Change` word. I had thought of `Switch`
Or maybe show 'Multi` when in single and `Single` when in a multi.

akphi:
> for the `Change` word. I had thought of `Switch` Or maybe show 'Multi`when in single and`Single` when in a multi.

Is this action really that prominent? Like we would need to have a `primary` button like `execute` just to catch people's attention? Could we be more subtle and make this a normal toggle button (like the `hackermode` icon?). I think it's safer for now to do that, till we got more feedback that we need to make this more _in-the-face_, we could do that later. What do you think?

akphi:
> > > for the `Change` word. I had thought of `Switch` Or maybe show 'Multi`when in single and`Single` when in a multi.
> > 
> > 
> > Is this action really that prominent? Like we would need to have a `primary` button with text like `execute` just to draw people's attention? Could we be more subtle and make this a normal toggle button (like the `hackermode` icon?). I think it's safer, for now, to do that, till we got more feedback that we need to make this more _in-the-face_, we could do that later. What do you think?
> 
> Trouble is finding the right icon. or maybe use the icon I have with it :)

yeah, looks good to me!

akphi:
Um, actually no, the icon you use is for the action `switch`, this is a toggle, so maybe we need to change it to something indicating `multiple`, `branching`?

Just some ideas

`multiple`

image

`branching`

image

Source: https://react-icons.github.io/react-icons/search?q=split

akphi (approved):

Issue #1274, opened 2022-07-07
Feature request: Warn users that they lose their query changes when exiting query builder


YannanGao-gs:
### Similar requests

- [X] I have searched and found no existing similar requests

### How are you using Studio?

Legend Query

### What problems are you trying to solve?

Users can navigate back to Legend Studio without getting any warning that their changes may be lost.

### Describe the solution you would like to see implemented

Enable a pop up

### Contribution

- [X] I would like to work on this feature

PR #1278, opened 2022-07-08
Warn users that they lose their query changes when exiting query builder

YannanGao-gs:
## Summary
Fixes https://github.com/finos/legend-studio/issues/1274

- [x] Do not alert when exiting query builder opened by right-clicking a class
- [x] only alert users when there are changes in query builder opened from a mapping-test or service

- [ ] Test(s) added

https://user-images.githubusercontent.com/73408381/178505536-b26e446a-5189-4929-8caa-d9c9e4c731df.mov

### DIFF
--- a/packages/legend-graph/src/models/protocols/pure/v1/V1_PureGraphManager.ts
+++ b/packages/legend-graph/src/models/protocols/pure/v1/V1_PureGraphManager.ts
@@ -1463,8 +1463,15 @@ export class V1_PureGraphManager extends AbstractPureGraphManager {
   async pureCodeToLambda(
     lambda: string,
     lambdaId?: string,
+    options?: {
+      pruneSourceInformation?: boolean;
+    },
   ): Promise<RawLambda> {
-    const result = await this.engine.transformCodeToLambda(lambda, lambdaId);
+    const result = await this.engine.transformCodeToLambda(
+      lambda,
+      lambdaId,
+      options,
+    );
     return new RawLambda(result.parameters, result.body);
   }
 
--- a/packages/legend-query/src/stores/QueryBuilderState.ts
+++ b/packages/legend-query/src/stores/QueryBuilderState.ts
@@ -22,6 +22,7 @@ import {
   guaranteeNonNullable,
   guaranteeType,
   filterByType,
+  hashObject,
 } from '@finos/legend-shared';
 import {
   type QueryBuilderFilterOperator,
@@ -149,6 +150,24 @@ export class StandardQueryBuilderMode extends QueryBuilderMode {
   }
 }
 
+class QueryBuilderChangeDetectionState {
+  querybuildState: QueryBuilderState;
+  queryHashCode = hashObject(new RawLambda(undefined, undefined));
+  isEnabled = false;
+
+  constructor(queryBuilderState: QueryBuilderState) {
+    this.querybuildState = queryBuilderState;
+  }
+
+  setQueryHashCode(val: string): void {
+    this.queryHashCode = val;
+  }
+
+  setIsEnabled(val: boolean): void {
+    this.isEnabled = val;
+  }
+}
+
 export class QueryBuilderState {
   applicationStore: ApplicationStore<LegendApplicationConfig>;
   graphManagerState: GraphManagerState;
@@ -207,6 +226,7 @@ export class QueryBuilderState {
   showFunctionPanel = false;
   showParameterPanel = false;
   showPostFilterPanel = false;
+  changeDetectionState: QueryBuilderChangeDetectionState;
 
   constructor(
     applicationStore: ApplicationStore<LegendApplicationConfig>,
@@ -231,6 +251,7 @@ export class QueryBuilderState {
       showFunctionPanel: observable,
       showParameterPanel: observable,
       showPostFilterPanel: observable,
+      changeDetectionState: observable,
       setMode: action,
       resetQueryBuilder: action,
       resetQuerySetup: action,
@@ -266,6 +287,7 @@ export class QueryBuilderState {
     this.observableContext = new ObserverContext(
       this.graphManagerState.pluginManager.getPureGraphManagerPlugins(),
     );
+    this.changeDetectionState = new QueryBuilderChangeDetectionState(this);
   }
 
   setMode(val: QueryBuilderMode): void {
--- a/packages/legend-studio-extension-query-builder/src/components/EmbeddedQueryBuilder.tsx
+++ b/packages/legend-studio-extension-query-builder/src/components/EmbeddedQueryBuilder.tsx
@@ -25,10 +25,12 @@ import {
 } from '@finos/legend-art';
 import { useEditorStore } from '@finos/legend-studio';
 import { flowResult } from 'mobx';
-import { noop } from '@finos/legend-shared';
+import { hashObject, noop } from '@finos/legend-shared';
 import { QueryBuilder_EditorExtensionState } from '../stores/QueryBuilder_EditorExtensionState.js';
 import {
   useApplicationNavigationContext,
+  ActionAlertActionType,
+  ActionAlertType,
   useApplicationStore,
 } from '@finos/legend-application';
 import { QueryBuilder } from '@finos/legend-query';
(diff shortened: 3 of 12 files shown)


packages/legend-studio-extension-query-builder/src/components/QueryBuilderDialog.tsx
```diff
@@ -42,13 +46,37 @@ export const QueryBuilderDialog = observer(() => {
...
+    'Unsaved changes to your query will be lost if you continue. Are you sure?';
+
   const closeQueryBuilder = (): void => {
     flowResult(
       queryBuilderExtensionState.setEmbeddedQueryBuilderMode(undefined),
     ).catch(applicationStore.alertUnhandledError);
     queryBuilderExtensionState.reset();
   };
 
+  const confirmCloseQueryBuilder = (): void => {
```
akphi, line 59: @MauricioUyaguari with this, do we want to do change detection for the query?

packages/legend-studio-extension-query-builder/src/components/QueryBuilderDialog.tsx
```diff
@@ -42,13 +46,37 @@ export const QueryBuilderDialog = observer(() => {
...
+    applicationStore.setActionAlertInfo({
+      message: USER_ATTESTATION_MESSAGE,
+      type: ActionAlertType.CAUTION,
+      actions: [
+        {
+          label: 'Accept',
```
akphi, line 65: Also, let's change the language here to sound less like compliance training. We're simply asking the users if they are aware and want to proceed. So `Proceed`, `Cancel`

akphi (changes requested):

packages/legend-studio-extension-query-builder/src/components/QueryBuilderDialog.tsx
MauricioUyaguari: For this I think we would only warn when they query is coming from a persistent state. 
I.e service editor, mapping editor, saved query. 
We should calculate if there has been any changes then ask this if there has been.

YannanGao-gs: I monitored the query lambda. If there are any changes in lambda, it means that users make some changes.

packages/legend-studio-extension-query-builder/src/components/EmbeddedQueryBuilder.tsx
```diff
@@ -57,6 +72,36 @@ const QueryBuilderDialog = observer(() => {
     QUERY_BUILDER_LEGEND_STUDIO_APPLICATION_NAVIGATION_CONTEXT_KEY.EMBEDDED_QUERY_BUILDER,
   );
 
+  const confirmCloseQueryBuilder = (): void => {
+    const currentQuery = JSON.stringify(
```
akphi, line 76: Let's add a small note here

```ts
// NOTE: This is poor-man change detection for query
// in the future, we could consider a similar approach to how we do change detection in Studio
```

akphi (changes requested):

packages/legend-studio-extension-query-builder/src/components/EmbeddedQueryBuilder.tsx
```diff
@@ -45,7 +48,19 @@ const QueryBuilderDialog = observer(() => {
     QueryBuilder_EditorExtensionState,
   );
   const [isMaximized, setIsMaximized] = useState(false);
+  const [originalQuery] = useState(
```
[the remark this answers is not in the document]
akphi: Also, let's hash the JSON using `hashObject` function and compare the hash instead of doing `JSON.stringify` comparison.

packages/legend-studio-extension-query-builder/src/components/EmbeddedQueryBuilder.tsx
akphi, line 76: Let's do hash comparison instead

MauricioUyaguari:
seems like we have not thought about a couple of things: 

Do we want to initialize every query with change detection? Should it be part of the options when you initiate query builder (enableSoftChangeDetection)? 

Additionally, I don't think you handle when the user saves the query and then clicks exist. Are you updating the hashLambda when a user has hit save?

MauricioUyaguari (changes requested):
seems like we have not thought about a couple of things:

Do we want to initialize every query with change detection? Should it be part of the options when you initiate query builder (enableSoftChangeDetection)?

Additionally, I don't think you handle when the user saves the query and then clicks exist. Are you updating the hashLambda when a user has hit save?

YannanGao-gs:
1. Do we want to initialize every query with change detection? Should it be part of the options when you initiate query builder (enableSoftChangeDetection)?

       For the query builder opened by right-clicking a class, there is no alert anyhow. To achieve this, I default values of QueryBuilderChangeDetectionState.

2. Additionally, I don't think you handle when the user saves the query and then clicks exist. Are you updating the hashLambda when a user has hit save?
     hashLambda is not updated when a user hits save but it will be updated every time they reopen the query builder.

packages/legend-query/src/stores/QueryBuilderState.ts
```diff
@@ -149,6 +150,24 @@ export class StandardQueryBuilderMode extends QueryBuilderMode {
   }
 }
 
+class QueryBuilderChangeDetectionState {
+  hashOfLambda: string;
+  enableChangeDetector: boolean;
```
akphi, line 155: --> `isEnabled = false`

packages/legend-query/src/stores/QueryBuilderState.ts
```diff
@@ -149,6 +150,24 @@ export class StandardQueryBuilderMode extends QueryBuilderMode {
   }
 }
 
+class QueryBuilderChangeDetectionState {
+  hashOfLambda: string;
```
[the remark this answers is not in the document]
akphi: Please update the getter accordingly

packages/legend-query/src/stores/QueryBuilderState.ts
akphi: Please update the getter accordingly

packages/legend-query/src/stores/QueryBuilderState.ts
```diff
@@ -149,6 +150,24 @@ export class StandardQueryBuilderMode extends QueryBuilderMode {
...
+  enableChangeDetector: boolean;
+
+  constructor(hashOfLambda: string, enableChangeDetector: boolean) {
+    this.hashOfLambda = hashOfLambda;
```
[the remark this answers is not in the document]
akphi: let's simplify the constructor to not get any param for now except for `queryBuilderState: QueryBuilderState`

packages/legend-query/src/stores/QueryBuilderState.ts
```diff
@@ -263,6 +283,10 @@ export class QueryBuilderState {
     this.observableContext = new ObserverContext(
       this.graphManagerState.pluginManager.getPureGraphManagerPlugins(),
     );
+    this.changeDetectionState = new QueryBuilderChangeDetectionState(
+      hashObject(new RawLambda(undefined, undefined)),
```
akphi, line 287: Let's set the `hashCode` and the isEnabled` flag more declaratively here:

```ts
this.changeDetectionState = new QueryBuilderChangeDetectionState(this);
```

packages/legend-query/src/stores/QueryBuilderState.ts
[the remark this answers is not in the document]
akphi: you can initialize the hash value here

packages/legend-query/src/stores/QueryBuilderState.ts
```diff
@@ -276,10 +300,19 @@ export class QueryBuilderState {
   setShowFunctionPanel(val: boolean): void {
     this.showFunctionPanel = val;
   }
+
   setShowParameterPanel(val: boolean): void {
     this.showParameterPanel = val;
   }
 
+  updateChangeDetectionState(
```
akphi, line 308: Don't think we need this abstraction, let's remove it

akphi (changes requested):

akphi:
> I think we need this, couldnt we add this in the save actions the user does?  @akphi what do you think

Yes, this makes sense, and I totally missed this case. Thanks @MauricioUyaguari

YannanGao-gs:
> > I think we need this, couldnt we add this in the save actions the user does?  @akphi what do you think
> 
> Yes, this makes sense, and I totally missed this case. Thanks @MauricioUyaguari

I don't think it bothers us because queryBuilder will be reset(`queryBuilderExtension.reset()`) every time we click the edit icon. It means even though I save the latest lambda, it will be reset to initial value first then be updated based on latest query when we reopen the query builder.

akphi:
@YannanGao-gs you are relying on the implementation of the save action here, and now since this is a state in `QueryBuilderState`, it should be considered in a more generic fashion. For example, we need to think how this can work with standalone query builder. But I think for now, this will do, we can leave this be the way you have it

MauricioUyaguari (approved):

packages/legend-graph/src/models/protocols/pure/v1/V1_PureGraphManager.ts
```diff
@@ -1463,9 +1463,19 @@ export class V1_PureGraphManager extends AbstractPureGraphManager {
   async pureCodeToLambda(
     lambda: string,
     lambdaId?: string,
+    options?: {
+      pruneSourceInformation?: boolean;
+    },
   ): Promise<RawLambda> {
     const result = await this.engine.transformCodeToLambda(lambda, lambdaId);
-    return new RawLambda(result.parameters, result.body);
+    return options?.pruneSourceInformation
```
akphi, line 1471: no, let's make use the engine API to do the pruning for us

akphi (dismissed):

PR #893, opened 2022-02-28
Introduce extension mechanism consistently and exhaustively for classmapping

gayathrir11:
Closes #687 

## How did you test this change?

- [ ] Test(s) added
- [ ] Manual testing (please provide screenshots/recordings)
- [x] No testing (please provide an explanation)

PR #1286, opened 2022-07-11
Fix a bug introduced in #893

gayathrir11:
## Summary

When we have introduced extension mechanism for `class mappings` we have modified logic of `changeClassMappingSourceDriver` to check only the type of `setImplementation` when we change the source driver. It could be the case that we will change among various `setImplementations` when we change the source of a mapping element. We missed this scenario by removing checks on the type of newSource in the initial stage of creating a new driver which caused a bug that we are unable to add a relational db as source to a mapping element from form mode because we deafult new sorce setImp to pureInstance when nothing was mentioned as we removed checks for newSource we are unable to set source as relational db

- [ ] Test(s) added

https://user-images.githubusercontent.com/87973126/178327893-c122320e-0dd1-4b2d-b990-fbc692c20260.mp4

### DIFF
--- a/packages/legend-studio/src/stores/editor-state/element-editor-state/mapping/MappingEditorState.ts
+++ b/packages/legend-studio/src/stores/editor-state/element-editor-state/mapping/MappingEditorState.ts
@@ -899,19 +899,17 @@ export class MappingEditorState extends ElementEditorState {
       this.editorStore.pluginManager.getStudioPlugins(),
     );
     if (currentSource !== newSource) {
-      if (setImplementation instanceof PureInstanceSetImplementation) {
-        if (newSource instanceof Class || newSource === undefined) {
-          pureInstanceSetImpl_setSrcClass(setImplementation, newSource);
-        }
-      } else if (
-        setImplementation instanceof FlatDataInstanceSetImplementation
+      if (
+        setImplementation instanceof PureInstanceSetImplementation &&
+        (newSource instanceof Class || newSource === undefined)
       ) {
-        if (
-          newSource instanceof RootFlatDataRecordType &&
-          !getEmbeddedSetImplementations(setImplementation).length
-        ) {
-          flatData_setSourceRootRecordType(setImplementation, newSource);
-        }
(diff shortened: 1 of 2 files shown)


MauricioUyaguari (approved):

MauricioUyaguari:
@gayathrir11 seems like we gone back to the old logic. We may want to add comment on why we put all the conditions within the initial if statements. Otherwise I could see us going back to the wrong way again.

packages/legend-studio/src/stores/editor-state/element-editor-state/mapping/MappingEditorState.ts
```diff
@@ -899,19 +899,17 @@ export class MappingEditorState extends ElementEditorState {
       this.editorStore.pluginManager.getStudioPlugins(),
     );
     if (currentSource !== newSource) {
-      if (setImplementation instanceof PureInstanceSetImplementation) {
-        if (newSource instanceof Class || newSource === undefined) {
-          pureInstanceSetImpl_setSrcClass(setImplementation, newSource);
-        }
+      if (
```
akphi, line 902: I asked @gayathrir11 in #893 to follow this approach. Unfortunately, I missed the logic of the `else` block here. However, I believe the new approach would be cleaner. The logic here is that we try to check if the class mapping and the source type are compatible, if so, we can just update the source, without replacing the class mapping with a new **empty** class mapping that is compatible with the new source.

As such, we should make the logic here cleaner like this:
1. Create a boolean flag called `sourceUpdated`
2. If the compatibility test passes and the source is updated, we will set `sourceUpdated = true`
3. If we go through all the cases (including those coming from the plugins) and yet, source is still not updated, we will proceed to the second part of the logic where we create the replacement class mapping

In other words, I split the `else` block logic to make this flow more readable. I did this in #1285 and also included the compatibility check for `relational` case 😭. As for why I think the `instanceof` `if...else` should be structured like this, I think it's more readable to scope it by the type of the set implementation, here it means, all the logic for flat-data is already done here, all the logic for `pure-instance` is done here, plugins should not take care of neither.

PR #1289, opened 2022-07-12
Fix a bug where repeatedly adding empty string when filtering by/out on null

YannanGao-gs:
## Summary
Little improvement - Fix a bug where repeatedly adding empty string when filtering by/out on null

- [ ] Test(s) added

https://user-images.githubusercontent.com/73408381/178526023-47975073-8156-4e19-9a1a-c18ff7494fa8.mov

### DIFF
--- a/.changeset/plenty-terms-tickle.md
+++ b/.changeset/plenty-terms-tickle.md
@@ -1,0 +1,5 @@
+---
+'@finos/legend-query': patch
+---
+
+Fix a bug where repeatedly adding empty string when filtering by/out on null.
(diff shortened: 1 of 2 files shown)


MauricioUyaguari:
For this, are we sure the data is returning an empty string ? or is it return `null` ? 
If thats the case, the user probably means to add an isEmpty or IsNotEmpty filter

MauricioUyaguari (changes requested):
For this, are we sure the data is returning an empty string ? or is it return null ?
If thats the case, the user probably means to add an isEmpty or IsNotEmpty filter

akphi:
@YannanGao-gs could you check if this is still a bug? if not, can we close this PR, if it is, can you file an issue and close this PR? Thanks!

PR #1304, opened 2022-07-14
Provide a UI feedback when fetching mapped properties from Engine

irisyngao:
## Summary

Fixes https://github.com/finos/legend-studio/issues/1307
It usually takes a relatively long time to get output from mapping-algo API for a complex mapping, it's necessary to provide a UI feedback for users to let them know it's in progress.

- [ ] Test(s) added

https://user-images.githubusercontent.com/73408381/179024161-48bb0d0a-4a58-4cee-b998-5c73226c2a5b.mov

### DIFF
--- a/packages/legend-query/src/components/QueryBuilderExplorerPanel.tsx
+++ b/packages/legend-query/src/components/QueryBuilderExplorerPanel.tsx
@@ -44,6 +44,7 @@ import {
   PURE_ClassIcon,
   CheckIcon,
   SearchIcon,
+  PanelLoadingIndicator,
 } from '@finos/legend-art';
 import {
   type QueryBuilderExplorerTreeDragSource,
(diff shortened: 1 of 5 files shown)


akphi:
If that's the case, let's add `...` to the loading message and use `PanelLoadingIndicator` component to emphasize we're in the loading state 😄

packages/legend-query/src/components/QueryBuilderExplorerPanel.tsx
```diff
@@ -833,11 +833,13 @@ export const QueryBuilderExplorerPanel = observer(
     };
 
     useEffect(() => {
+      explorerState.initState.setMessage('Fetching mapped properties');
```
MauricioUyaguari, line 836: could we move this inside the method where we call `modelCoverage`. 
adding in this useEffect adds an unnecessary watch on initState here.

MauricioUyaguari (approved):

Issue #1265, opened 2022-07-06
Post Filter Minor Enhancements


MauricioUyaguari:
### Similar requests

- [X] I have searched and found no existing similar requests

### How are you using Studio?

Legend Studio

### What problems are you trying to solve?

Minor post filter enhancements to guide users on the feature. 

### Describe the solution you would like to see implemented

- [ ] add logic to toggle of show post filter. Disable if post filter condition on. 
- [ ] If `GraphFetch` selected show Unsupported if post filter is selected
- [ ] Disabled deleting column if used in post filter condition

### Contribution

- [X] I would like to work on this feature

PR #1314, opened 2022-07-18
post filter minor enhancements

YannanGao-gs:
## Summary

Fixes https://github.com/finos/legend-studio/issues/1265

- [x] Make the toggle behavior of post-filter consistent with parameter/function panel.

     e.g. If post filter condition is on, post-filter panel shouldn't disappear when toggling between projection panel and graph fetch panel.

- [x] Disable `show post filter` button if fetch mode is graph fetch
- [x] Warn users that post filters will lost when switching from projection to graph fetch

- [ ] Test(s) added

https://user-images.githubusercontent.com/73408381/179578857-6930bd4e-8392-44ff-b7f6-708a9001bfab.mov

### DIFF
--- a/.changeset/loud-rockets-attend.md
+++ b/.changeset/loud-rockets-attend.md
@@ -1,0 +1,5 @@
+---
+'@finos/legend-query': patch
+---
+
+Made the toggle behavior of post-filter consistent with parameter/function panel and added a warning popup when switching to graph fetch panel with post filter panel open [[#1265](https://github.com/finos/legend-studio/issues/1265) .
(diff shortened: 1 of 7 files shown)


MauricioUyaguari (approved):

beekemarie:
HI @YannanGao-gs & @MauricioUyaguari - would you mind adding the functionality to restrict removing of properties when selected in the post filter to the description? I think we should also add 2 more post filter enhancements: 1) not being able to hide post-filters when filter is applied and 2) not being able to remove aggregations when selected in post filter. Mind sharing the Github link when you opened it up?

Issue #1316, opened 2022-07-18
Bug: property editor panel is disabled when switching class editor panels


YannanGao-gs:
### Similar issues

- [X] I have searched and found no existing similar issues

### How are you using Studio?

Legend Studio

### Current and expected behavior

link https://github.com/finos/legend/issues/519

1. Select an attribute and click on the attribute details from the attribute tab: everything is fine.
2. Open a new tab attribute and do the same: everything is fine
3. Try to do the same on the previously opened tab attribute: doesn't work

### Steps to reproduce

1. Select an attribute and click on the attribute details from the attribute tab: everything is fine.
2. Open a new tab attribute and do the same: everything is fine
3. Try to do the same on the previously opened tab attribute: doesn't work

### Contribution

- [X] I would like to work on the fix for this issue

PR #1317, opened 2022-07-18
Fix a bug where the property editor panel is disabled when switching class editor panels

YannanGao-gs:
## Summary
Fixes https://github.com/finos/legend-studio/issues/1316
Fixes https://github.com/finos/legend/issues/519

- [ ] Test(s) added

https://user-images.githubusercontent.com/73408381/179628291-5854bf12-7ae4-4c6e-830c-7d720560172c.mov

### DIFF
--- a/.changeset/nine-numbers-eat.md
+++ b/.changeset/nine-numbers-eat.md
@@ -1,0 +1,5 @@
+---
+'@finos/legend-art': patch
+---
+
+Fix the bug where property editor panel is disabled when switching class editor panels ([[#1316](https://github.com/finos/legend-studio/issues/1316)]).
--- a/.changeset/nine-numbers-eat.md
+++ b/.changeset/nine-numbers-eat.md
@@ -2,4 +2,4 @@
 '@finos/legend-art': patch
 ---
 
-Fix the bug where property editor panel is disabled when switching class editor panels ([[#1316](https://github.com/finos/legend-studio/issues/1316)]).
+Fix a bug where the class property editor panel is hidden when switching between class editors ([#1316](https://github.com/finos/legend-studio/issues/1316)).
--- a/packages/legend-art/src/components/ResizablePanel.tsx
+++ b/packages/legend-art/src/components/ResizablePanel.tsx
@@ -93,6 +93,12 @@ export const getControlledResizablePanelProps = (
     options?.onStopResize?.(handleProps);
   },
   size: !minimizeCondition && options?.size ? options?.size : 0,
+  /**
+   * NOTE: An issue, the window collapses when reopening the parent complement,
+   * occurs by always setting flex to 0. Instead, When minimizeCondition is false,
+   * the value of flex should be handled programmatically based on size.
+   * Setting it to undefined will call an API at runtime to retrieve the actual value.
+   */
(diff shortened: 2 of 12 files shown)


packages/legend-studio/src/components/editor/edit-panel/uml-editor/ClassEditor.tsx
```diff
@@ -1248,7 +1248,6 @@ export const ClassFormEditor = observer(
       (tab: UML_EDITOR_TAB): (() => void) =>
       (): void => {
         editorState.setSelectedTab(tab);
-        editorState.setSelectedProperty(undefined);
```
akphi, line 1251: actually, I don't think we can remove this line here, since I suspect it would mean when we select a property and switch to another tab, say `Constraint`, the panel would still be there, which is weird. We need to hide it when switching between tabs of `class editor`

packages/legend-studio/src/components/editor/edit-panel/uml-editor/ClassEditor.tsx
```diff
@@ -1482,6 +1481,7 @@ export const ClassFormEditor = observer(
             flex={0}
             direction={-1}
             size={editorState.selectedProperty ? 250 : 0}
+            minSize={editorState.selectedProperty ? 250 : 0}
```
akphi, line 1484: We can't do this, `minSize` will make it not possible for people to collapse by `drag-and-drop` if I understood correctly. I think the approach is to look at the API of `react-reflex` a bit here.

Also, this code pattern is being used in several places, perhaps, it's best to (whatever change you do here) apply the same change to those places in the end

akphi (changes requested):

.changeset/nine-numbers-eat.md
```diff
@@ -0,0 +1,5 @@
+---
+'@finos/legend-art': patch
+---
+
+Fix the bug where property editor panel is disabled when switching class editor panels ([[#1316](https://github.com/finos/legend-studio/issues/1316)]).
```
akphi, line 5:
```suggestion
Fix a bug where the class property editor panel is hidden when switching between class editors ([#1316](https://github.com/finos/legend-studio/issues/1316)).
```

packages/legend-art/src/components/ResizablePanel.tsx
```diff
@@ -91,6 +92,8 @@ export const getControlledResizablePanelProps = (
     }
     options?.onStopResize?.(handleProps);
   },
+  size: !minimizeCondition && options?.size ? options?.size : 0,
```
akphi, line 95: Can you document about this?

akphi (changes requested):

akphi (dismissed):

packages/legend-art/src/components/ResizablePanel.tsx
```diff
@@ -91,6 +92,14 @@ export const getControlledResizablePanelProps = (
...
```
akphi, line 95:
```suggestion
  size: !minimizeCondition && options?.size ? options.size : 0,
```

akphi (approved):

PR #1318, opened 2022-07-19
add support for `MultiExecutionTestResult`

MauricioUyaguari:
## Summary

Handle MultiExecutionTestResult in Service Test Editor

- [ ] Test(s) added


packages/legend-graph/src/models/protocols/pure/v1/V1_PureGraphManager.ts
```diff
@@ -1734,6 +1733,9 @@ export class V1_PureGraphManager extends AbstractPureGraphManager {
           ),
       );
       const result = results[0];
+      if (result instanceof TestError) {
+        throw new IllegalStateError(result.error);
+      }
       assertType(result, TestFailed);
```
akphi, line 1739: Sorry could you explain why we need this assertion here?

akphi (changes requested):
Nothing major, why didn't we re-enable the test for global test runner?

packages/legend-graph/src/models/protocols/pure/v1/V1_PureGraphManager.ts
MauricioUyaguari: the way i generate expected result is run the test expecting to fail. 
Then use the diff to get expected result

akphi: Did we document that somewhere, else could we add that here?

akphi (approved):

PR #1327, opened 2022-07-22
Fix creating new milestoned query from dataspace

### LINKED ISSUE #1315: Bug: Query Builder opens misconfigured milestone query when coming from dataspace diagram
### Similar issues

### How are you using Studio?

Legend Query

When opening a milestone class from the dataspace diagram, the usual logic to generate milestone query is not evoked causing a query that does not execute nor gives the user a chance to fix it. 

### Steps to reproduce

- Open dataspace with milestone class in diagram
- double click on milestone class in diagram
- query builder opens the milestone class and does not have a date generated. When executing we get no date error
- workaround right now is to change classes and come back to the original class which fixes the issue

- also make sure opening to query builder on milestone class with url opens correctly

### Model data

_No response_

### Environment

_No response_

_No response_

### Contribution

gayathrir11:
## Summary

closes #1315 

- [ ] Test(s) added

### DIFF
--- a/packages/legend-application-query/src/stores/QueryEditorStore.ts
+++ b/packages/legend-application-query/src/stores/QueryEditorStore.ts
@@ -443,22 +443,21 @@ export class CreateQueryEditorStore extends QueryEditorStore {
       ),
     );
     if (this.classPath) {
-      this.queryBuilderState.querySetupState._class =
-        this.queryBuilderState.graphManagerState.graph.getClass(this.classPath);
+      this.queryBuilderState.changeClass(
+        this.queryBuilderState.graphManagerState.graph.getClass(this.classPath),
+      );
       this.queryBuilderState.querySetupState.setClassIsReadOnly(true);
     } else {
-      this.queryBuilderState.querySetupState._class = getNullableFirstElement(
-        this.queryBuilderState.querySetupState.mapping
-          ? getAllClassMappings(
-              this.queryBuilderState.querySetupState.mapping,
-            ).map((classMapping) => classMapping.class.value)
-          : [],
+      this.queryBuilderState.changeClass(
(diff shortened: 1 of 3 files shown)


YannanGao-gs:
@gayathrir11  Looks good to me :)

packages/legend-application-query/src/stores/QueryEditorStore.ts
```diff
@@ -455,10 +456,6 @@ export class CreateQueryEditorStore extends QueryEditorStore {
           : [],
       );
     }
-
-    // initialize query builder state after setting up
```
MauricioUyaguari, line 459: why are we removing these lines ?

MauricioUyaguari (changes requested):

packages/legend-application-query/src/stores/QueryEditorStore.ts
```diff
@@ -443,8 +443,9 @@ export class CreateQueryEditorStore extends QueryEditorStore {
...
     );
     if (this.classPath) {
-      this.queryBuilderState.querySetupState._class =
-        this.queryBuilderState.graphManagerState.graph.getClass(this.classPath);
+      this.queryBuilderState.changeClass(
+        this.queryBuilderState.graphManagerState.graph.getClass(this.classPath),
+      );
       this.queryBuilderState.querySetupState.setClassIsReadOnly(true);
     } else {
       this.queryBuilderState.querySetupState._class = getNullableFirstElement(
```
MauricioUyaguari, line 451: what happens if the class selected here is milestoned. Should we also use the `changeClass` method here

MauricioUyaguari (changes requested):

MauricioUyaguari (approved):

PR #1351, opened 2022-07-29
fix regression: do not highlight property icon when it is unmapped

gayathrir11:
## Summary

Grey out property icon when it is unmapped. It's a bug introduced in #1338 

- [ ] Test(s) added

https://user-images.githubusercontent.com/87973126/182217626-e4238a0e-c660-44c6-9edc-a9845afd1270.mp4

### DIFF
--- a/packages/legend-application-query/style/_query-builder-explorer.scss
+++ b/packages/legend-application-query/style/_query-builder-explorer.scss
@@ -335,12 +335,14 @@
 
   &__expand-icon svg {
     font-size: 1rem;
-    color: var(--color-light-grey-200) !important;
+    color: var(--color-light-grey-200);
   }
 
   &__node__container:hover {
     background: var(--color-dark-blue-shade-100);
+  }
 
+  &__node__container:hover &__node__action {
     svg {
       color: var(--color-light-grey-400);
     }
@@ -459,7 +461,7 @@
   }
 
   &__icon {
-    color: var(--color-primitive) !important;
+    color: var(--color-primitive);
 
     &__boolean,
     &__number,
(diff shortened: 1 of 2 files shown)


packages/legend-application-query/style/_query-builder-explorer.scss
```diff
@@ -474,7 +474,7 @@
   &__node__container--unmapped &__type-icon &__icon,
   &__node__container--unmapped &__node__label__derived-property,
   &__node__container--unmapped &__node__label__multiple {
-    color: var(--color-dark-grey-350);
+    color: var(--color-dark-grey-350) !important;
```
akphi, line 477: generally, we try to avoid `!important` since later, it's hard to override and we would have to make everything __more__ `!important`. So, let's try the followings:
1. re-arrange classes so the rule for `unmapped` goes later (I think even the order of CSS classes matter so consider reordering in the component is also another way)
2. have a more specific rule, so you can say if the node is `selected` and `unmapped`, show text as grey

akphi (approved):

akphi (changes requested):
sorry, mis-clicked approved

packages/legend-application-query/style/_query-builder-explorer.scss
akphi: actually, if this is introduced by #1338 where you set `color: var(--color-primitive) !important;`, we should revert that change and find a different strategy.

MauricioUyaguari (approved):

MauricioUyaguari (dismissed):
clicked approve by accident as well :)

akphi (approved):

PR #1358, opened 2022-07-30
support query execution cancellation

### LINKED ISSUE #1357: Feature request: allow cancelling query execution
### Similar requests

### How are you using Studio?

Legend Query

Currently, users can't cancel a query whose execution time is long and are blocked to re-run a new query.
We should support execution cancellation.
Link to https://github.com/finos/legend-studio/issues/937

Currently, as a quick workaround, I chose to ignore the result of previous request instead of cutting connection from server side.

_No response_

_No response_

### Contribution

YannanGao-gs:
## Summary

Fixes #1357

- [ ] Test(s) added

https://user-images.githubusercontent.com/73408381/182652267-58296dcb-9e93-4cfd-8fb1-1a4fd8d4779a.mov

### DIFF
--- a/packages/legend-application-query/src/components/QueryBuilderResultPanel.tsx
+++ b/packages/legend-application-query/src/components/QueryBuilderResultPanel.tsx
@@ -430,7 +430,7 @@ export const QueryBuilderResultPanel = observer(
       }
     };
     const cancelQuery = (): void => {
-      resultState.isRunningQuery = false;
+      resultState.setIsRunningQuery(false);
       queryBuilderState.resultState.setQueryRunPromise(undefined);
     };
     const generatePlan = applicationStore.guardUnhandledError(() =>
--- a/packages/legend-application-query/src/components/QueryBuilderResultPanel.tsx
+++ b/packages/legend-application-query/src/components/QueryBuilderResultPanel.tsx
@@ -480,7 +480,7 @@ export const QueryBuilderResultPanel = observer(
                 onClick={cancelQuery}
                 tabIndex={-1}
               >
-                <div className="query-builder__result__stop-btn__label">
+                <div className="btn--dark btn--caution query-builder__result__stop-btn__label">
                   <PauseCircleIcon className="query-builder__result__stop-btn__label__icon" />
                   <div className="query-builder__result__stop-btn__label__title">
                     Stop
--- a/packages/legend-application-query/src/components/QueryBuilderResultPanel.tsx
+++ b/packages/legend-application-query/src/components/QueryBuilderResultPanel.tsx
@@ -417,11 +417,16 @@ export const QueryBuilderResultPanel = observer(
     const runQuery = (): void => {
       if (queryParametersState.parameterStates.length) {
         queryParametersState.parameterValuesEditorState.open(
-          (): Promise<void> => resultState.runQuery(),
+          (): Promise<void> =>
+            flowResult(resultState.runQuery()).catch(
+              applicationStore.alertUnhandledError,
+            ),
           PARAMETER_SUBMIT_ACTION.EXECUTE,
         );
       } else {
-        resultState.runQuery().catch(applicationStore.alertUnhandledError);
+        flowResult(resultState.runQuery()).catch(
+          applicationStore.alertUnhandledError,
+        );
       }
     };
     const cancelQuery = (): void => {
--- a/packages/legend-application-query/src/components/QueryBuilderResultPanel.tsx
+++ b/packages/legend-application-query/src/components/QueryBuilderResultPanel.tsx
@@ -414,14 +414,14 @@ export const QueryBuilderResultPanel = observer(
         ],
       });
     };
-    const runQuery = async (): Promise<void> => {
+    const runQuery = (): void => {
       if (queryParametersState.parameterStates.length) {
         queryParametersState.parameterValuesEditorState.open(
           (): Promise<void> => resultState.runQuery(),
           PARAMETER_SUBMIT_ACTION.EXECUTE,
         );
       } else {
-        await resultState.runQuery();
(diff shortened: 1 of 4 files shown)


packages/legend-application-query/src/stores/QueryBuilderResultState.ts
```diff
@@ -172,7 +173,12 @@ export class QueryBuilderResultState {
     }
   }
 
-  *execute(): GeneratorFn<void> {
+  async execute(
```
akphi, line 176: Generally, I agree with the approach, however, I think this logic should be put in the state instead. My suggestion is to create a new field in the reuslt state:

```ts
queryRunPromise: Promise<void> | undefined = undefined;
```

Also, let's rename this method to `runQuery` and the boolean flag to `isRunningQuery` instead of `isExecutingQuery`

packages/legend-application-query/src/components/QueryBuilderResultPanel.tsx
```diff
@@ -474,23 +489,18 @@ export const QueryBuilderResultPanel = observer(
...
-                resultState.isExecutingQuery || resultState.isGeneratingPlan
-              }
+              onClick={isRunningQuery ? cancelQuery : runQuery}
               tabIndex={-1}
             >
               <div className="query-builder__result__execute-btn__label">
                 <PlayIcon className="query-builder__result__execute-btn__label__icon" />
                 <div className="query-builder__result__execute-btn__label__title">
-                  Run Query
+                  {isRunningQuery ? 'Cancel' : 'Run Query'}
```
akphi, line 498: --> `Stop` and also we might want to change the styling here, certain operation in the dropdown wouldn't make sense anymore`. As such, I think we could consider replacing this button by another one while executing. What do you think @MauricioUyaguari @YannanGao-gs?

packages/legend-application-query/src/components/QueryBuilderResultPanel.tsx
```diff
@@ -413,19 +425,22 @@ export const QueryBuilderResultPanel = observer(
...
       } else {
-        flowResult(resultState.execute()).catch(
-          applicationStore.alertUnhandledError,
-        );
+        await resultState.execute(abortQueryExecution, setIsRunningQuery);
+      }
+    };
+    const cancelQuery = (): void => {
+      setIsRunningQuery(false);
+      if (abortQueryExecution.current) {
```
akphi, line 442: instead, let's make this reset the `runQueryPromise` instead

packages/legend-application-query/src/components/QueryBuilderResultPanel.tsx
```diff
@@ -78,6 +78,16 @@ import {
 } from '../stores/postFilterOperators/QueryBuilderPostFilterOperator_In.js';
 import type { QueryBuilderPostFilterOperator } from '../stores/QueryBuilderPostFilterOperator.js';
 
+export class AbortQueryExecutionHelper {
```
akphi, line 81: we really don't need this, also the dependency you introduce is weird since now the state/store depends on something from `components`, let's remove this

akphi (changes requested):

packages/legend-application-query/style/_query-builder.scss
```diff
@@ -336,6 +336,41 @@
     @include flexVCenter;
   }
 
+  &__stop-btn {
```
akphi, line 339: let's just use `btn--dark btn--caution`

packages/legend-application-query/src/components/QueryBuilderResultPanel.tsx
```diff
@@ -373,6 +374,7 @@ export const QueryBuilderResultPanel = observer(
     const resultState = queryBuilderState.resultState;
     const queryParametersState = queryBuilderState.queryParametersState;
     const executionResult = resultState.executionResult;
+    const [isRunningQuery, setIsRunningQuery] = useState(false);
```
akphi, line 377: we don't need the flag `isRunningQuery` at all. Please help me rename `QueryBuilderResultState.isExecutingQuery` to `isRunningQuery` and use that flag instead.

akphi (changes requested):

packages/legend-application-query/style/_query-builder.scss
YannanGao-gs: Screen Shot 2022-08-03 at 9 31 33 AM

packages/legend-application-query/src/stores/QueryBuilderResultState.ts
```diff
@@ -172,9 +181,9 @@ export class QueryBuilderResultState {
     }
   }
 
-  *execute(): GeneratorFn<void> {
+  async runQuery(): Promise<void> {
```
akphi, line 184: Can we leave this as `generator` here. I'm worried since we're relying on `makeAutoObservable` to pick this up as `asyncAction`, but we should properly use `flow` here instead

akphi (changes requested):

packages/legend-application-query/src/components/QueryBuilderResultPanel.tsx
```diff
@@ -413,21 +414,25 @@ export const QueryBuilderResultPanel = observer(
...
         );
       } else {
-        flowResult(resultState.execute()).catch(
+        flowResult(resultState.runQuery()).catch(
           applicationStore.alertUnhandledError,
         );
       }
     };
+    const cancelQuery = (): void => {
+      resultState.isRunningQuery = false;
```
akphi, line 433: you would need to call the setter here else `mobx` will complain.

akphi (changes requested):

akphi (approved):

PR #1359, opened 2022-08-01
Block certain actions when query is not valid

### LINKED ISSUE #1345: Feature request: Block certain actions when query is not valid
### Similar requests

### How are you using Studio?

Legend Studio

Block certain actions when query is no valid

Avoid users to run options in result panel when
- [x] They have duplicated columns in projection panel
- [x] There are no columns in the projection panel
- [x] Do not disable run options when `graphFetch` is selected

We already handle the case where user tries to create duplicate parameters in form mode. We diable the `create` button in this scenario. We don't want to handle the case where user creates duplicate parameters in text mode.

_No response_

_No response_

### Contribution

gayathrir11:
## Summary

closes #1345 

- [ ] Test(s) added

https://user-images.githubusercontent.com/87973126/182402667-672d26ed-3c85-4172-a39d-8d3732b4d09f.mp4

### DIFF
--- a/packages/legend-application-query/src/stores/QueryBuilderProjectionState.ts
+++ b/packages/legend-application-query/src/stores/QueryBuilderProjectionState.ts
@@ -793,4 +793,19 @@ export class QueryBuilderProjectionState {
       );
     }
   }
+
+  isValidProjectionState(): boolean {
+    if (this.queryBuilderState.fetchStructureState.isProjectionMode()) {
+      // duplicate columns check
+      const hasDuplicatedProjectionColumns = this.columns.some(
+        (column) =>
+          this.columns.filter((c) => c.columnName === column.columnName)
+            .length > 1,
+      );
+      // no columns check
+      const hasNoProjectionColumns = this.columns.length === 0;
+      return !hasDuplicatedProjectionColumns && !hasNoProjectionColumns;
+    }
+    return true;
+  }
 }
--- a/.changeset/fast-spiders-flow.md
+++ b/.changeset/fast-spiders-flow.md
@@ -1,0 +1,5 @@
+---
+"@finos/legend-application-query": patch
+---
+
+Block certain actions when query is not valid([#1345](https://github.com/finos/legend-studio/issues/1345)).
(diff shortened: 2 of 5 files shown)


YannanGao-gs:
@gayathrir11  What's the current behavior if users want to set the filter value to an empty string?

packages/legend-application-query/src/components/QueryBuilderFilterPanel.tsx
```diff
@@ -214,6 +217,9 @@ const QueryBuilderFilterConditionEditor = observer(
         node.condition.operator.getDefaultFilterConditionValue(node.condition),
       );
     };
+    const isEmptyNode = (): boolean =>
```
MauricioUyaguari, line 220: Empty string values are valid inputs.

packages/legend-application-query/src/stores/QueryBuilderProjectionState.ts
```diff
@@ -793,4 +793,17 @@ export class QueryBuilderProjectionState {
       );
     }
   }
+
+  isValidProjectionState(): boolean {
+    const hasDuplicatedProjectionColumns = this.columns.some(
```
MauricioUyaguari, line 798: lets make this a bit more readable. 
We can do something like 
```
if(this.queryBuilderState.fetchStructureState.isProjectionMode)
{
   // duplicate cols check
  // no cols check 
}
return true;
````

**Review by MauricioUyaguari — CHANGES_REQUESTED** [review]

File: packages/legend-application-query/src/stores/QueryBuilderProjectionState.ts [hunk 3]
```diff
@@ -793,4 +793,19 @@ export class QueryBuilderProjectionState {
       );
     }
   }
+
+  isValidProjectionState(): boolean {
+    if (this.queryBuilderState.fetchStructureState.isProjectionMode()) {
+      // duplicate columns check
+      const hasDuplicatedProjectionColumns = this.columns.some(
+        (column) =>
+          this.columns.filter((c) => c.columnName === column.columnName)
```
YannanGao-gs, line 802: the time complexity of duplication check is O(n^2) 
too high

packages/legend-application-query/src/stores/QueryBuilderFilterState.ts
```diff
@@ -415,6 +415,7 @@ export class QueryBuilderFilterState
     makeAutoObservable(this, {
       queryBuilderState: false,
       isValidMove: false,
+      nodes: observable,
```
MauricioUyaguari, line 418: why do we need this?

MauricioUyaguari (changes requested):
why do we need this?

MauricioUyaguari (approved):

PR #1367, opened 2022-08-02
Highlight already used properties in explorer tree of query builder

### LINKED ISSUE #1344: Feature request: Highlight already used properties in explorer tree of query builder
### Similar requests

### How are you using Studio?

Legend Studio

Highlight already used properties in explorer tree of query builder

- [x] Highlight already used options in explorer tree when used in `projection` panel or `graphFetch` panel
- [x] Add a config option to explorer tree whether to grey out already used properties or not
- [x]  Set the above mentioned config option as default

_No response_

_No response_

### Contribution

gayathrir11:
## Summary

Closes #1344 

- [ ] Test(s) added

https://user-images.githubusercontent.com/87973126/186336131-b80ecac4-930a-4f95-a3e0-e690565b57ab.mp4

***Light Theme***
Capture

***Graph fetch***
hh

### DIFF
--- a/packages/legend-application-query/src/stores/__tests__/QueryBuilder_ExplorerState.test.ts
+++ b/packages/legend-application-query/src/stores/__tests__/QueryBuilder_ExplorerState.test.ts
@@ -21,16 +21,22 @@ import TEST_DATA__COVIDDataSimpleModel from './TEST_DATA__QueryBuilder_Model_COV
 import TEST_DATA__AssociationMappingModel from '../../stores/__tests__/TEST_DATA__QueryBuilder_Model_AssociationMappingModel.json';
 import TEST_DATA__M2MAutoMapped from '../../stores/__tests__/TEST_DATA__QueryBuilder_Model_M2MAutoMapped.json';
 import TEST_DATA__RelationalInline from '../../stores/__tests__/TEST_DATA__QueryBuilder_Model_RelationalInline.json';
+import TEST_DATA_SimpleSubtypeModel from '../../stores/__tests__/TEST_DATA__QueryBuilder_Model_SimpleSubtype.json';
+import TEST_DATA_ComplexSubtypeModel from '../../stores/__tests__/TEST_DATA__QueryBuilder_Model_ComplexSubType.json';
 import {
   type TEMPORARY__JestMatcher,
   type PlainObject,
   integrationTest,
+  guaranteeNonNullable,
+  guaranteeType,
 } from '@finos/legend-shared';
 import type { Entity } from '@finos/legend-storage';
 import {
   Class,
+  create_RawLambda,
   getAllClassDerivedProperties,
   getAllClassProperties,
+  stub_RawLambda,
   type AbstractProperty,
   type GraphManagerState,
   type Mapping,
@@ -40,10 +46,15 @@ import {
   type QueryBuilderPropertyMappingData,
   getRootMappingData,
   generatePropertyNodeMappingData,
+  QueryBuilderExplorerTreeRootNodeData,
+  QueryBuilderExplorerTreeSubTypeNodeData,
 } from '../explorer/QueryBuilderExplorerState.js';
 import { LegendQueryPluginManager } from '../../application/LegendQueryPluginManager.js';
 import { QueryBuilder_GraphManagerPreset } from '../../graphManager/QueryBuilder_GraphManagerPreset.js';
-import { TEST__provideMockedQueryEditorStore } from '../../components/QueryEditorComponentTestUtils.js';
+import {
+  TEST__provideMockedQueryEditorStore,
+  TEST__setUpQueryEditor,
+} from '../../components/QueryEditorComponentTestUtils.js';
 import {
   TEST_DATA__MappingData__ComplexM2MModel,
   TEST_DATA__MappingData__AssociationMapping,
@@ -59,7 +70,21 @@ import {
   TEST_DATA__ModelCoverageAnalysisResult_COVIDDataSimple,
   TEST_DATA__ModelCoverageAnalysisResult_RelationalInline,
   TEST_DATA__ModelCoverageAnalysisResult_SimpleRelationalInheritance,
+  TEST_DATA__ModelCoverageAnalysisResult_HighlightProperties,
+  TEST_DATA__ModelCoverageAnalysisResult_SimpleSubtype,
+  TEST_DATA__ModelCoverageAnalysisResult_ComplexSubtype,
 } from './TEST_DATA__ModelCoverageAnalysisResult.js';
+import TEST_DATA__QueryBuilder_Model_HighlightProperties from './TEST_DATA__QueryBuilder_Model_HiglightProperties.json';
+import {
+  TEST_DATA_simpleGraphFetchWithSubType,
+  TEST_DATA__simpleGraphFetch,
+  TEST_DATA__simpleProjection,
+} from './TEST_DATA__QueryBuilder_HighlightProperties.js';
+import { waitFor, getByText, getAllByText } from '@testing-library/react';
+import { act } from 'react-dom/test-utils';
+import { QUERY_BUILDER_TEST_ID } from '../../components/QueryBuilder_TestID.js';
+import { TEST_DATA__simpleProjectionWithSubtypeFromSubtypeModel } from './TEST_DATA__QueryBuilder_Generic.js';
+import { isNodeAlreadyUsed } from '../../components/explorer/QueryBuilderExplorerPanel.js';
 
 interface NodePropertyMappingData {
   property: AbstractProperty;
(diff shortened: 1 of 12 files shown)


packages/legend-application-query/src/components/QueryBuilderExplorerPanel.tsx
```diff
@@ -113,6 +116,24 @@ const checkForDeprecatedNode = (
   return false;
 };
 
+const checkForAlreadyUsedNode = (
```
YannanGao-gs, line 119: i prefer to update this name.
e.g isNodeAlreadyUsed?

packages/legend-application-query/src/components/QueryBuilderExplorerPanel.tsx
```diff
@@ -113,6 +116,24 @@ const checkForDeprecatedNode = (
...
+  node: QueryBuilderExplorerTreeNodeData,
+  projectionState: QueryBuilderProjectionState,
+): boolean => {
+  if (node.id === '@dummy_rootNode' && projectionState.columns.length) {
+    return true;
+  } else if (
+    projectionState.columns.some(
+      (column) =>
+        column instanceof QueryBuilderSimpleProjectionColumnState &&
```
YannanGao-gs, line 128: what should we do with QueryBuilderDerivationProjectionColumnState?

packages/legend-application-query/src/components/QueryBuilderExplorerPanel.tsx
```diff
@@ -113,6 +116,24 @@ const checkForDeprecatedNode = (
...
+        column.propertyExpressionState.path.startsWith(node.id),
```
YannanGao-gs, line 129: don't think it will work for subtype.

packages/legend-application-query/src/components/QueryBuilderExplorerPanel.tsx
gayathrir11: It's hard to analysis for that so we decided not to do

packages/legend-application-query/src/components/QueryBuilderExplorerPanel.tsx
```diff
@@ -113,6 +116,46 @@ const checkForDeprecatedNode = (
...
+    return true;
+  }
+  // For subType
+  else if (node.id.startsWith(TYPE_CAST_TOKEN)) {
+    let propName = '';
+    node.id
+      .split('.')
+      .slice(1)
+      .forEach((p) => {
+        propName = `${propName + p}.`;
```
YannanGao-gs, line 142: Another approach is to generate the property chain name for this node e.g. (@Firm)/Employee Size

1. buildPropertyExpressionFromExplorerTreeNodeData
2. getPropertyChainName
3. compare column's title

YannanGao-gs: @akphi @MauricioUyaguari Which one do you guys prefer?

packages/legend-application-query/src/components/QueryBuilderExplorerPanel.tsx
```diff
@@ -856,6 +907,10 @@ export const QueryBuilderExplorerPanel = observer(
       explorerState.setHumanizePropertyName(
         !explorerState.humanizePropertyName,
       );
+    const toggleHighlightUsedProperties = (): void =>
```
akphi, line 910: Nice!

packages/legend-application-query/src/stores/QueryBuilderExplorerState.ts
```diff
@@ -517,6 +518,7 @@ export class QueryBuilderExplorerState {
       refreshTreeData: action,
```
MauricioUyaguari, line 518: dont we need an observable for `highlightUsedProperties`

MauricioUyaguari (changes requested):

packages/legend-application-query/src/stores/QueryBuilderLambdaProcessor.ts
```diff
@@ -1186,6 +1186,7 @@ export class QueryBuilderLambdaProcessor
         valueSpecification,
         false,
       );
+      columnState.setNodeIdsForProjectionColumn();
```
akphi, line 1189: So the approach you're adopting is to find all the places where we create `projection column` then decorate the state with an `ID`, which you can use when rendering the explorer tree to check if the property is already used or not.

It's a sound approach. However, I think we should modify this slightly. With the current way, we decorate the projection column actively, which is fragile since we could miss spots. To me, for this feature, it's more an analytics than an attribute. As such, let's create a getter/computation instead.

For example, have a computation to get used properties in `QueryBuilderProjectionState`

```ts
get usedProperties(): AbstractProperty[]
```

or something like that. Then have explorer use this computation/analytics to render the tree.

packages/legend-application-query/src/components/QueryBuilderExplorerPanel.tsx
```diff
@@ -113,6 +116,48 @@ const checkForDeprecatedNode = (
   return false;
 };
 
+const isNodeAlreadyUsed = (
+  node: QueryBuilderExplorerTreeNodeData,
+  projectionState: QueryBuilderProjectionState,
+): boolean => {
+  if (
+    node.id === '@dummy_rootNode' &&
```
akphi, line 124: either we publish the root node ID as a constant and use it here, or check for root node using some other means, i.e. check that it doesn't have a parentNode?

packages/legend-application-query/src/stores/QueryBuilderLambdaProcessor.ts
akphi: or maybe we can think of getting the ID with something like `get usedPropertyIds(): string[]`.

Regardless, I want the relationship between explorer (Exp) and fetch-structure (FS) states go like this:

```
Exp: Hey, can you tell me what you have used?
FS: Sure, this is the list ...
Exp: Thanks, let me try to understand it
```

At least, that's what I think makes sense for now. _I'm worried a bit that this kind of check could tangle up and make it harder for us later to do generic query builder, but maybe I'm just overthinking here_

akphi (changes requested):

packages/legend-application-query/src/stores/QueryBuilderProjectionState.ts
```diff
@@ -791,4 +795,44 @@ export class QueryBuilderProjectionState {
     }
     return undefined;
   }
+
+  get usedPropertyNodeIds(): string[] {
```
akphi, line 799: This makes sense now. However, im worried about the fact that we maintain in parallel 2 pieces of logic to compute node Id. Is there a way we share the utility between this and explorer state so if one changes, they would not go out of sync?

akphi (changes requested):

packages/legend-application-query/src/stores/QueryBuilderProjectionState.ts
akphi: To clarify further, what I meant is more like we come up with a good way to ensure the way we generate node ID here is the same way we generate the node ID when we do `new QueryBuilderExplorerTreeSubTypeNodeData(...)`. So for these, let's create some shared utility methods in `ExplorerState` to generate part of the ID.

Also, let's create a test for this, so we know these 2, `get usedPropertyNodeIds()` and `QueryBuilderExplorerState` always generate IDs the same way.

packages/legend-application-query/src/stores/__tests__/QueryBuilder_ExplorerState.test.ts
```diff
@@ -284,3 +303,130 @@ describe(integrationTest('Build property mapping data'), () => {
     },
   );
 });
+
+test(
+  integrationTest('Highlight already used properties in explorer tree'),
```
akphi, line 308: --> `Highlight already used properties in projection`

packages/legend-application-query/src/stores/__tests__/QueryBuilder_ExplorerState.test.ts
```diff
@@ -284,3 +303,130 @@ describe(integrationTest('Build property mapping data'), () => {
...
+      ).filter((node) =>
+        isNodeAlreadyUsed(node, queryBuilderState.fetchStructureState),
+      ).length,
+    ).toBe(2);
+  },
+);
+
+test(
+  integrationTest(
+    'Highlight already used properties in explorer tree for subType',
```
akphi, line 359: --> `Highlight already used subtype in projection`

akphi: Let's add a test for graph-fetch

akphi (changes requested):
Looking good! Let's rebase and add a test for graph-fetch. Thanks!

akphi (approved):

Issue #1320, opened 2022-07-19
Productionize Query Through Service


MauricioUyaguari:
### Overview

Users should have a workflow to promote a query from Legend Query to a service.  Right now the only option is for users to open associated project and import query

### Implementation plan

Provide a `promote to service` button in legend query (for create , existing, dataspace query ? )
When clicked will show the following inputs will show in the model 

- [ ] `SDLCInstance`: config added to legend-query to load all sdlc instances
  - [ ] Ensure we authorize with corresponding sdlc instance
- [ ] `Project` list of projects from sdlc instance. default to project listed in depot from gav coordinates? 
   - [ ] `Improvement` Have the create project option here ? 
- [ ] `ProjectSetupCompleted` flag to indicate whether or not project has the corresponding metadata and dependencies to support query. If unchecked we will add this meatdata to the project and dependency to said project.
- [ ] `Workspace` -> default to groupWorkspace 
- [ ] `Service Metadata`
  - [ ] path
  - [ ] url (can generate from lambda parameters)
  - [ ] documentation ? 
  - [ ] owners ?

### Productionize Workflow
 - ✔️  `ProjectSetupCompleted`
    - `validation` verify project setup is complete. check dependencies ? check entities? 
    - fetch project entities
    - create service from form -> serialize to the entity
    - build graph from project entities + service (or use engine to compile ? )
    - if compile -> continue
    - if no compile -> show error to user
    - create group workspace
    - commit with SINGLE service entity
    - open window to corresponding group workspace service 
- 🚫  `ProjectSetupCompleted`
    - fetch project entities
    - create service from form -> serialize to the entity
    - project entities + pmcd -> entities + service entities -> compile/build graph
    - if compile -> continue
    - if no compile -> show error to user
    - create group workspace
    - commit with (pmcd -> entities) + service entity (create entity change)
    - open window to corresponding group workspace service

PR #1375, opened 2022-08-03
Promote Query To Service

MauricioUyaguari:
## Summary

- Closes https://github.com/finos/legend-studio/issues/1320
- Blocked by https://github.com/finos/legend-studio/issues/1403

- [ ] Test(s) added

https://user-images.githubusercontent.com/29690824/182667506-ec1ab267-922c-4a79-87f3-70582e170f4e.mp4

### DIFF
--- a/packages/legend-application-query/src/stores/QueryPromoteToServiceState.ts
+++ b/packages/legend-application-query/src/stores/QueryPromoteToServiceState.ts
@@ -1,0 +1,623 @@
+/**
+ * Copyright (c) 2020-present, Goldman Sachs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {
+  ActionAlertActionType,
+  ActionAlertType,
+} from '@finos/legend-application';
+import {
+  getOrCreateGraphPackage,
+  PackageableElementExplicitReference,
+  PureSingleExecution,
+  resolvePackagePathAndElementName,
+  Service,
+} from '@finos/legend-graph';
+import {
+  type ProjectData,
+  generateGAVCoordinates,
+  LATEST_VERSION_ALIAS,
+  ProjectDependencyCoordinates,
+  ProjectVersionEntities,
+  SNAPSHOT_VERSION_ALIAS,
+} from '@finos/legend-server-depot';
+import {
+  EntityChange,
+  EntityChangeType,
+  ProjectConfiguration,
+  ProjectDependency,
+  UpdateProjectConfigurationCommand,
+  Workspace,
+  WorkspaceType,
+  Project,
+} from '@finos/legend-server-sdlc';
+import {
+  type GeneratorFn,
+  type PlainObject,
+  assertNonEmptyString,
+  assertTrue,
+  IllegalStateError,
+  generateEnumerableNameFromToken,
+  ActionState,
+  assertErrorThrown,
+  guaranteeNonNullable,
+  uuid,
+  prettyCONSTName,
+  UnsupportedOperationError,
+} from '@finos/legend-shared';
+import type { Entity } from '@finos/legend-storage';
+import { action, flow, flowResult, makeObservable, observable } from 'mobx';
+import type { QueryEditorStore } from './QueryEditorStore.js';
+
+enum PROMOTING_SERVICE_PHASE {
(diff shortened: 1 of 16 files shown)


packages/legend-application-query/src/stores/QueryPromoteToServiceState.ts
```diff
@@ -0,0 +1,586 @@
...
+          const coordinates =
+            this.transformProjectDependencyToProjectCoordinates([
+              dependencyToAdd,
+            ]);
+          dependencyCoordinates = [...dependencyCoordinates, ...coordinates];
+          serviceCommitCommand.configuration = projectConfiguration;
+          serviceCommitCommand.dependenciesToAdd = [dependencyToAdd];
+        }
+      } else {
+        throw new IllegalStateError('TODOFIX');
```
MauricioUyaguari, line 472: hanlde this

packages/legend-application-query/src/application/LegendQueryApplicationConfig.ts
```diff
@@ -16,14 +16,26 @@
...
   type LegendApplicationConfigurationData,
 } from '@finos/legend-application';
 
+type SDLCConfig = {
+  default: string;
+  instances: {
+    url: string;
+    key: string;
+    label?: string;
+    studioUrl: string;
```
akphi, line 35: why is there a `studioUrl` in `SDLC` config? should we rename this to `ServiceProdutionizationConfig` or this field to `correspondingStudioInstanceUrl`. Also, we could remove `studioUrl` from `query config` and use this one.

Also, let's remove `default` field and do it like this since we have been it for other places (i.e. taxonomy tree config)

```
  default: string;
  instances: {
    url: string;
    key: string;
    default?: boolean;
    label?: string;
    correspondingStudioInstanceUrl: string;
  }
```

packages/legend-graph/src/graphManager/AbstractPureGraphManager.ts
```diff
@@ -233,7 +233,7 @@ export abstract class AbstractPureGraphManager {
   ): Promise<Map<string, string>>;
 
   // ------------------------------------------- Compile -------------------------------------------
-
+  abstract compileEntities(entities: Entity[]): Promise<void>;
```
akphi, line 236: let's add a new line before this block

packages/legend-server-depot/src/DepotServerClient.ts
```diff
@@ -236,23 +256,13 @@ export class DepotServerClient extends AbstractServerClient {
...
-    } else {
-      dependencies = await this.getDependencyEntities(
-        project.groupId,
-        project.artifactId,
-        versionId === LATEST_VERSION_ALIAS ? project.latestVersion : versionId,
-        true,
-        false,
-      );
-    }
+    const dependencies = await this.getDependencyEntities(
```
akphi, line 259: can you check if we can remove `getLatestRevisionDependencyEntities()` and remove that method as well? It's such a confusing one 😅

packages/legend-application-query/style/_query-editor.scss
```diff
@@ -157,3 +157,10 @@
     width: 100%;
   }
 }
+
+.query-productionize {
+  &-modal {
```
akphi, line 162: --> why don't we flatten this class -> `.query-productionize-modal`?

packages/legend-application-query/src/stores/QueryStudioSDLCState.ts
```diff
@@ -0,0 +1,152 @@
...
+import { SDLCServerClient } from '@finos/legend-server-sdlc';
+import {
+  type GeneratorFn,
+  assertErrorThrown,
+  guaranteeNonNullable,
+} from '@finos/legend-shared';
+import { action, flow, flowResult, makeObservable, observable } from 'mobx';
+import type { QueryEditorStore } from './QueryEditorStore.js';
+
+export class StudioSDLCInstance {
```
akphi, line 27: Do we _foresee_ this SDLC usage in Query is a one-time thing for the sake of only this flow, or it should become more and more useful. I can't tell, but if it's a one-time thing, I would rather we chuck all this logic into `QueryPromoteToServiceState`.

Also, this flow is not super straightforward, you're refreshing the page to make people go to the authorize URL over on gitlab. Would this disrupt the flow?

I also enjoy the fact that Query app starts with breeze and does not do some OAuth thing, so would this disrupt that? Could we limit the SDLC interaction to only the modal and somehow find a reasonable way to authenticate/authorize for this flow? You could test this by deleting your cache + cookie and try to see the refresh behavior

akphi (changes requested):
I haven't had time to review the component and state part, will take another look later. I'm concerned about the involvement of SDLC in this flow

packages/legend-application-query/src/stores/QueryStudioSDLCState.ts
MauricioUyaguari: It wouldn't disrupt this flow.. We only call authorize during the `init` flow when the modal is open for the first time or when we are chaning sdlc instances.

akphi (approved):

akphi:
Closes in favor of https://github.com/finos/legend-studio/pull/1506

Issue #1391, opened 2022-08-09
Bug: Default date value specification editor is throwing cast compilation error


YannanGao-gs:
### Similar issues

- [X] I have searched and found no existing similar issues

### How are you using Studio?

Legend Query

### Current and expected behavior

<img width="628" alt="Screen Shot 2022-08-09 at 4 15 59 PM" src="https://user-images.githubusercontent.com/73408381/183752767-8c67de03-56d3-4c87-9777-03756f06c008.png">

### Contribution

- [X] I would like to work on the fix for this issue

akphi:
Reopened, see the issue in https://github.com/finos/legend-studio/issues/1460 when drag-and-drop a derived property with a parameter of type `Date`, we automatically populate a dummy value of type `StrictDate`, causing a compilation error

PR #1392, opened 2022-08-10
fix a bug where default value of date type will throw a cast error

YannanGao-gs:
## Summary

Fixes https://github.com/finos/legend-studio/issues/1391

- [ ] Test(s) added


MauricioUyaguari (changes requested):
Don't we need to do this for every operator that supports date ?
So we don't get this casting issue there as well ?

MauricioUyaguari (changes requested):
thanks! lets just check the other operators to see whether this is needed

YannanGao-gs:
> thanks! lets just check the other operators to see whether this is needed

We do have the same issue for other operators. Since the default parameter is 'is' and we could choose other operators from the dropdown, therefore we update the valuespecification once is enough.

akphi:
@YannanGao-gs if you already had an issue and linked it with this PR, let's not milestone it

MauricioUyaguari (approved):

Issue #1403, opened 2022-08-15
Feature request: Add SDLC server to legend query application


MauricioUyaguari:
### Similar requests

- [X] I have searched and found no existing similar requests

### How are you using Studio?

Legend Query

### What problems are you trying to solve?

In order to support the editing of queries (services)  via their respective sdlc projects, we will add sdlcServerClient to legend query application. 

### Describe the solution you would like to see implemented

For now there will be a single instance of sdlc instances tied to legend query. 
Going forward we can improve the workflow and support multiple sdlc instances as multiple sdlc instances can be used to publish artifacts for which `legend-depot` reads from. 

### Contribution

- [X] I would like to work on this feature

PR #1405, opened 2022-08-15
add sdlc server to legend query application

MauricioUyaguari:
## Summary

- Closes https://github.com/finos/legend-studio/issues/1403

- [ ] Test(s) added

### DIFF
--- a/packages/legend-application-query-bootstrap/scripts/setup.js
+++ b/packages/legend-application-query-bootstrap/scripts/setup.js
@@ -45,7 +45,7 @@ export const setup = (outputDir) => {
           url: 'http://localhost:6060/api',
         },
         depot: {
-          url: 'http://localhost:8075/depot/api',
+          url: 'http://localhost:9090/depot/api',
         },
         studio: {
           url: 'http://localhost:8080/studio',
--- a/packages/legend-application-query-bootstrap/scripts/setup.js
+++ b/packages/legend-application-query-bootstrap/scripts/setup.js
@@ -45,10 +45,11 @@ export const setup = (outputDir) => {
           url: 'http://localhost:6060/api',
         },
         depot: {
-          url: 'http://localhost:9090/depot/api',
+          url: 'http://localhost:8075/depot/api',
         },
         studio: {
           url: 'http://localhost:8080/studio',
+          sdlcUrl: 'http://localhost:7070/api',
         },
       },
       undefined,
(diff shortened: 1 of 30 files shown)


packages/legend-application-studio/src/stores/LegendSdlcSetupState.ts
```diff
@@ -0,0 +1,144 @@
...
+import {
+  type GeneratorFn,
+  assertErrorThrown,
+  HttpStatus,
+  LogEvent,
+  NetworkClientError,
+} from '@finos/legend-shared';
+import { flow, makeObservable } from 'mobx';
+
+export class LegendSdlcSetupState {
```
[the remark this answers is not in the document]
akphi: Did I miss something, why do we need this state anymore?

akphi (changes requested):
Generally, I agree with the approach. It bothers me slightly that we now depend on `sdlc` in `legend-application`. I guess it's the right place though since it's the common place where different applications can reuse stuffs, in the future, we could consider splitting this into `@finos/legend-application-sdlc`.

That asides, there's also the need to address what one application can share with the rest of the world without the world depending on that application explicitly, so example is like `@finos/legend-application-studio-external` which holds the `router` info of `Legend Studio`

packages/legend-application-query/src/application/LegendQueryApplicationConfig.ts
```diff
@@ -66,7 +66,8 @@ export interface LegendQueryApplicationConfigurationData
     TEMPORARY__useLegacyDepotServerAPIRoutes?: boolean;
   };
   engine: { url: string; queryUrl?: string };
-  studio: { url: string };
+  studio: { url: string; sdlcUrl: string };
+  extensions?: Record<PropertyKey, unknown>;
```
akphi, line 70:
```suggestion
```

This comes naturally from `LegendApplicationConfigurationData` so we can remove it here

packages/legend-application-query-bootstrap/scripts/setup.js
```diff
@@ -45,10 +45,11 @@ export const setup = (outputDir) => {
           url: 'http://localhost:6060/api',
         },
         depot: {
-          url: 'http://localhost:9090/depot/api',
+          url: 'http://localhost:8075/depot/api',
```
akphi, line 48: Let's revert this

```suggestion
          url: 'http://localhost:9090/depot/api',
```

akphi (changes requested):

akphi (approved):

Issue #1435, opened 2022-08-24
Feature request: Allow loading query from inside of query builder


YannanGao-gs:
### Similar requests

- [X] I have searched and found no existing similar requests

### How are you using Studio?

Legend Query

### What problems are you trying to solve?

Allow loading query from inside of query builder: we might need to add more metrics around a query, such as latestDate, etc.

### Describe the solution you would like to see implemented

1. Load Query button
2. Query Loader panel
3. loading bar
4. Filter button
5. Search Bar
6. Data table

### Contribution

- [X] I would like to work on this feature

PR #1436, opened 2022-08-24
Allow loading query from inside of query builder

YannanGao-gs:
## Summary

Fixes https://github.com/finos/legend-studio/issues/1435

- [ ] Test(s) added

loading queries

https://user-images.githubusercontent.com/73408381/189791498-e5a545c4-b921-444d-bb3a-8c5d3abd8492.mov

https://user-images.githubusercontent.com/73408381/189760488-6ee1ea23-5597-4f1b-a672-e5660602d4ad.mov

No Query to Review

Screen Shot 2022-09-12 at 5 19 12 PM

### DIFF
--- a/packages/legend-application-query/src/components/QueryEditor.tsx
+++ b/packages/legend-application-query/src/components/QueryEditor.tsx
@@ -15,6 +15,7 @@
  */
 
 import {
+  type SelectComponent,
   Dialog,
   ArrowLeftIcon,
   ExternalLinkSquareIcon,
@@ -25,10 +26,14 @@ import {
   clsx,
   EmptyLightBulbIcon,
   LightBulbIcon,
+  SearchIcon,
+  TimesIcon,
+  CheckSquareIcon,
+  SquareIcon,
 } from '@finos/legend-art';
-import { getQueryParameters } from '@finos/legend-shared';
+import { debounce, getQueryParameters } from '@finos/legend-shared';
 import { observer } from 'mobx-react-lite';
-import { useEffect, useRef } from 'react';
+import { useEffect, useMemo, useRef, useState } from 'react';
 import { DndProvider } from 'react-dnd';
 import { HTML5Backend } from 'react-dnd-html5-backend';
 import { useParams } from 'react-router';
@@ -41,6 +46,7 @@ import {
   LEGEND_QUERY_QUERY_PARAM_TOKEN,
   LEGEND_QUERY_PATH_PARAM_TOKEN,
   generateStudioProjectViewUrl,
+  generateExistingQueryEditorRoute,
 } from '../stores/LegendQueryRouter.js';
 import {
   type QueryEditorStore,
@@ -50,7 +56,10 @@ import {
   ServiceQueryCreatorStore,
 } from '../stores/QueryEditorStore.js';
 import { QueryBuilder } from './query-builder/QueryBuilder.js';
-import { useApplicationStore } from '@finos/legend-application';
+import {
+  type ApplicationStore,
+  useApplicationStore,
+} from '@finos/legend-application';
 import {
   MappingQueryCreatorStoreProvider,
   ExistingQueryEditorStoreProvider,
@@ -58,12 +67,14 @@ import {
   useQueryEditorStore,
 } from './QueryEditorStoreProvider.js';
 import {
-  extractElementNameFromPath,
   type RawLambda,
+  extractElementNameFromPath,
 } from '@finos/legend-graph';
 import { flowResult } from 'mobx';
 import { useLegendQueryApplicationStore } from './LegendQueryBaseStoreProvider.js';
 import type { QueryBuilderState } from '../stores/query-builder/QueryBuilderState.js';
+import type { LegendQueryApplicationConfig } from '../application/LegendQueryApplicationConfig.js';
+import type { LegendQueryApplicationPlugin } from '../stores/LegendQueryApplicationPlugin.js';
 
 const QueryExportDialogContent = observer(
   (props: { exportState: QueryExportState }) => {
@@ -190,11 +201,221 @@ const renderQueryEditorHeaderLabel = (
   return null;
 };
 
+const QueryLoader = observer(
+  (props: {
+    editorStore: QueryEditorStore;
+    applicationStore: ApplicationStore<
+      LegendQueryApplicationConfig,
+      LegendQueryApplicationPlugin
+    >;
+  }) => {
+    const { editorStore, applicationStore } = props;
+    const queryFinderRef = useRef<SelectComponent>(null);
+    const searchInputRef = useRef<HTMLInputElement>(null);
+    const [selectedQueryID, setSelectedQueryID] = useState('');
+    const [isMineOnly, setIsMineOnly] = useState(false);
+    const [searchText, setSearchText] = useState('');
+    const closeQueryImporter = (): void => {
+      editorStore.queryLoaderState.setIsQueryLoaderOpen(false);
+    };
+    const handleEnterQueryImporter = (): void =>
+      queryFinderRef.current?.focus();
+    const toggleIsMineOnly = (): void => {
+      setIsMineOnly(!isMineOnly);
+    };
+    const loadSelectedQuery = (): void => {
+      if (selectedQueryID) {
+        editorStore.queryLoaderState.setIsQueryLoaderOpen(false);
+        applicationStore.navigator.jumpTo(
+          generateExistingQueryEditorRoute(selectedQueryID),
+        );
+      }
+    };
+    // search text
+    const debouncedLoadQueries = useMemo(
+      () =>
+        debounce((input: string): void => {
+          flowResult(editorStore.queryLoaderState.loadQueries(input)).catch(
+            applicationStore.alertUnhandledError,
+          );
+        }, 500),
+      [applicationStore, editorStore.queryLoaderState],
+    );
+    const onSearchTextChange: React.ChangeEventHandler<HTMLInputElement> = (
+      event,
+    ) => {
+      if (event.target.value !== searchText) {
+        setSearchText(event.target.value);
+        debouncedLoadQueries.cancel();
+        debouncedLoadQueries(event.target.value);
+      }
+    };
+    const clearQuerySearching = (): void => {
+      setSearchText('');
+      debouncedLoadQueries.cancel();
+      debouncedLoadQueries('');
+    };
+
+    useEffect(() => {
+      flowResult(editorStore.queryLoaderState.loadQueries('')).catch(
+        applicationStore.alertUnhandledError,
+      );
+    }, [applicationStore, editorStore.queryLoaderState]);
+
+    return (
+      <Dialog
+        open={editorStore.queryLoaderState.isQueryLoaderOpen}
+        onClose={closeQueryImporter}
+        TransitionProps={{
+          onEnter: handleEnterQueryImporter,
+        }}
+        classes={{ container: 'search-modal__container' }}
+        PaperProps={{ classes: { root: 'search-modal__inner-container' } }}
+      >
+        <div className="modal modal--dark search-modal">
+          <div className="modal__title">Load Query</div>
+          <div className="query-editor__query-loader__filter-section">
+            <div className="query-editor__query-loader__filter-section__section__toggler">
+              <button
+                className={clsx(
+                  'query-editor__query-loader__filter-section__section__toggler__btn',
+                  {
+                    'query-editor__query-loader__filter-section__section__toggler__btn--toggled':
+                      isMineOnly,
+                  },
+                )}
+                onClick={toggleIsMineOnly}
+                tabIndex={-1}
+              >
+                {isMineOnly ? <CheckSquareIcon /> : <SquareIcon />}
+              </button>
+              <div
+                className="query-editor__query-loader__filter-section__section__toggler__prompt"
+                onClick={toggleIsMineOnly}
+              >
+                Mine Only
+              </div>
+            </div>
+          </div>
+          <div className="query-editor__query-loader__search-section">
+            <div className="query-editor__query-loader__search-section__input__container">
+              <input
+                ref={searchInputRef}
+                className={clsx(
+                  'query-editor__query-loader__search-section__input input--dark',
+                  {
+                    'query-editor__query-loader__search-section__input--searching':
+                      searchText,
+                  },
+                )}
+                onChange={onSearchTextChange}
+                value={searchText}
+                placeholder="Search a query by name"
+              />
+              {!searchText ? (
+                <div className="query-editor__query-loader__search-section__input__search__icon">
+                  <SearchIcon />
+                </div>
+              ) : (
+                <>
+                  <button
+                    className="query-editor__query-loader__search-section__input__clear-btn"
+                    tabIndex={-1}
+                    onClick={clearQuerySearching}
+                    title="Clear"
+                  >
+                    <TimesIcon />
+                  </button>
+                </>
+              )}
+            </div>
+          </div>
+          <div className="query-editor__query-loader__body">
+            {editorStore.queryLoaderState.loadQueriesState.hasCompleted && (
+              <>
+                {editorStore.queryLoaderState.queries.length > 0 && (
+                  <>
+                    <table className="table">
+                      <thead>
+                        <tr>
+                          <th className="table__cell--left">Name</th>
+                          <th className="table__cell--left">Author</th>
+                          <th className="table__cell--left">Version</th>
+                        </tr>
+                      </thead>
+                      <tbody>
+                        {(isMineOnly
+                          ? editorStore.queryLoaderState.queries.filter(
+                              (q) => q.isCurrentUserQuery,
+                            )
+                          : editorStore.queryLoaderState.queries
+                        ).map((query) => (
+                          <tr
+                            key={query.id}
+                            className={clsx(
+                              'query-editor__query-loader__body__table__row',
+                              {
+                                'query-editor__query-loader__body__table__row--selected':
+                                  selectedQueryID === query.id,
+                              },
+                            )}
+                            onClick={(event) => setSelectedQueryID(query.id)}
+                          >
+                            <td className="table__cell--left">{query.name}</td>
+                            <td className="table__cell--left">
+                              {query.owner ?? 'anonymous'}
+                            </td>
+                            <td className="table__cell--left">
+                              {query.versionId}
+                            </td>
+                          </tr>
+                        ))}
+                      </tbody>
+                    </table>
+                  </>
+                )}
+                {editorStore.queryLoaderState.queries.length === 0 && (
+                  <BlankPanelContent>No query available</BlankPanelContent>
+                )}
+              </>
+            )}
+            {!editorStore.queryLoaderState.loadQueriesState.hasCompleted && (
+              <>
+                <PanelLoadingIndicator
+                  isLoading={
+                    !editorStore.queryLoaderState.loadQueriesState.hasCompleted
+                  }
+                />
+                <BlankPanelContent>Loading queries...</BlankPanelContent>
+              </>
+            )}
+          </div>
+          <div className="search-modal__actions">
+            <button
+              className="btn btn--dark"
+              onClick={loadSelectedQuery}
+              disabled={selectedQueryID === ''}
+            >
+              Load Query
+            </button>
+            <button className="btn btn--dark" onClick={closeQueryImporter}>
+              Close
+            </button>
+          </div>
+        </div>
+      </Dialog>
+    );
+  },
+);
+
 const QueryEditorHeaderContent = observer(
   (props: { queryBuilderState: QueryBuilderState }) => {
     const { queryBuilderState } = props;
     const editorStore = useQueryEditorStore();
     const applicationStore = useLegendQueryApplicationStore();
+    const openQueryLoader = (): void => {
+      editorStore.queryLoaderState.setIsQueryLoaderOpen(true);
+    };
     const viewQueryProject = (): void => {
       const { groupId, artifactId, versionId } = editorStore.getProjectInfo();
       applicationStore.navigator.openNewWindow(
@@ -233,6 +454,15 @@ const QueryEditorHeaderContent = observer(
         </div>
         <div className="query-editor__header__actions">
           <button
+            className="query-editor__header__action btn--dark"
+            tabIndex={-1}
+            onClick={openQueryLoader}
+          >
+            <div className="query-editor__header__action__label">
+              Load Query
+            </div>
+          </button>
+          <button
             className="query-editor__header__action query-editor__header__action--simple btn--dark"
             tabIndex={-1}
             title="View project"
@@ -253,6 +483,12 @@ const QueryEditorHeaderContent = observer(
                 <LightBulbIcon />
               )}
             </button>
+          )}
+          {editorStore.queryLoaderState.isQueryLoaderOpen && (
+            <QueryLoader
+              editorStore={editorStore}
+              applicationStore={applicationStore}
+            />
           )}
           <button
             className="query-editor__header__action btn--dark"
--- a/packages/legend-application-query/src/stores/QueryEditorStore.ts
+++ b/packages/legend-application-query/src/stores/QueryEditorStore.ts
@@ -50,6 +50,7 @@ import {
   GraphManagerTelemetry,
   extractElementNameFromPath,
   isSystemElement,
+  QuerySearchSpecification,
   Mapping,
   type Runtime,
 } from '@finos/legend-graph';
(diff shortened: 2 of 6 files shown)


packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -189,9 +201,225 @@ const renderQueryEditorHeaderLabel = (
...
+      LegendQueryApplicationPlugin
+    >;
+  }) => {
+    const { editorStore, applicationStore } = props;
+    const queryFinderRef = useRef<SelectComponent>(null);
+    const [selectedQueryID, setSelectedQueryID] = useState('');
+    const queryLoaderColumns = ['Name', 'Author', 'Version'];
+    const closeQueryImporter = (): void => {
+      editorStore.setIsQueryLoaderOpen(!editorStore.isQueryLoaderOpen);
+    };
```
gayathrir11, line 218: Can we explicitly mention that we set `isQueryLoaderOpen` to false?

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -189,9 +201,225 @@ const renderQueryEditorHeaderLabel = (
...
+    const handleEnterQueryImporter = (): void =>
+      queryFinderRef.current?.focus();
+    const loadSelectedQuery = (): void => {
+      if (selectedQueryID) {
+        editorStore.setIsQueryLoaderOpen(!editorStore.isQueryLoaderOpen);
```
gayathrir11, line 223: Can we explicitly mention that we set isQueryLoaderOpen to false?

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -189,9 +201,225 @@ const renderQueryEditorHeaderLabel = (
...
+      query.name,
+      query.owner ?? 'anonymous',
+      query.versionId,
+    ];
+    const searchInputRef = useRef<HTMLInputElement>(null);
+    const searchQuery: React.ChangeEventHandler<HTMLInputElement> = (event) => {
+      editorStore.setSearchText(event.target.value);
+      if (editorStore.searchText.length >= 0) {
+        flowResult(editorStore.loadQueries(editorStore.searchText))
+          .then(() => editorStore.setIsLoadingQueryInProgress(false))
```
gayathrir11, line 244: Can we may be abstract this logic in a function because we are using piece of code in three different places?
```
flowResult(editorStore.loadQueries(editorStore.searchText))
          .then(() => editorStore.setIsLoadingQueryInProgress(false))
          .catch(applicationStore.alertUnhandledError);
```

packages/legend-application-query/src/stores/QueryEditorStore.ts
```diff
@@ -205,6 +206,11 @@ export abstract class QueryEditorStore {
   initState = ActionState.create();
   queryBuilderState: QueryBuilderState;
   exportState?: QueryExportState | undefined;
+  isQueryLoaderOpen = false;
```
MauricioUyaguari, line 209: I would move this out to its own state.

packages/legend-application-query/src/stores/QueryEditorStore.ts
```diff
@@ -391,6 +428,24 @@ export abstract class QueryEditorStore {
       graphBuilderReportData,
     );
   }
+
+  *loadQueries(searchText: string): GeneratorFn<void> {
+    const isValidSearchString = searchText.length >= 3;
```
MauricioUyaguari, line 433: 3 should be a constant.

packages/legend-application-query/src/stores/QueryEditorStore.ts
```diff
@@ -391,6 +428,24 @@ export abstract class QueryEditorStore {
...
+    this.isLoadingQueryInProgress = true;
+    try {
+      const searchSpecification = new QuerySearchSpecification();
+      searchSpecification.searchTerm = isValidSearchString
+        ? searchText
+        : undefined;
+      searchSpecification.limit = 10;
```
MauricioUyaguari, line 440: same with 10

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -189,9 +201,222 @@ const renderQueryEditorHeaderLabel = (
...
+      </Dialog>
+    );
+  },
+);
+
 const QueryEditorHeaderContent = observer(() => {
   const editorStore = useQueryEditorStore();
   const applicationStore = useLegendQueryApplicationStore();
+  const loadQuery = (): void => {
+    editorStore.setIsQueryLoaderOpen(!editorStore.isQueryLoaderOpen);
```
MauricioUyaguari, line 415: can we call this `openQueryLoader` also we should be setting `true` or `false` in the input to make it clearer what we are expecting the query loader to do, unless the action is specific to `toggle`

MauricioUyaguari (changes requested):
small comments. Looks good. Thanks

akphi:
@YannanGao-gs If you already linked this PR to an issue which is already milestoned (`8.0.0` in this case), let's not milestone the PR, this creates duplicates in our project board. Thanks!

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -189,9 +201,219 @@ const renderQueryEditorHeaderLabel = (
...
+                        ))}
+                      </tbody>
+                    </table>
+                    <div className="query-editor__query-loader__body__warning">
+                      Show relevant queries by searching names
+                    </div>
+                  </>
+                )}
+                {editorStore.queries.length === 0 && (
+                  <BlankPanelContent>No Query to Preview</BlankPanelContent>
```
akphi, line 381: --> `No query available`

packages/legend-application-query/src/stores/QueryEditorStore.ts
akphi: +1, we have enough state here to justify that we need its own state

packages/legend-application-query/src/stores/QueryEditorStore.ts
```diff
@@ -236,6 +253,26 @@ export abstract class QueryEditorStore {
...
+
+  setIsLoadingQueryInProgress(val: boolean): void {
+    this.isLoadingQueryInProgress = val;
+  }
+
+  setQueries(val: LightQuery[]): void {
+    this.queries = val;
+  }
+
+  setSearchText(val: string): void {
```
akphi, line 268: why can't we do this the same way we do it in `ExistingQuerySetupState.loadQueries()`? We don't need to store these many fields in the state, we can store them as component state instead, bebcause the modal is opened up and then the 2 actions people can do is:
1. Load a new query that will reload the app
2. Cancel and close the modal

I don't see much point in persisting these states in the `QueryEditorState` like this.

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -189,9 +201,219 @@ const renderQueryEditorHeaderLabel = (
...
+    editorStore: QueryEditorStore;
+    applicationStore: ApplicationStore<
+      LegendQueryApplicationConfig,
+      LegendQueryApplicationPlugin
+    >;
+  }) => {
+    const { editorStore, applicationStore } = props;
+    const queryFinderRef = useRef<SelectComponent>(null);
+    const [selectedQueryID, setSelectedQueryID] = useState('');
+    const queryLoaderColumns = ['Name', 'Author', 'Version'];
```
akphi, line 215: It's fine, but let's inline this in the component, I don't think we need to do something like this

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -189,9 +201,219 @@ const renderQueryEditorHeaderLabel = (
...
+                <PanelLoadingIndicator
+                  isLoading={editorStore.isLoadingQueryInProgress}
+                />
+                <BlankPanelContent>loading queries...</BlankPanelContent>
+              </>
+            )}
+          </div>
+          <div className="search-modal__actions">
+            <button className="btn btn--dark" onClick={loadSelectedQuery}>
+              Load
```
akphi, line 396: -> `Load Query`

Also while we're here, I think we should create an issue to handle these kinds of flow, user could lose their progress on the query here and we should block that, the same way we block in `Studio`.

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -189,9 +201,219 @@ const renderQueryEditorHeaderLabel = (
...
+        TransitionProps={{
+          onEnter: handleEnterQueryImporter,
+        }}
+        classes={{ container: 'search-modal__container' }}
+        PaperProps={{ classes: { root: 'search-modal__inner-container' } }}
+      >
+        <div className="modal modal--dark search-modal">
+          <div className="modal__title">Select a Query to Load</div>
+          <div className="query-editor__query-loader__filter-section">
+            <div className="query-editor__query-loader__filter-section__header__label">
```
akphi, line 270: Hm, I would say for now, let's remove this, it looks similar to the header of the modal. Please remove the CSS class accordingly

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -189,9 +201,219 @@ const renderQueryEditorHeaderLabel = (
...
+                              {
+                                'query-editor__query-loader__body__table__row--selected':
+                                  selectedQueryID === q.id,
+                              },
+                            )}
+                            onClick={(event) => setSelectedQueryID(q.id)}
+                          >
+                            {buildQueryTableRow(q).map((value, columnIdx) => (
+                              // eslint-disable-next-line react/no-array-index-key
+                              <td key={columnIdx} className="table__cell--left">
```
akphi, line 367: could we index here by the query ID?

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -189,9 +201,219 @@ const renderQueryEditorHeaderLabel = (
...
+                                {value}
+                              </td>
+                            ))}
+                          </tr>
+                        ))}
+                      </tbody>
+                    </table>
+                    <div className="query-editor__query-loader__body__warning">
```
akphi, line 375: Didn't we discuss about this? I think for now we could remove this message. The way I think about this is, if we are to show this message at all, we need to do a little more work. Let's say, you are trying to show 10 queries, you need to query for 11, if 11 comes back, you can show 10 and show this warning. But that's some work to do, @MauricioUyaguari what do you think about this?

YannanGao-gs: I deleted this section for now.

packages/legend-application-query/src/stores/QueryEditorStore.ts
```diff
@@ -211,6 +216,9 @@ export abstract class QueryEditorStore {
   initState = ActionState.create();
   queryBuilderState?: QueryBuilderState | undefined;
   exportState?: QueryExportState | undefined;
+  isQueryLoaderOpen = false;
+  loadQueriesState = ActionState.create();
+  queries: LightQuery[] = [];
 
   constructor(
     applicationStore: LegendQueryApplicationStore,
```
akphi, line 224: Hm, let's organize these into `QueryLoaderState` like what we did in `QueryExportState`

packages/legend-application-query/src/stores/QueryEditorStore.ts
```diff
@@ -393,6 +415,25 @@ export abstract class QueryEditorStore {
...
+      const searchSpecification = new QuerySearchSpecification();
+      searchSpecification.searchTerm = isValidSearchString
+        ? searchText
+        : undefined;
+      searchSpecification.limit = QUERY_BUILDER_QUERY_LOAD_MAX_AMOUNT;
+      this.queries = (yield this.graphManagerState.graphManager.searchQueries(
+        searchSpecification,
+      )) as LightQuery[];
+    } catch (error) {
+      assertErrorThrown(error);
```
akphi, line 433: where do you reset `loadQueriesState`?

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -190,11 +201,216 @@ const renderQueryEditorHeaderLabel = (
...
+          message: 'Loading query...',
+          showLoading: true,
+        });
+      } else {
+        applicationStore.notifyWarning(
+          'No query is selected. Please select a query to load or close.',
+        );
+      }
+    };
+    const searchInputRef = useRef<HTMLInputElement>(null);
```
akphi, line 239: see what we do in `EditExistingQuerySetup` and please bring over the `debounce` logic as well as `useEffect()` logic. We cannot search on every keystroke user makes and also, in the very beginning, we would like to already show some queries

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -190,11 +201,216 @@ const renderQueryEditorHeaderLabel = (
...
+    const loadQueries = (text: string): void => {
+      flowResult(editorStore.loadQueries(text))
+        .then(() => editorStore.loadQueriesState.pass())
+        .catch(applicationStore.alertUnhandledError);
+    };
+    const searchQuery: React.ChangeEventHandler<HTMLInputElement> = (event) => {
+      setSearchText(event.target.value);
+      loadQueries(event.target.value);
+    };
```
akphi, line 248: Isn't this expensive? You can just set `queries = []`, this will make another network call!

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -190,11 +201,216 @@ const renderQueryEditorHeaderLabel = (
...
+
 const QueryEditorHeaderContent = observer(
   (props: { queryBuilderState: QueryBuilderState }) => {
     const { queryBuilderState } = props;
     const editorStore = useQueryEditorStore();
     const applicationStore = useLegendQueryApplicationStore();
+    const openQueryLoader = (): void => {
+      editorStore.setIsQueryLoaderOpen(true);
+      flowResult(editorStore.loadQueries(''))
+        .then(() => editorStore.loadQueriesState.pass())
```
akphi, line 411: please put this in the method `loadQueries()` why do you have this here?

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -190,11 +201,216 @@ const renderQueryEditorHeaderLabel = (
...
+);
+
 const QueryEditorHeaderContent = observer(
   (props: { queryBuilderState: QueryBuilderState }) => {
     const { queryBuilderState } = props;
     const editorStore = useQueryEditorStore();
     const applicationStore = useLegendQueryApplicationStore();
+    const openQueryLoader = (): void => {
+      editorStore.setIsQueryLoaderOpen(true);
+      flowResult(editorStore.loadQueries(''))
```
akphi, line 410: also, it's fine to make this call here for the first time we bring this loader up, but see my note on mimicking what we do in `EditExistingQuerySetup`, we will do this in the `useEffect` there.

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -190,11 +201,216 @@ const renderQueryEditorHeaderLabel = (
...
+                            <td className="table__cell--left">
+                              {query.owner ?? 'anonymous'}
+                            </td>
+                            <td className="table__cell--left">
+                              {query.versionId}
+                            </td>
+                          </tr>
+                        ))}
+                      </tbody>
+                    </table>
```
akphi, line 372: Did you miss my previous comment on this? We should not show this anymore. If you insist, we need to do this slightly differently

akphi (changes requested):
Nice attempt to approach the problem from scratch, but you haven't accounted for some minor details like `debouncing`, let's mimic what we do in `EditExistingQuerySetup` and revise this PR. Thanks!

packages/legend-application-query/src/stores/QueryEditorStore.ts
akphi: @YannanGao-gs let's not forget about this :)

packages/legend-application-query/src/components/QueryEditor.tsx
YannanGao-gs: Are you referring to this one? i think i already deleted it as requested. 
Screen Shot 2022-09-12 at 3 56 14 PM

packages/legend-application-query/src/stores/query-builder/QueryBuilderConfig.ts
```diff
@@ -25,6 +25,8 @@ export const DEFAULT_POST_FILTER_LAMBDA_VARIABLE_NAME = 'row';
 
 export const QUERY_BUILDER_PROPERTY_SEARCH_MAX_DEPTH = 5;
 export const QUERY_BUILDER_PROPERTY_SEARCH_RESULTS_LIMIT = 100;
+export const QUERY_BUILDER_SEARCH_TEXT_MIN_LENGTH = 3;
```
akphi, line 28: this is not the right place to put it. this file is meant for `QueryBuilder` core, the stuff you're adding is meant for `QueryLoader` which is part of the `Legend Query` app. Please follow my comment in the previous review, these should be put in the same file as `QueryEditorStore` I believe.

packages/legend-application-query/src/components/QueryEditor.tsx
YannanGao-gs: By setting queries = [], it will make table empty, which is not consistent with what we will show when the query loader is open.
I could create a new vairable to store the first 10 (QUERY_BUILDER_QUERY_LOAD_MAX_AMOUNT) queries to avoid another network call.

packages/legend-application-query/src/components/QueryEditor.tsx
akphi: strangely, I saw it in my review, nvm then, looks like it's gone now!

YannanGao-gs: thanks! resolved this comment!

packages/legend-application-query/src/components/QueryEditor.tsx
akphi: I see your point, so let's query for `''` then, don't store the first 10 queries

YannanGao-gs: gotcha. I changed it back. thanks!

packages/legend-application-query/src/stores/QueryEditorStore.ts
```diff
@@ -202,6 +206,55 @@ export class QueryExportState {
   }
 }
 
+export class QueryLoaderState {
+  editorStore: QueryEditorStore;
+  isQueryLoaderOpen = false;
+  loadQueriesState = ActionState.create();
+  queries: LightQuery[] = [];
+
```
akphi, line 214: Let's not do this, query list can go stale

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -190,11 +201,232 @@ const renderQueryEditorHeaderLabel = (
...
+                        <tr>
+                          {['Name', 'Author', 'Version'].map((column, idx) => (
+                            <th
+                              key={column}
+                              style={{ fontWeight: 'bold' }}
+                              className="table__cell--left"
+                            >
+                              {column}
+                            </th>
+                          ))}
```
akphi, line 358: why don't use css for this instead?

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -190,11 +201,232 @@ const renderQueryEditorHeaderLabel = (
...
+              <>
+                {editorStore.queryLoaderState.queries.length > 0 && (
+                  <>
+                    <table className="table">
+                      <thead style={{ position: 'sticky', top: -1 }}>
+                        <tr>
+                          {['Name', 'Author', 'Version'].map((column, idx) => (
+                            <th
+                              key={column}
+                              style={{ fontWeight: 'bold' }}
```
akphi, line 353: why don't use css for this instead?

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -190,11 +201,232 @@ const renderQueryEditorHeaderLabel = (
...
+                              className="table__cell--left"
+                            >
```
akphi, line 355: Let's just flatten these to 3 `th`s instead. Thanks!

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -190,11 +201,232 @@ const renderQueryEditorHeaderLabel = (
...
+        classes={{ container: 'search-modal__container' }}
+        PaperProps={{ classes: { root: 'search-modal__inner-container' } }}
+      >
+        <div className="modal modal--dark search-modal">
+          <div className="modal__title">Select a Query to Load</div>
+          <div className="query-editor__query-loader__filter-section">
+            <div className="query-editor__query-loader__filter-section__section__toggler">
+              <button
+                className={clsx(
+                  'query-editor__query-loader__filter-section__section__toggler__btn',
```
akphi, line 290: --> `Load Query`

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -190,11 +201,232 @@ const renderQueryEditorHeaderLabel = (
...
+        editorStore.queryLoaderState.setIsQueryLoaderOpen(false);
+        applicationStore.navigator.goTo(
+          generateExistingQueryEditorRoute(selectedQueryID),
+        );
+        window.location.reload();
+        applicationStore.setBlockingAlert({
+          message: 'Loading query...',
+          showLoading: true,
+        });
+      } else {
```
akphi, line 237: I think we don't need this else block, instead, let's change the tooltip of the `Load Query` button when it's disabled because user hasn't selected any query.

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -190,11 +201,232 @@ const renderQueryEditorHeaderLabel = (
...
+              </>
+            )}
+          </div>
+          <div className="search-modal__actions">
+            <button className="btn btn--dark" onClick={loadSelectedQuery}>
+              Load Query
+            </button>
+            <button className="btn btn--dark" onClick={closeQueryImporter}>
+              Close
+            </button>
```
akphi, line 414: we should disable this when there's no query selected

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -190,11 +201,232 @@ const renderQueryEditorHeaderLabel = (
...
+    const clearQuerySearching = (): void => {
+      setSearchText('');
+      debouncedLoadQueries.cancel();
+      debouncedLoadQueries('');
+    };
+
+    useEffect(() => {
+      flowResult(editorStore.queryLoaderState.loadQueries('')).catch(
+        applicationStore.alertUnhandledError,
+      );
```
akphi, line 271: let's not do this

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -190,11 +201,232 @@ const renderQueryEditorHeaderLabel = (
...
+    ) => {
+      if (event.target.value !== searchText) {
+        setSearchText(event.target.value);
+        debouncedLoadQueries.cancel();
+        debouncedLoadQueries(event.target.value);
+      }
+    };
+    const clearQuerySearching = (): void => {
+      setSearchText('');
+      debouncedLoadQueries.cancel();
```
akphi, line 264: probably don't need this. Let's just clear the search text, this will trigger the `useEffect()`

akphi (changes requested):

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -190,11 +201,226 @@ const renderQueryEditorHeaderLabel = (
...
+    const toggleIsMineOnly = (): void => {
+      setIsMineOnly(!isMineOnly);
+    };
+    const loadSelectedQuery = (): void => {
+      if (selectedQueryID) {
+        editorStore.queryLoaderState.setIsQueryLoaderOpen(false);
+        applicationStore.navigator.goTo(
+          generateExistingQueryEditorRoute(selectedQueryID),
+        );
+        window.location.reload();
```
akphi, line 232: let's use `jumpTo` so you don't need to do `window.location.reload()`. We use methods from `navigator` and rarely poke directly into the `window` object. Also, I think we don't need `setBlockingAlert` here

akphi (changes requested):

packages/legend-application-query/src/components/QueryEditor.tsx
YannanGao-gs: re: setBlockingAlert.  I added it for this previous comment. 
Screen Shot 2022-09-13 at 11 49 32 AM
May I delete it?

akphi (approved):

PR #1442, opened 2022-08-26
query builder: cleanup fetch-structure state

akphi:
## Summary

- [x] Upgraded to `Yarn@3.2.3`
- [x] Bump dependencies
- [x] Cleanup post https://github.com/finos/legend-studio/pull/1439
- [x] Move projection actions inside of projection panel
  - [x] Move logics out of `ValueSpecificationBuilder` into smaller builders and do the same for `ValueSpecificationProcessor`
  - [x] Simplify and remove `QueryBuilderGraphFetchTreeState.initialize`
  - [x] Refactor fetch-structure state to only have one implementation (either `graph-fetch` or `projection`) at a time
- [x] Improve Docker workflow
  - [x] Make workflow run for Docker on schedule and only when we modify the Dockerfiles - refer to https://github.com/finos/legend-sdlc/pull/501
  - [x] Also allow us to trigger that Check Docker workflow manually
  - [x] Show a badge for Docker check status
- [x] Cleanup and improve subtype supports
  - [x] For already used properties, highlight the text instead of the icon of the tree node
  - [x] Cleanup post https://github.com/finos/legend-studio/pull/1367
    - [x] Fix algorithm to determine already used properties
      - [x] For projection, let's flat down chains of subtype, also let's scan upwards to find all supertypes used, and let's use set to avoid duplication
      - [x] For graph-fetch, let's use BFS to navigate and build the ID
  - [x] Change the projection column alias generation
  - [x] Also when we generate the projection expression using DnD, simplify to just 1 cast instead of nested casting like what we do currently, i.e. simplify subtypes chain
  - [x] Add and cleanup tests

- [x] Test(s) added


MauricioUyaguari (approved):

MauricioUyaguari (changes requested):
conflicts

MauricioUyaguari (approved):

PR #1468, opened 2022-09-08
Propagate default milestoning dates for aggregate operations in projection

gayathrir11:
## Summary

- Related to https://github.com/finos/legend-studio/issues/1471
- handle propagation for aggregation functions. We will not remove the date since propagation is not supported

https://user-images.githubusercontent.com/87973126/189183263-32341fc5-1109-4d29-affe-2fd66c027185.mp4

- [ ] Test(s) added

### DIFF
--- a/packages/legend-application-query/src/stores/QueryBuilderValueSpecificationBuilderHelper.ts
+++ b/packages/legend-application-query/src/stores/QueryBuilderValueSpecificationBuilderHelper.ts
@@ -77,7 +77,11 @@ const isDefaultDatePropagationSupported = (
   currentPropertyExpression: AbstractPropertyExpression,
   queryBuilderState: QueryBuilderState,
   prevPropertyExpression?: AbstractPropertyExpression | undefined,
+  isDatePropagationNotSupported?: boolean,
 ): boolean => {
+  if (isDatePropagationNotSupported) {
+    return false;
+  }
   const property = currentPropertyExpression.func;
   const graph = queryBuilderState.graphManagerState.graph;
   // Default date propagation is not supported for current expression when the previous property expression is a derived property.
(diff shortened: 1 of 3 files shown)


MauricioUyaguari (approved):

MauricioUyaguari (approved):

packages/legend-application-query/src/stores/QueryBuilderValueSpecificationBuilderHelper.ts
```diff
@@ -175,6 +180,7 @@ export const buildPropertyExpressionChain = (
               nextExpression instanceof AbstractPropertyExpression
                 ? nextExpression
                 : undefined,
+              isDatePropagationNotSupported,
```
akphi, line 183: why do we need to pass this to `isDefaultDatePropagationSupported()` anymore? we can just stop here and do `!disableDatePropagation && isDefaultDatePropagationSupported(...)`

packages/legend-application-query/src/stores/QueryBuilderValueSpecificationBuilderHelper.ts
```diff
@@ -138,6 +142,7 @@ const isDefaultDatePropagationSupported = (
 export const buildPropertyExpressionChain = (
   propertyExpression: AbstractPropertyExpression,
   queryBuilderState: QueryBuilderState,
+  isDatePropagationNotSupported?: boolean,
```
akphi, line 145: --> `disableDatePropagation`, let's also add documentation this flag. Every option-orientated param needs an explanation, else we could keep adding options like this on and on and at some point, we won't know what is each for anymore 😓

akphi (changes requested):
Since we don't really file an issue for this, I don't know if this solve what issue, could we add some sort of a test for this? Thanks!

MauricioUyaguari:
Agree with An. Additionally we should add tests for this PR and do some more investigation on what functions aren't supported for date propagation. 
I've create issue: https://github.com/finos/legend-studio/issues/1471

akphi:
@gayathrir11 we could think of a way to also industrialize these tests, basically, these would be similar to lambda roundtrip test, but this is about rewriting query a bit right? I might be wrong, but let's figure out a way so we can systematically test these. Milestoning is complex and solidifying them is the only way to get things right eventually. Even #1471 should be carried out in this fashion I think, every new case we find, even if we don't decide to support that function, we need a test to back that decision

Issue #304, opened 2021-06-28
Feature request: Tab bar UX improvements


akphi:
This regards the main element tab bar, but it might also be applicable to element editor (e.g. `mapping editor`) tab bars.

- [x] https://github.com/finos/legend-studio/issues/130
- [x] ~~Support translating vertical to horizontal scroll for tab bars: use `onWheel` event to catch scroll event and then translate that to horizontal scrolling - see `vscode` for example - solution in https://github.com/finos/legend-studio/pull/1711 is not complete~~ <-- this turns out to be not so trivial, also users can use `Shift + scroll` to translate vertical to horizontal scrolls
- [ ] #1952
```
Here we should follow the behaviour of Chrome. Let's say if we have the following tabs:
a,b,c,d,e
If we open a new tab from c, we will have:
a,b,c,f,d,e
The order of tabs being opened will also help us find tab to switch to when we close a tab
Also, with this, we can support `go back` feature and support `open last closed tab` feature
```
- [x] Apply changes in https://github.com/finos/legend-studio/pull/1711 to `mapping editor`

akphi:
@xannem I think point (2) is very doable, and I know I told you about not doing point (3), but the right way to do it I think is implement a naive address system in `Legend Studio` where we will just store the history as a stack of addresses, instead of the actual tab state (<- this is dangerous! can cause mem-leak)

akphi:
Will close this for now as the majority of the work is done by #1711

PR #1474, opened 2022-09-10
feat: allow rearranging main element tabs

### LINKED ISSUE #130: Feature request: Allow re-arranging editor tabs by drag-and-drop
### Motivation
We should allow re-arranging tabs by drag-and-drop. First, we can start with the general editor tab bar, for the bonus, we can implement the same changes to the mapping editor.

xannem:
## Summary

fixes #130 
builds on #304 

- [x] allowing rearranging to main element tabs 
Kapture 2022-09-18 at 23 37 37
- [x]  allows for vertical scrolling translated to horizontal scrolling to main element and mapped editor tabs
- [x] allowing rearranging to mapped editor tabs
Kapture 2022-09-22 at 21 12 13

<s>
- [ ]  option for opening last closed tab</s>  will not included for now as method through stack implementation is major potential root cause for memory leak

- [x] componentize EditPanel and MappingEditor tabs to something like `GeneralMainTabEditor`
- [x] dropdown button on the bar to show all open tabs 

Kapture 2022-11-04 at 15 22 46

- [ ] Test(s) added

### DIFF
--- a/packages/legend-application-studio/src/components/editor/edit-panel/mapping-editor/PurePropertyMappingEditor.tsx
+++ b/packages/legend-application-studio/src/components/editor/edit-panel/mapping-editor/PurePropertyMappingEditor.tsx
@@ -67,7 +67,9 @@ const SimplePropertyMappingEditor = observer(
     const propertyMapping = propertyMappingState.propertyMapping;
     const expectedType =
       propertyMapping.property.value.genericType.value.rawType;
-    const canDrop = dragItem && dragItem.data.type === expectedType;
+
+    const canDrop = dragItem?.data?.type === expectedType;
+
     const onExpectedTypeLabelSelect = (): void =>
       propertyMappingState.instanceSetImplementationState.setSelectedType(
         expectedType,
@@ -183,9 +185,9 @@ const EnumerationPropertyMappingEditor = observer(
     // NOTE: when we drag enum, we should highlight if the enumeration where that enum is part of matches
     const canDrop =
       dragItem &&
-      ((dragItem.data.type && dragItem.data.type === expectedType) ||
+      ((dragItem.data?.type && dragItem.data.type === expectedType) ||
         (dragItemType === CORE_DND_TYPE.TYPE_TREE_ENUM &&
-          dragItem.data.parent === expectedType));
+          dragItem.data?.parent === expectedType));
 
     return (
       <div className="property-mapping-editor__entry__container">
@@ -271,7 +273,8 @@ const ClassPropertyMappingEditor = observer(
       }
     };
     // Drag and Drop
-    const canDrop = dragItem?.data.type && dragItem.data.type === expectedType;
+
+    const canDrop = dragItem?.data?.type && dragItem.data.type === expectedType;
 
     return (
       <div className="property-mapping-editor__entry__container">
@@ -333,10 +336,12 @@ export const PurePropertyMappingEditor = observer(
             // if the dragged node is enum, when dropped, we want to have it as a constant
             let toAppend = '';
             if (dropType === CORE_DND_TYPE.TYPE_TREE_ENUM) {
-              toAppend = `${dropItem.data.parent.path}.${dropItem.data.label}`;
+              toAppend = `${dropItem.data?.parent.path}.${dropItem.data?.label}`;
             } else {
-              toAppend = dropItem.data.id;
-              if (dropItem.data.property instanceof DerivedProperty) {
+              if (dropItem.data?.id) {
+                toAppend = dropItem.data.id;
+              }
+              if (dropItem.data?.property instanceof DerivedProperty) {
                 toAppend += '()';
               }
             }
--- a/packages/legend-application-studio/src/stores/EditorStore.ts
+++ b/packages/legend-application-studio/src/stores/EditorStore.ts
@@ -283,6 +283,7 @@ export class EditorStore implements CommandRegistrar {
     this.pluginManager = applicationStore.pluginManager;
 
     this.editorMode = new StandardEditorMode(this);
+    this.tabHistoryBack = new Stack<EditorState>();
 
     this.sdlcState = new EditorSDLCState(this);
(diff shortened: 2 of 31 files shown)


packages/legend-application-studio/src/stores/EditorStore.ts
```diff
@@ -224,6 +226,7 @@ export class EditorStore {
   // Tabs
   currentEditorState?: EditorState | undefined;
   openedEditorStates: EditorState[] = [];
+  tabHistory: Stack<EditorState>;
```
akphi, line 229: very nice, we used to have this actually, but we removed as it could cause memory leak. Let's remove this for now. For further explanation, see the note in `updateGraphAndApplication`

packages/legend-art/src/components/panel/PanelDropZone.tsx
```diff
@@ -57,9 +57,26 @@ export const PanelEntryDropZonePlaceholder: React.FC<{
...
-        {label && (
-          <div className="dnd__entry-dropzone__placeholder__label">{label}</div>
-        )}
+        {label && <div> {label}</div>}
+      </div>
+    </div>
+  );
+};
+
+export const PanelEntryDropZonePlaceholderWithoutBorder: React.FC<{
```
akphi, line 66: why don't we make this a configuration instead? Or again, let people pass in some `children` should they want to customize

packages/legend-application-studio/src/components/editor/edit-panel/mapping-editor/PurePropertyMappingEditor.tsx
```diff
@@ -67,7 +67,9 @@ const SimplePropertyMappingEditor = observer(
     const propertyMapping = propertyMappingState.propertyMapping;
     const expectedType =
       propertyMapping.property.value.genericType.value.rawType;
-    const canDrop = dragItem && dragItem.data.type === expectedType;
+
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
```
akphi, line 71: please do not do this.

packages/legend-application-studio/src/components/editor/edit-panel/mapping-editor/PurePropertyMappingEditor.tsx
```diff
@@ -271,7 +273,10 @@ const ClassPropertyMappingEditor = observer(
       }
     };
     // Drag and Drop
-    const canDrop = dragItem?.data.type && dragItem.data.type === expectedType;
+
+    const canDrop =
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+      dragItem?.data?.type && dragItem.data.type === expectedType;
```
akphi, line 279: Please fix the problem, don't do this!

akphi (changes requested):
Please rebase. Also, I haven't reviewed `EditPanel` and `MappingEditor` in detail, but could we think of a way to componentize this?

packages/legend-application-studio/src/stores/EditorStore.ts
xannem: oh no not the most common source of memory leak 😓 understandable, removed!

MauricioUyaguari:
loving the enhancements!!!!

packages/legend-art/style/base/_common.scss
```diff
@@ -105,6 +105,17 @@
     width: 100%;
     margin: 0;
     font-weight: 700;
+
+    &--borderless {
+      @include flexVCenter;
+
+      border: none;
```
akphi, line 112: No need to repeat already specified attributes

packages/legend-art/src/components/panel/PanelDropZone.tsx
```diff
@@ -49,17 +49,22 @@ export const PanelEntryDropZonePlaceholder: React.FC<{
...
   return (
     <div className={clsx(['dnd__entry-dropzone__placeholder', className])}>
-      <div className="dnd__entry-dropzone__placeholder__content">
-        {label && (
-          <div className="dnd__entry-dropzone__placeholder__label">{label}</div>
-        )}
+      <div
+        className={
+          borderless
+            ? 'dnd__entry-dropzone__placeholder__content--borderless'
```
akphi, line 63: why don't we use `clsx` here?

packages/legend-art/src/components/panel/PanelDropZone.tsx
```diff
@@ -49,17 +49,22 @@ export const PanelEntryDropZonePlaceholder: React.FC<{
   showPlaceholder: boolean;
   className?: string;
   label?: string;
+  borderless?: boolean;
 }> = (props) => {
-  const { children, showPlaceholder, className, label } = props;
+  const { children, showPlaceholder, className, borderless, label } = props;
```
akphi, line 54: seems concerning to me, like always, is there a way to customize by allowing users to pass in some component for the placeholder they wish? if you don't like the placeholder, either use `showPlaceholder=false` or pass in that `placeholder` renderer, what do you think? We need to refactor a bit here though since now the relationship between `label`, `showPlaceholder` and `placeholder` seem messy

akphi (changes requested):
This is unfortunately not a trivial change so I'll have to review them in phases. Anyhow, what do you think about componentizing this? does it mean much?

I would like us to also do something similar to `Pure IDE` or Chrome 

image

eventually where we have a dropdown button on the bar to show all open tabs and allow to search them.

I see you do something in `EditorStore` so that's the non-trivial bit that I need to take a better look at.

packages/legend-art/src/components/panel/PanelDropZone.tsx
xannem: passing in placeholder renderer with the label for that one, maybe makes it clearer? 🤔

akphi:
Closing in favor of https://github.com/finos/legend-studio/pull/1711

PR #1457, opened 2022-09-05
query builder: rework query builder setup using workflows

akphi:
## Summary

- [x] docs: add guide for cross-project setup
- [x] bump dependencies
- [x] Have a different set of methods in `PureModel` for `get databases()` and etc. so we can safely exclude generated elements or unexposed system elements
- [x] Refactor `QueryBuilderState` core
  - [x] Get rid of `QueryBuilderMode`
  - [x] Move milestoning logic out of `QueryBuilderSetupState`
  - [x] Reorganize codebase: all core `query-builder` logic will be moved to `(components|stores)/query-builder`
  - [x] Cleanup `QueryEditorStore`: renaming stores and change `QueryEditorStore.queryBuilderState` to optional and use that as indicator if the `QueryEditorStore` has been properly initialized or not
  - [x] **BREAKING CHANGE:** Update `URL pattern` for edit page:
    - [x] `/query/dataspace` -> `/query/create/create-from-dataspace/`
    - [x] `/query/service` -> `/query/create/create-from-service/`
    - [x] `/query/create` -> `/query/create/manual/`
  - [x] Make `QueryBuilderSetupState` abstract: Make `Class, Mapping, Runtime` the basis of `QueryBuilderSetupState`, refactor this state so that we could make it abstract, as long as we could (at any point in time) answer the question `what is the class, mapping, and runtime?`
    - [x] Flatten `QueryBuilderSetupState` and inline it into `QueryBuilderState`
    - [x] Make `changeClass` to no longer allow `undefined`, remove the `isRebuildingState` flag
    - [x] Componentize `class selector with milestoning`: `QueryBuilderClassSelector`
    - [x] Generalize the current setup panel: it will now just allow setting `class`, `mapping`, `runtime` without any cascading, i.e. `BasicQueryBuilderState`
    - [x] Fix a problem with going from text to form mode when the query is `unsupported` duplicates the number of query parameters
- [x] Create different workflows for `query builder`
  - [x] Closes #319 
  - [x] Closes #1300 

### Introduce workflows to query builder
| |QueryBuilderState implementation|Usage                                                         |Setup Cascading Ordering|
|------|--------------------------------|--------------------------------------------------------------|------------------------|
|   <ul><li>[x] </li></ul>   |MappingQueryBuilderState        |create a manual query                                                              |MRC (1)                 |
|   <ul><li>[x] </li></ul>    |ServiceQueryBuilderState (0)     |load service query, (studio) edit query                       |EcC/MRC (2)             |
[...]

PR #1489, opened 2022-09-19
Fix regression introduced for Parameter panel in #1457

gayathrir11:
## Summary

We don't show the option to add parameters in parameter panel eventhough parameters are supported via that workflow

- [ ] Test(s) added

Capture

### DIFF
--- a/packages/legend-query-builder/src/components/QueryBuilderParametersPanel.tsx
+++ b/packages/legend-query-builder/src/components/QueryBuilderParametersPanel.tsx
@@ -370,7 +370,7 @@ export const QueryBuilderParametersPanel = observer(
           <div className="panel__header__title">
             <div className="panel__header__title__label">parameters</div>
           </div>
-          {!isReadOnly && queryBuilderState.isParameterSupportDisabled && (
+          {!isReadOnly && !queryBuilderState.isParameterSupportDisabled && (
             <div className="panel__header__actions">
               <button
                 className="panel__header__action"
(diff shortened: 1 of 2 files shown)


packages/legend-query-builder/src/components/QueryBuilderParametersPanel.tsx
```diff
@@ -370,7 +370,7 @@ export const QueryBuilderParametersPanel = observer(
           <div className="panel__header__title">
             <div className="panel__header__title__label">parameters</div>
           </div>
-          {!isReadOnly && queryBuilderState.isParameterSupportDisabled && (
+          {!isReadOnly && !queryBuilderState.isParameterSupportDisabled && (
```
gayathrir11, line 373: Do we need to have this check `!isReadOnly && !queryBuilderState.isParameterSupportDisabled` again in `addParameter` function because we already have made similar checks when we are using the function

packages/legend-query-builder/src/components/QueryBuilderParametersPanel.tsx
```diff
@@ -370,7 +370,7 @@ export const QueryBuilderParametersPanel = observer(
           <div className="panel__header__title">
             <div className="panel__header__title__label">parameters</div>
           </div>
-          {!isReadOnly && queryBuilderState.isParameterSupportDisabled && (
```
akphi, line 373: image

packages/legend-query-builder/src/components/QueryBuilderParametersPanel.tsx
akphi: hm, technically no, but I prefer to be defensive here.

akphi (approved):

PR #1500, opened 2022-09-22
minor add label for post processor selector panel

xannem:
## Summary

A small addition to add a label to the type of post processor on the viewer so that differentiation is easier for users 

image

- [ ] Test(s) added

### DIFF
--- a/packages/legend-art/src/components/panel/Panel.tsx
+++ b/packages/legend-art/src/components/panel/Panel.tsx
@@ -56,6 +56,18 @@ export const PanelHeader: React.FC<{
   );
 };
 
+export const Badge: React.FC<{
+  title: string;
+  className?: string;
+}> = (props) => {
+  const { title, className } = props;
+  return (
+    <div className={clsx('panel__header__badge__label', className)}>
+      {title.toLowerCase()}
+    </div>
+  );
+};
+
 export const PanelTabs: React.FC<{
   tabTitles: string[];
   changeTheTab: <T>(
--- a/packages/legend-art/src/components/panel/Panel.tsx
+++ b/packages/legend-art/src/components/panel/Panel.tsx
@@ -56,18 +56,6 @@ export const PanelHeader: React.FC<{
   );
 };
 
-export const Badge: React.FC<{
-  title: string;
-  className?: string;
-}> = (props) => {
-  const { title, className } = props;
-  return (
-    <div className={clsx('panel__header__badge__label', className)}>
-      {title.toLowerCase()}
-    </div>
-  );
-};
-
 export const PanelTabs: React.FC<{
   tabTitles: string[];
   changeTheTab: <T>(
--- a/packages/legend-art/src/components/panel/PanelListSelectorItem.tsx
+++ b/packages/legend-art/src/components/panel/PanelListSelectorItem.tsx
@@ -17,11 +17,13 @@ import { clsx } from 'clsx';
 
 export const PanelListSelectorItem: React.FC<{
   title?: string;
+  badgeTitle?: string | undefined;
   onSelect: () => void;
   isSelected: boolean;
   validationErrorMessage?: string | undefined;
 }> = (props) => {
-  const { title, onSelect, isSelected, validationErrorMessage } = props;
+  const { title, badgeTitle, onSelect, isSelected, validationErrorMessage } =
+    props;
   return (
     <div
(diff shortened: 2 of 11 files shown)


akphi:
@xannem let's add the badge after (😢) and let's make it smaller and more subtle so it doesn't visually clash with the panel header badge 🙏

akphi:
And you got the #1500 🎉 🎉 🌮

akphi (changes requested):

packages/legend-art/style/base/_panel.scss
```diff
@@ -183,13 +183,13 @@
   &__item {
     display: flex;
     align-items: center;
-    justify-content: space-between;
```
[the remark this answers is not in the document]
xannem: I switched to children allowable in panel list selector component but I was thinking maybe in general even with whatever they want to add, maybe the default aligning left looks better? 

image

image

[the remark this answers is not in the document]
akphi: perhaps, but for the latter, you need to account for text-overflow and eclipses for the title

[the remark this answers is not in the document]
xannem: 🤔 makes sense, justify-content will remain justified 👍

packages/legend-art/src/components/panel/Panel.tsx
```diff
@@ -56,6 +56,18 @@ export const PanelHeader: React.FC<{
   );
 };
 
+export const Badge: React.FC<{
```
akphi, line 59: It's fine for now, but let's have this in its own separate file and separate the styling when possible! Or else, we call this `PanelHeaderBadge` or something, not sure :D

packages/legend-art/src/components/panel/PanelListSelectorItem.tsx
```diff
@@ -17,11 +17,13 @@ import { clsx } from 'clsx';
 
 export const PanelListSelectorItem: React.FC<{
   title?: string;
+  children?: React.ReactNode;
```
akphi, line 20: When you design a component, it's weird to do thing like `title? children?` we need to handle the logic where both is provided, which one will take precedence. I think you misunderstood my comment earlier as well, see my comment about usage of `{children}` below

packages/legend-art/src/components/panel/PanelListSelectorItem.tsx
```diff
@@ -50,6 +52,7 @@ export const PanelListSelectorItem: React.FC<{
       >
         {title}
       </div>
+      {children}
```
akphi, line 55: This looks strange, `children` is the ultimate escape hatch when people want to fully customize the list item, as such, if `chilren` is provided, we won't need `title` and vice-versa. What choice do you make here?

packages/legend-art/src/components/panel/PanelListSelectorItem.tsx
akphi: A neat way is to create another component

`PanelListSelectorItemLabel` which takes just the title and can be used as child of `<PanelListSelectorItem>` which takes a `children` prop only.

packages/legend-art/style/base/_panel.scss
```diff
@@ -183,13 +195,14 @@
...
     height: 2.4rem;
```
akphi, line 198: let's revert this unecessary change

packages/legend-art/style/base/_panel.scss
```diff
@@ -77,6 +77,18 @@
     @include ellipsisTextOverflow;
   }
 
+  &__header__badge__label {
```
akphi, line 80: is the badge here pertained to only the context of panel? or is this generic badge which we could separate from panel?

akphi (changes requested):

packages/legend-art/src/components/panel/Panel.tsx
xannem: I think it can stand on its own tho it looks kinda weak! Separated file 👍

packages/legend-art/src/components/panel/PanelListSelectorItem.tsx
xannem: true!

akphi (approved):

Issue #1309, opened 2022-07-15
Bug: Fix Executing/Testing Milestone Classes in Mapping Editor


MauricioUyaguari:
### Similar requests

- [X] I have searched and found no existing similar requests

### What problems are you trying to solve?

Opening query builder on milestone class via the class editor brings us to an unsupported state in query builder. 
This is due to no support for parameters in mapping test/execute. 

https://user-images.githubusercontent.com/29690824/179244757-696741a1-c811-4eb6-8bd0-08a8a05d2cb2.mp4

### Possible Solutions
We may want to generate a .all with a hardcoded date for the milestone class. 

### Grammar

```
Class <<meta::pure::profiles::temporal.businesstemporal>> model::MilstoneClass
{
  id: String[1];
  name: String[1];
}

Class model::Source
{
  id: String[1];
}


###Mapping
Mapping model::M2MMapping
(
  *model::MilstoneClass: Pure
  {
    ~src model::Source
    id: $src.id,
    name: $src.id
  }
)

```

Issue #1149, opened 2022-05-09
Feature request: Milestoning (last mile)


gayathrir11:
### Similar requests

- [X] I have searched and found no existing similar requests

### How are you using Studio?

Legend Query

### What problems are you trying to solve?

Improve the user experience for milestoning in query builder.

### Describe the solution you would like to see implemented

- [ ] Milestoning date-propagation support in `graphFetch` - blocked by https://github.com/finos/legend-studio/issues/1188
- [ ] Add integration tests to cover all the edge cases for milestoning, we should revisit:
    https://github.com/finos/legend-studio/pull/1134
    https://github.com/finos/legend-studio/pull/1128
    https://github.com/finos/legend-studio/pull/891
- [ ] Improve stability, handle bugs:
  - [ ] `BUG` When using `Studio`, we need to check and potentially generate `milestoned` versions for new properties being added to the graph
  - [ ] https://github.com/finos/legend-studio/issues/959
  - [x] https://github.com/finos/legend-studio/issues/1148
  - [ ] https://github.com/finos/legend-studio/issues/1471
  - [ ] https://github.com/finos/legend-studio/issues/1309 
  - [ ] Support `allVersions` and `allVersionsInRange` for properties
- [ ] Minor UX improvements:
  - [ ] Show a clock icon in front of the property expression badge if people set a custom date: this is a simplification of #959 
  - [ ] When they first switch to a new class, we might want to show a popup, or when the `$businessDate` and `$processingDate` parameters being added for the first time, we want to notify users about that
  - [ ] Handle the edge case where the variable `$businessDate` and `$processingDate` already existed, what should we do here? Should we just hardcode a date in the query and notify (blocking alert) about that? as we cannot really add the milestoning date parameters?
  - [ ] If we found out some parameters have been used in `QueryBuilderMilestoningState`, we will show some special indicator/help icon around the 2 special parameters

### Contribution

- [x] I would like to work on this feature

PR #1507, opened 2022-09-28
Add basic tests for milestoning to accomodate testing via workflows

gayathrir11:
## Summary

- [x] Add basic tests for milestoning which tests the loading of milestoned classes via workflows
- [x] Fix a bug introduced for loading a milestoned class from explorer tree in #1457 

- [x] Test(s) added

https://user-images.githubusercontent.com/87973126/192771269-62ebf238-3383-4bbc-81ec-840c33b5dd96.mp4

### DIFF
--- a/packages/legend-application-studio/src/components/__tests__/EmbeddedQueryBuilder.test.tsx
+++ b/packages/legend-application-studio/src/components/__tests__/EmbeddedQueryBuilder.test.tsx
@@ -38,7 +38,7 @@ import { LEGEND_STUDIO_TEST_ID } from '../LegendStudioTestID.js';
 import { TEST__getLegendStudioApplicationConfig } from '../../stores/EditorStoreTestUtils.js';
 import type { EditorStore } from '../../stores/EditorStore.js';
 
-const TEST__buildQueryBuilderMockedEditorStore = (): EditorStore => {
+export const TEST__buildQueryBuilderMockedEditorStore = (): EditorStore => {
   const pluginManager = LegendStudioPluginManager.create();
   pluginManager.usePresets([new QueryBuilder_GraphManagerPreset()]).install();
 
--- a/packages/legend-query-builder/src/components/__tests__/QueryBuilder_Milestoning.test.tsx
+++ b/packages/legend-query-builder/src/components/__tests__/QueryBuilder_Milestoning.test.tsx
@@ -17,6 +17,10 @@
 import { describe, test, expect } from '@jest/globals';
 import { act } from '@testing-library/react';
 import {
+  TEST_DATA__getAllWithHardcodedDateInput,
+  TEST_DATA__getAllWithHardcodedDateOutput,
+  TEST_DATA__simpleProjectionWithAggregationInput,
+  TEST_DATA__simpleProjectionWithAggregationOutput,
   TEST_DATA__simpleProjectionWithBiTemporalSourceAndBiTemporalTarget,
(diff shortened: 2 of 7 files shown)


gayathrir11:
@akphi  please let me know if i missed adding tests for any workflow. From what I understand `initializeWithQuery` and `queryClass` are two potential ways to create a query. I haven't added tests for `MappingExecution` flow and `MappingTest` flow as parameters are not supported over there we don't do any milestoning transformations for class when we load it over there. Will add those tests when we address this #1309 
I tested this by making changes to the code so that these tests fail when we load classes incorrectly. It was catching that.

packages/legend-query-builder/src/stores/QueryBuilderState.ts
```diff
@@ -253,6 +253,7 @@ export abstract class QueryBuilderState {
   changeMapping(val: Mapping): void {
     this.resetQueryResult();
     this.resetQueryContent();
+    this.milestoningState.updateMilestoningConfiguration();
```
akphi, line 256: I will move this statement one line down, but I totally missed the fact that we need to call this again post calling `resetQueryContent`. Thanks!

packages/legend-query-builder/src/components/__tests__/QueryBuilder_Milestoning.test.tsx
```diff
@@ -52,6 +52,7 @@ type TestCase = [
     rawLambda: { parameters?: object; body?: object };
```
akphi, line 52: This is just a comment, but this kind of test unnerves me, it's strange because we're just checking the number of arguments provided. Anyhow, let's hope when we do #1149 we can reshape this test to something that is more solid 🤞

packages/legend-application-studio/src/components/__tests__/EmbeddedQueryBuilder.test.tsx
```diff
@@ -253,6 +429,175 @@ test(integrationTest('Open query builder by querying a class'), async () => {
   );
 });
 
+test(
```
akphi, line 432: Let's separate these tests to another file, since these focus one particular workflow, feel free to name it something like `ClassQueryBuilder.test.tsx`, the test data can be moved to its own file. Also, these tests can be cleaned up to use `TestCase` I believe - that's what I meant by making these tests scale better

akphi (changes requested):

packages/legend-application-studio/src/components/__tests__/EmbeddedQueryBuilder.test.tsx
```diff
@@ -253,6 +429,175 @@ test(integrationTest('Open query builder by querying a class'), async () => {
...
+  integrationTest(
+    'Open query builder by querying a processing temporal milestoned class',
+  ),
```
MauricioUyaguari, line 435: Do we have a test for the most recent bug we found where date propagation is not supported for groupBy Functions. 
In this case we are expecting query builder not to force propagate the date. 
We should also have tests when this behavior is expected. 
Right now if the user opens a query that has dates defined but are expected to be propagated we rewrite this, we should add tests for those on the expected input vs output.

gayathrir11: yeah i think expected input vs output tests would make much sense for milestoning. Thanks for the idea

akphi (approved):
Looking great! Thanks so much for this 🙌

PR #1528, opened 2022-10-04
fix: visual indicator for system elements

xannem:
## Summary
Pointed out by @Yasirmod17, when a user selects a source element from the mapping editor, there are no color indicators for what type it is. Therefore, adding `getPackageableElementOptionFormatter` to that selector input as well. 

For example, the elements below should have the lime green color designated for dependencies 
image

- [ ] Test(s) added

### DIFF
--- a/packages/legend-application-studio/src/components/editor/edit-panel/mapping-editor/InstanceSetImplementationSourceSelectorModal.tsx
+++ b/packages/legend-application-studio/src/components/editor/edit-panel/mapping-editor/InstanceSetImplementationSourceSelectorModal.tsx
@@ -45,6 +45,7 @@ import { useEditorStore } from '../../EditorStoreProvider.js';
 import {
   useApplicationStore,
   buildElementOption,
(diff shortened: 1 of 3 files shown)


akphi (approved):

packages/legend-application-studio/src/components/editor/edit-panel/mapping-editor/InstanceSetImplementationSourceSelectorModal.tsx
```diff
@@ -205,6 +206,7 @@ export const InstanceSetImplementationSourceSelectorModal = observer(
             placeholder={`Select a source...`}
             isClearable={true}
             filterOption={filterOption}
+            formatOptionLabel={getPackageableElementOptionFormatter({})}
```
akphi, line 209: we need to modify this slightly. This won't work well afaik because instance set implementation source selector allows more stuffs than just elements

PR #1543, opened 2022-10-07
minor improvements

YannanGao-gs:
## Summary
Fixes https://github.com/finos/legend-studio/issues/1542

## How did you test this change?

- [ ] Test(s) added
- [ ] Manual testing (please provide screenshots/recordings)
- [x] No testing (please provide an explanation)

PR #1553, opened 2022-10-12
fix a regression introduced in #1543

YannanGao-gs:
## Summary
**BREAKING CHANGE:** Make multiplicity's upper bound and lower bond `readonly`.

- [ ] Test(s) added

### DIFF
--- a/.changeset/slimy-feet-fly.md
+++ b/.changeset/slimy-feet-fly.md
@@ -1,0 +1,5 @@
+---
+'@finos/legend-graph': major
+---
+
+**BREAKING CHANGE:** Make multiplicity's upper bound and lower bound `readonly`.
--- a/.changeset/swift-lies-speak.md
+++ b/.changeset/swift-lies-speak.md
@@ -1,0 +1,5 @@
+---
+'@finos/legend-query-builder': patch
+---
+
+Fix a regression introduced in ([#1543](https://github.com/finos/legend-studio/pull/1543)): Avoid updating multiplicity in place, create a new instance instead.
--- a/packages/legend-graph/src/graph/metamodel/pure/packageableElements/domain/Multiplicity.ts
+++ b/packages/legend-graph/src/graph/metamodel/pure/packageableElements/domain/Multiplicity.ts
@@ -18,8 +18,8 @@ import { CORE_HASH_STRUCTURE } from '../../../../../graph/Core_HashUtils.js';
 import { hashArray, type Hashable } from '@finos/legend-shared';
 
 export class Multiplicity implements Hashable {
-  lowerBound: number;
-  upperBound?: number | undefined;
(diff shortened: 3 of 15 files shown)


.changeset/rare-mirrors-wonder.md
```diff
@@ -0,0 +1,5 @@
+---
+'@finos/legend-graph': patch
```
akphi, line 2:
```suggestion
'@finos/legend-graph': major
```

.changeset/rare-mirrors-wonder.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Make multiplicity's upperbond and lowerbond readonly.
```
akphi, line 5:
```suggestion
**BREAKING CHANGE:** Make multiplicity's upper bound and lower bond `readonly`.
```

.changeset/swift-lies-speak.md
```diff
@@ -0,0 +1,5 @@
+---
+'@finos/legend-query-builder': patch
+---
+
+Fix a regression introduced in ([#1543](https://github.com/finos/legend-studio/pull/1543)).
```
akphi, line 5: Can we briefly mention what this regression is about so people can figure it out from the changelog rather than having to go to the link?

packages/legend-graph/src/graph/metamodel/pure/packageableElements/domain/Multiplicity.ts
```diff
@@ -18,8 +18,8 @@ import { CORE_HASH_STRUCTURE } from '../../../../../graph/Core_HashUtils.js';
 import { hashArray, type Hashable } from '@finos/legend-shared';
 
 export class Multiplicity implements Hashable {
-  lowerBound: number;
-  upperBound?: number | undefined;
+  readonly lowerBound: number;
```
akphi, line 21: did you check the `...Observer` stuff to see if we decorate these as `:observable`? since `readonly` fields should not be `observable`, hence the `hashCode` for multiplicity probably don't need to be `computed` as well

akphi (changes requested):

.changeset/slimy-feet-fly.md
```diff
@@ -0,0 +1,7 @@
+---
+'@finos/legend-application-studio': major
+'@finos/legend-extension-store-service-store': major
+'@finos/legend-graph': major
+---
+
+**BREAKING CHANGE:** Make multiplicity's upper bound and lower bond `readonly`.
```
akphi, line 7: This major changeset should only be applied to one of the module.

akphi (changes requested):
Looks good to me, but you would need to update the changeset. Thanks!

PR #1610, opened 2022-11-01
fix: remove display of dependency problems

xannem:
## Summary

When compiling projects with dependencies, legend engine can return warnings from them and this can easily lead to thousands of errors being rendered on the page. These warnings are a bit extraneous and cause unnecessary clutter, so this change chooses to not display those errors.

Example of project warning panel with dependencies: 
image
 

- [ ] Test(s) added

### DIFF
--- a/packages/legend-application-studio/src/stores/EditorGraphState.ts
+++ b/packages/legend-application-studio/src/stores/EditorGraphState.ts
@@ -186,7 +186,12 @@ export class EditorGraphState {
     return [this.error, ...this.warnings].filter(isNonNullable);
   }
 
-  //TODO: remove once engine changes are added with better source information
+  /**
+   * This function is temporary. There is no good way to detect if a problem not coming from
+   * the main graph at the moment. In text mode, we can rely on the fact that the source information
+   * has line 0 column 0. But this is not the case for form mode, so this is just temporary
+   * to help with text-mode.
+   */
   TEMPORARY__removeDependencyProblems(
     problems: Problem[] | CompilationWarning[],
   ): Problem[] | CompilationWarning[] {
--- a/packages/legend-application-studio/src/stores/EditorGraphState.ts
+++ b/packages/legend-application-studio/src/stores/EditorGraphState.ts
@@ -186,6 +186,23 @@ export class EditorGraphState {
     return [this.error, ...this.warnings].filter(isNonNullable);
   }
 
+  //TODO: remove once engine changes are added with better source information
+  TEMPORARY__removeDependencyProblems(
+    problems: Problem[] | CompilationWarning[],
+  ): Problem[] | CompilationWarning[] {
+    return problems.filter((problem) => {
+      if (problem.sourceInformation) {
+        return !(
+          problem.sourceInformation.startLine === 0 &&
+          problem.sourceInformation.startColumn === 0 &&
+          problem.sourceInformation.endLine === 0 &&
+          problem.sourceInformation.endColumn === 0
+        );
+      }
+      return true;
+    });
+  }
+
   get areProblemsStale(): boolean {
     return (
       (this.editorStore.isInFormMode &&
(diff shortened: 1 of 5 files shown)


packages/legend-application-studio/src/stores/EditorGraphState.ts
```diff
@@ -186,6 +186,22 @@ export class EditorGraphState {
     return [this.error, ...this.warnings].filter(isNonNullable);
   }
 
+  removeDependencyProblems(
+    problems: Problem[] | CompilationWarning[],
+  ): Problem[] | CompilationWarning[] {
+    return problems.filter((problem) => {
+      if (problem.sourceInformation) {
```
akphi, line 193: I don't see how this filtering work in form mode where the problem probably doesn't have any source information at all

packages/legend-application-studio/src/components/editor/aux-panel/Problems.tsx
```diff
@@ -75,7 +75,10 @@ const ProblemItem = observer((props: { problem: Problem }) => {
 
 export const Problems = observer(() => {
   const editorStore = useEditorStore();
-  const problems = editorStore.graphState.problems;
+  const problems = editorStore.graphState.removeDependencyProblems(
```
akphi, line 78: I think this is unnecessary, we already handled this in `EditorGraphState`

packages/legend-application-studio/src/components/editor/aux-panel/AuxiliaryPanel.tsx
```diff
@@ -70,7 +70,9 @@ export const AuxiliaryPanel = observer(() => {
       name: 'PROBLEMS',
       icon: undefined,
       isVisible: true,
-      counter: editorStore.graphState.problems.length,
+      counter: editorStore.graphState.removeDependencyProblems(
```
akphi, line 73: this is redundant

packages/legend-application-studio/src/components/editor/StatusBar.tsx
```diff
@@ -99,7 +99,9 @@ export const StatusBar = observer((props: { actionsDisabled: boolean }) => {
 
   // Problems
   const error = editorStore.graphState.error;
-  const warnings = editorStore.graphState.warnings;
+  const warnings = editorStore.graphState.removeDependencyProblems(
```
akphi, line 102: this is redundant, we can do this in `EditorGraphState`

akphi (changes requested):

packages/legend-application-studio/src/components/editor/StatusBar.tsx
xannem: true!

packages/legend-application-studio/src/stores/EditorGraphState.ts
xannem: 😔

packages/legend-application-studio/src/stores/EditorGraphState.ts
```diff
@@ -186,6 +186,22 @@ export class EditorGraphState {
     return [this.error, ...this.warnings].filter(isNonNullable);
   }
 
+  removeDependencyProblems(
```
akphi, line 189: Let's add a comment here explaining the limitation of this approach

packages/legend-application-studio/src/stores/EditorGraphState.ts
```diff
@@ -186,6 +186,23 @@ export class EditorGraphState {
     return [this.error, ...this.warnings].filter(isNonNullable);
   }
 
+  //TODO: remove once engine changes are added with better source information
```
akphi, line 189:
```suggestion
/**
 * This function is temporary. There is no good way to detect if a problem not coming from
 * the main graph at the moment. In text mode, we can rely on the fact that the source information
 * has line 0 column 0. But this is not the case for form mode, so this is just temporary
 * to help with text-mode.
 */
```

akphi (approved):

PR #1611, opened 2022-11-01
fix: select all properties from root class in explorer

### LINKED ISSUE #1608: Bug: add all properties to fetch structure on root class does not work
### Similar issues

### How are you using Studio?

Legend Query

Adding all properties does not work on the root class. Works on properties under main class but not root class. 

xannem:
fixes #1608 

## Summary

Kapture 2022-11-01 at 16 09 27

- [ ] Test(s) added

### DIFF
--- a/.changeset/eight-ways-brake.md
+++ b/.changeset/eight-ways-brake.md
@@ -1,0 +1,3 @@
+---
+'@finos/legend-query-builder': patch
+---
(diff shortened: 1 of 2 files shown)


MauricioUyaguari (approved):

MauricioUyaguari:
Thanks @xannem  could we think of ways to add a test for this. This is the second time this bug has come up. Maybe assert when we right click we see the option

xannem:
@MauricioUyaguari A test PR #1613 🙇

PR #1682, opened 2022-11-11
fix: keeping query states when switching execution contexts

xannem:
## Summary
Currently when users switch execution environments in query builder, things present in their query will be reset. We would like to keep those changes when switching. 

Kapture 2022-11-11 at 11 54 12

- [ ] Test(s) added

### DIFF
--- a/.changeset/swift-rockets-draw.md
+++ b/.changeset/swift-rockets-draw.md
@@ -2,4 +2,4 @@
 '@finos/legend-query-builder': patch
 ---
 
-fix: keeping query states when switching execution contexts
+fix: keeping query states when switching service execution contexts
--- a/.changeset/swift-rockets-draw.md
+++ b/.changeset/swift-rockets-draw.md
@@ -1,0 +1,5 @@
+---
+'@finos/legend-query-builder': patch
+---
+
+fix: keeping query states when switching execution contexts
--- a/packages/legend-query-builder/src/stores/QueryBuilderState.ts
+++ b/packages/legend-query-builder/src/stores/QueryBuilderState.ts
@@ -284,9 +284,9 @@ export abstract class QueryBuilderState implements CommandRegistrar {
     this.milestoningState.updateMilestoningConfiguration();
   }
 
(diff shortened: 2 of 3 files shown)


packages/legend-query-builder/src/stores/QueryBuilderState.ts
```diff
@@ -284,9 +284,9 @@ export abstract class QueryBuilderState implements CommandRegistrar {
     this.milestoningState.updateMilestoningConfiguration();
   }
 
-  changeMapping(val: Mapping): void {
+  changeMapping(val: Mapping, keepQueryContent?: boolean): void {
```
MauricioUyaguari, line 287: we could maybe add a parameter called options? : { keepQueryContent?: boolean}

we should also disable updating milestone config if this flag is set.

.changeset/swift-rockets-draw.md
```diff
@@ -0,0 +1,5 @@
+---
+'@finos/legend-query-builder': patch
+---
+
+fix: keeping query states when switching execution contexts
```
MauricioUyaguari, line 5: fix: keeping query states when switching service execution contexts

MauricioUyaguari (changes requested):

packages/legend-query-builder/src/stores/QueryBuilderState.ts
```diff
@@ -284,10 +284,12 @@ export abstract class QueryBuilderState implements CommandRegistrar {
     this.milestoningState.updateMilestoningConfiguration();
   }
 
-  changeMapping(val: Mapping): void {
+  changeMapping(val: Mapping, options?: { keepQueryContent?: boolean }): void {
     this.resetQueryResult();
-    this.resetQueryContent();
-    this.milestoningState.updateMilestoningConfiguration();
+    if (!options || options.keepQueryContent === false) {
```
MauricioUyaguari, line 289: `!options?.keepQueryContent`

MauricioUyaguari (approved):

PR #1693, opened 2022-11-16
Remove graph building form text mode

### LINKED ISSUE #966: Enhancement: remove graph building in text mode to improve performance
c### Similar requests

### How are you using Studio?

Legend Studio

In order to improve performance in text mode, we could consider removing `Studio` graph building during text mode. 

Currently, we rebuild our graph during each compilation. We will move to rebuild only when leaving text mode. 

- [ ] remove logic of rebuilding when firing `F9` in text-mode
- [ ] handle explorer tree logic which requires graph: we can reason out the explorer tree from just the element paths
- [ ] handle change detection (show that we disabled change detection in text mode and prevent users from saving, or we could use another change detection mode - `maybe` compare hashes index built from protocol models)
- [ ] Move from a soft transition from/to text mode to a hard transition. Right now we try to keep the form mode state of studio in memory when going to text mode. we should delete that and only rebuild when going back to form mode - we currently already somewhat doing this, but we haven't pushed this hard enough, this would really help avoiding memory-leak: genericize `EditorStore` and implement `TextEditorStore`, `FormEditorStore` to further separate these out (See #317 )
  - [ ] `?` would remove the `reprocessElementEditorState`

_No response_

_No response_

### Contribution

gayathrir11:
## Summary

Closes #966 

- [ ] Test(s) added

### DIFF
--- a/packages/legend-application-studio/src/components/editor/side-bar/LocalChanges.tsx
+++ b/packages/legend-application-studio/src/components/editor/side-bar/LocalChanges.tsx
@@ -51,6 +51,7 @@ import { useApplicationStore } from '@finos/legend-application';
 import { useEffect } from 'react';
 import { EntityChangeConflictEditorState } from '../../../stores/editor-state/entity-diff-editor-state/EntityChangeConflictEditorState.js';
 import { EntityChangeConflictSideBarItem } from '../edit-panel/diff-editor/EntityChangeConflictEditor.js';
+import { FormLocalChangesState } from '../../../stores/sidebar-state/LocalChangesState.js';
 
 const PatchLoader = observer(() => {
   const editorStore = useEditorStore();
@@ -172,8 +173,11 @@ export const LocalChanges = observer(() => {
     editorStore.changeDetectionState.workspaceLocalLatestRevisionState.changes;
   const openChange =
     (diff: EntityDiff): (() => void) =>
-    (): void =>
-      localChangesState.openLocalChange(diff);
+    (): void => {
+      if (localChangesState instanceof FormLocalChangesState) {
+        localChangesState.openLocalChange(diff);
+      }
+    };
   // Local/Remote Workspace Conflicts
   const conflicts =
     editorStore.changeDetectionState.potentialWorkspacePullConflicts;
@@ -182,8 +186,11 @@ export const LocalChanges = observer(() => {
     conflict.entityPath === currentTabState.entityPath;
   const openPotentialConflict =
     (conflict: EntityChangeConflict): (() => void) =>
-    (): void =>
-      localChangesState.openPotentialWorkspacePullConflict(conflict);
+    (): void => {
+      if (localChangesState instanceof FormLocalChangesState) {
+        localChangesState.openPotentialWorkspacePullConflict(conflict);
+      }
+    };
   // Local/Remote Workspace Changes
   const workspacePullChanges =
     editorStore.changeDetectionState.aggregatedWorkspaceRemoteChanges;
@@ -195,8 +202,11 @@ export const LocalChanges = observer(() => {
   );
   const openWorkspacePullChange =
     (diff: EntityDiff): (() => void) =>
-    (): void =>
-      localChangesState.openWorkspacePullChange(diff);
+    (): void => {
+      if (localChangesState instanceof FormLocalChangesState) {
+        localChangesState.openWorkspacePullChange(diff);
+      }
+    };
 
   // check if workspace is still in-sync
   useEffect(() => {
--- a/packages/legend-application-studio/src/stores/EditorGraphState.ts
+++ b/packages/legend-application-studio/src/stores/EditorGraphState.ts
@@ -97,12 +97,15 @@ import {
 import {
   ActionAlertActionType,
   ActionAlertType,
+  type TabState,
 } from '@finos/legend-application';
 import { CONFIGURATION_EDITOR_TAB } from './editor-state/project-configuration-editor-state/ProjectConfigurationEditorState.js';
 import { graph_dispose } from './shared/modifier/GraphModifierHelper.js';
 import { PACKAGEABLE_ELEMENT_TYPE } from './shared/ModelClassifierUtils.js';
 import { GlobalTestRunnerState } from './sidebar-state/testable/GlobalTestRunnerState.js';
 import { LEGEND_STUDIO_APP_EVENT } from './LegendStudioAppEvent.js';
+import { ExplorerTreeState } from './ExplorerTreeState.js';
+import { FileGenerationViewerState } from './editor-state/FileGenerationViewerState.js';
 
 export enum GraphBuilderStatus {
   SUCCEEDED = 'SUCCEEDED',
@@ -146,7 +149,8 @@ export class EditorGraphState {
   constructor(editorStore: EditorStore) {
     makeObservable<
       EditorGraphState,
-      | 'updateGraphAndApplication'
+      | 'updateGraphAndApplicationInFormMode'
+      | 'updateGraphAndApplicationInTextMode'
       | 'mostRecentFormModeCompilationGraphHash'
       | 'mostRecentTextModeCompilationGraphHash'
     >(this, {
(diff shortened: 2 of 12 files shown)


MauricioUyaguari (changes requested):
half way through the changes will continue tomorrow

packages/legend-application-studio/src/stores/EditorStore.ts
```diff
@@ -1262,41 +1261,41 @@ export class EditorStore implements CommandRegistrar {
...
-  ): EditorState | undefined => {
-    if (editorState instanceof ElementEditorState) {
-      const correspondingElement =
-        this.graphManagerState.graph.getNullableElement(
-          editorState.element.path,
+  /**
+   * This method helps to open the editor states which were open we updated the application with the
+   * new graph
+   */
+  buildOpenedEditorStates = (
```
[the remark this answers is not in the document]
gayathrir11: You mean they have no editors opened when they go back to form mode from text?

packages/legend-application-studio/src/stores/EditorGraphState.ts
```diff
@@ -1106,26 +1202,20 @@ export class EditorGraphState {
...
-      this.editorStore.openedEditorStates = openedEditorStates
-        .map((editorState) =>
-          this.editorStore.reprocessElementEditorState(editorState),
-        )
-        .filter(isNonNullable);
-      this.editorStore.setCurrentEditorState(
-        this.editorStore.findCurrentEditorState(currentEditorState),
-      );
+      this.updateExplorerTree();
+      if (!inTextMode) {
```
MauricioUyaguari, line 1206: Will this change once we have a text for each element. 
We should just have a function called 
like buildOpenedElementEditors, which will have an implementation for text mode and form mode?

packages/legend-application-studio/src/stores/EditorStore.ts
```diff
@@ -1262,41 +1261,41 @@ export class EditorStore implements CommandRegistrar {
...
+    openedElementEditorPaths: string[],
+    otherOpenedEditorStates: FileGenerationViewerState[],
```
MauricioUyaguari, line 1270: Seems like `otherOpenedEditorStates` only refers to opened generated files so maybe `openedGeneratedFiles`

MauricioUyaguari (changes requested):

packages/legend-application-studio/src/stores/EditorGraphState.ts
```diff
@@ -943,12 +1009,33 @@ export class EditorGraphState {
...
 
       /**
-       * Backup and editor states info before resetting
-       *
-       * @risk memory-leak
+       * Backup and editor states info before resetting. Here we store the element paths of the
+       * elements editors as element paths don't refer to the actual graph. We can find the element
+       * from the new graph that is built by using element path and can reprocess the element editor states.
+       * The other kind of editors we reprocess are file generation editors, we store them as is as they don't
+       * hold any reference to the actual graph.
```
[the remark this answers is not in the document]
gayathrir11: that totally makes sense

packages/legend-application-studio/src/stores/EditorGraphState.ts
gayathrir11: yeah sure

packages/legend-application-studio/src/components/editor/side-bar/LocalChanges.tsx
```diff
@@ -172,8 +173,11 @@ export const LocalChanges = observer(() => {
     editorStore.changeDetectionState.workspaceLocalLatestRevisionState.changes;
   const openChange =
```
MauricioUyaguari, line 174: Here I think this is just because of current behaviour but i think now we can actually open a diff if they click on it and they are in text mode right? 
Either ways if openChange only works for form mode, we should think about disabling the button or some kind of UI indiactor?

packages/legend-application-studio/src/stores/EditorTabManagerState.ts
```diff
@@ -93,44 +91,64 @@ export class EditorTabManagerState extends TabManagerState {
...
-        let newTab: EditorState | undefined = undefined;
-        if (tab instanceof ElementEditorState) {
+  recoverTabs = (
+    openedTabEditorPaths: string[],
+    openedGeneratedFileTabStates: FileGenerationViewerState[],
+    currentTabState: EditorState | undefined,
+    currentTabElementPath: string | undefined,
+    shouldRecoverTabs: boolean,
+  ): void => {
+    const getElemenetPathFromFilePath = (filePath: string): string => {
```
MauricioUyaguari, line 102: I don't think `getElemenetPathFromFilePath` captures everything we need to confirm if that file is still being used. 
What is the risk of just not reprocessing those files ? have the user re generate if they want those files to re appear ?

MauricioUyaguari (approved):
still got 2/3 more files to go over but I think im okay to merge and start testing. i will leave more comments tomorrow and you can address them in the other text mode PR you have up. Thanks

packages/legend-application-studio/src/stores/EditorTabManagerState.ts
gayathrir11: there is no risk in reprocessing them according to me. They might just have to open them when they go in form mode.

packages/legend-application-studio/src/components/editor/side-bar/LocalChanges.tsx
gayathrir11: I don't think they can open diff editor when they are in text mode. It just doesn't open anything. We can diable the button. That would be a good option. This reminds me we need to disable `globalTestRunner` too. 
Will discuss more on that offline

packages/legend-application-studio/src/stores/EditorTabManagerState.ts
MauricioUyaguari: yeah i'd rather not than. i know we plan to do more in artifact generation extension and `getElemenetPathFromFilePath` dosesn't caputre everything

packages/legend-application-studio/src/stores/sidebar-state/LocalChangesState.ts
```diff
@@ -428,66 +499,6 @@ export class LocalChangesState {
...
-      this.editorStore.applicationStore.log.error(
-        LogEvent.create(LEGEND_STUDIO_APP_EVENT.SDLC_MANAGER_FAILURE),
-        error,
-      );
-    } finally {
-      this.refreshWorkspaceSyncStatusState.complete();
-    }
-  }
-
   *pushLocalChanges(pushMessage?: string): GeneratorFn<void> {
```
MauricioUyaguari, line 502: I still think we can clean this state a bit more. 
I see a lot of repeated logic in both of the pushLocalChanges such as after committing and checking with sdlc. 
Maybe abstract speicifc actions inside pushLocalChanges and have each state implement those. 
like if for conflict like the conflict logic could be.

`processConflicts` where the form mode handles them and text mode doesn't. etc.

packages/legend-application-studio/src/stores/sidebar-state/LocalChangesState.ts
```diff
@@ -693,6 +707,268 @@ export class LocalChangesState {
...
+
+  setLocalChanges(val: EntityChange[]): void {
+    this.localChanges = val;
+  }
+
+  *refreshLocalChanges(): GeneratorFn<void> {
+    const startTime = Date.now();
+    this.refreshLocalChangesDetectorState.inProgress();
+    try {
+      // ======= (RE)START CHANGE DETECTION =======
```
MauricioUyaguari, line 760: dont think we use the change detection state at all here so should we still have that comment here ?

packages/legend-application-studio/src/stores/EditorGraphState.ts
```diff
@@ -1065,6 +1152,177 @@ export class EditorGraphState {
...
+      this.editorStore.applicationStore.log.info(
+        LogEvent.create(GRAPH_MANAGER_EVENT.GRAPH_UPDATED_AND_REBUILT),
+        '[TOTAL]',
+        Date.now() - startTime,
+        'ms',
+      );
+      this.isUpdatingGraph = false;
+    } catch (error) {
+      assertErrorThrown(error);
+      this.editorStore.applicationStore.log.error(
```
MauricioUyaguari, line 1298: if weare truly decoupling the two methods can we look through this logic to ensure it still makes sense. 
I see `updateGraphAndApplicationInTextMode` but in here we have a check `this.editorStore.isInFormMode`. 
Does the catch logic of this still make sense in the text mode context ?

gayathrir11: It wouldn't. I am removing this

MauricioUyaguari (commented):
more comments

PR #1704, opened 2022-11-17
Fix clearing of compilation errors when we expand the lambda editor

### LINKED ISSUE #1703: Bug: Fix clearing of compilation errors when we expand the lambda editor
### Similar issues

### How are you using Studio?

Legend Studio

Fix clearing of compilation errors when we expand the lambda editor

### Steps to reproduce

Copy paste below grammar and open the derived property editor. Compile the code. It fails at compilation. You see the error in lambda editor but when you expand the lambad editor the error is cleared

### Model data

```text
Class test::class
{
  name: String[1];
  age: Number[1];
  derivedProp() {$this.name.path}: String[1];
}
```

### Environment

_No response_

_No response_

### Contribution

gayathrir11:
## Summary

closes #1703

- [ ] Test(s) added

https://user-images.githubusercontent.com/87973126/205231893-092e6cbe-9ab5-45f0-9a70-75a1a998d051.mp4

### DIFF
--- a/packages/legend-application-studio/src/stores/editor/editor-state/element-editor-state/ClassState.ts
+++ b/packages/legend-application-studio/src/stores/editor/editor-state/element-editor-state/ClassState.ts
@@ -104,7 +104,10 @@ export class DerivedPropertyState extends LambdaEditorState {
     }
   }
 
-  *convertLambdaObjectToGrammarString(pretty: boolean): GeneratorFn<void> {
+  *convertLambdaObjectToGrammarString(
+    pretty: boolean,
+    preserveCompilationError?: boolean | undefined,
+  ): GeneratorFn<void> {
     if (this.derivedProperty.body) {
       try {
         const lambdas = new Map<string, RawLambda>();
@@ -126,7 +129,9 @@ export class DerivedPropertyState extends LambdaEditorState {
             ? this.extractLambdaString(grammarText)
             : '',
         );
-        this.clearErrors();
+        this.clearErrors({
+          preserveCompilationError: preserveCompilationError,
+        });
       } catch (error) {
         assertErrorThrown(error);
         this.editorStore.applicationStore.logService.error(
@@ -196,7 +201,10 @@ export class ConstraintState extends LambdaEditorState {
     }
   }
 
-  *convertLambdaObjectToGrammarString(pretty: boolean): GeneratorFn<void> {
+  *convertLambdaObjectToGrammarString(
+    pretty: boolean,
+    preserveCompilationError?: boolean | undefined,
+  ): GeneratorFn<void> {
     if (!isStubbed_RawLambda(this.constraint.functionDefinition)) {
       try {
         const lambdas = new Map<string, RawLambda>();
@@ -212,7 +220,9 @@ export class ConstraintState extends LambdaEditorState {
             ? this.extractLambdaString(grammarText)
             : '',
         );
-        this.clearErrors();
+        this.clearErrors({
+          preserveCompilationError: preserveCompilationError,
+        });
       } catch (error) {
         assertErrorThrown(error);
         this.editorStore.applicationStore.logService.error(
--- a/packages/legend-graph/src/graph-manager/action/EngineError.ts
+++ b/packages/legend-graph/src/graph-manager/action/EngineError.ts
@@ -16,6 +16,7 @@
 
 import { ApplicationError } from '@finos/legend-shared';
 import type { SourceInformation } from '../action/SourceInformation.js';
+import { makeObservable, observable } from 'mobx';
 
 export class EngineError extends ApplicationError {
   sourceInformation?: SourceInformation | undefined;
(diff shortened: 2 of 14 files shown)


packages/legend-application-studio/src/stores/editor-state/element-editor-state/ClassState.ts
```diff
@@ -126,7 +126,6 @@ export class DerivedPropertyState extends LambdaEditorState {
             ? this.extractLambdaString(grammarText)
             : '',
         );
-        this.clearErrors();
```
akphi, line 129: I think this is intended behavior, because we change the rendering style from `standard` to `pretty` so source information from the old error might be off.

MauricioUyaguari: should we recomplile here if there is an existing error ? that way the user keeps the error with the updated source information?

akphi: I don't think the simple expand/collapse should bear the cost of recompilation. Also, it might be weird if we expand/collapse when user hasn't compile anytime, they will be caught off surprise to know this toggling cause compilation to occur. So IMO, we should leave this as is

MauricioUyaguari: The issue tho is that when there is an exisiting compilation error on that lambda and they expand it disappears. 
Not suggesting we do a compile everytime they toggle expand (that would be too expensive). but more: 
if there is an existing compile error on that lambda then recompile so we get the updated source information and show the correct source information.
Or maybe if we don't want to recompile, just keep the error message but remove the source info pointers as they are no longer relevant, if the user wants the updated, then they would need to hit f9 again.

packages/legend-application-studio/src/stores/editor-state/element-editor-state/ClassState.ts
```diff
@@ -126,7 +126,7 @@ export class DerivedPropertyState extends LambdaEditorState {
...
```
MauricioUyaguari, line 129: Isnt this change too broad. We call `convertLambdaObjectToGrammarString` when typing so would this change not delete errors when a user fixes an error ? 
Also could we update the video to show clicking the expand button

gayathrir11: do you think if it makes sense to add a flag to this function which tells us if user is trying to expand the lambda editor and clearErrorSourceINformation in that case?

packages/legend-application-studio/src/stores/editor/editor-state/element-editor-state/ClassState.ts
```diff
@@ -104,7 +104,10 @@ export class DerivedPropertyState extends LambdaEditorState {
     }
   }
 
-  *convertLambdaObjectToGrammarString(pretty: boolean): GeneratorFn<void> {
+  *convertLambdaObjectToGrammarString(
+    pretty: boolean,
+    clearErrorSourceInformation?: boolean | undefined,
```
MauricioUyaguari, line 109: maybe the flag name should be more specific to error. I think the point of the flag is we want to keep the error. 
so maybe 
`keepError?: boolean`
then there we keep error but clear source info

packages/legend-graph/src/graph-manager/action/EngineError.ts
```diff
@@ -16,9 +16,17 @@
...
 import type { SourceInformation } from '../action/SourceInformation.js';
+import { makeObservable, observable } from 'mobx';
 
 export class EngineError extends ApplicationError {
   sourceInformation?: SourceInformation | undefined;
+
+  constructor(message: string | undefined) {
+    super(message);
+    makeObservable(this, {
+      sourceInformation: observable,
```
MauricioUyaguari, line 27: we want to avoid adding observable directly on the model. we could add it in its downstream usage?

MauricioUyaguari: i meant more on the actual creation of it. so maybe we create an observe method and call it here `V1_buildCompilationError`

MauricioUyaguari (changes requested):
small comment but important

packages/legend-query-builder/src/stores/shared/LambdaEditorState.ts
```diff
@@ -115,5 +123,6 @@ export abstract class LambdaEditorState {
   abstract convertLambdaGrammarStringToObject(): GeneratorFn<void>;
   abstract convertLambdaObjectToGrammarString(
     pretty: boolean,
```
akphi, line 125: TODO: refactor to `options?: { ... }`

packages/legend-query-builder/src/stores/shared/LambdaEditorState.ts
```diff
@@ -69,6 +70,13 @@ export abstract class LambdaEditorState {
     this.setParserError(undefined);
   }
 
+  keepError(): void {
```
akphi, line 73: this method seems a bit confusing, I think it's better to have an option to alter the behavior of `clearErrors` instead, e.g.

```ts
clearErrors(options: { preserveCompilationError?: boolean })
```

akphi (changes requested):
nice and simple change! minor comments

Issue #1712, opened 2022-11-20
Feature request: Enable virtual assistant for all applications


akphi:
### Similar requests

- [X] I have searched and found no existing similar requests

### What problems are you trying to solve?

Enable virtual assistant in all applications (currently, we only enable this in `Legend Studio`)

### Describe the solution you would like to see implemented

- [ ] Move `VirtualAssistant` to `LegendApplicationComponentFrameworkProvider`
- [ ] Stop exporting `VirtualAssistant` component from `@finos/legend-application`, no other modules should use it
- [ ] Set up config for virtual assistant accordingly in other applications, create a new config to hide virtual assistant by default

### Contribution

- [ ] I would like to work on this feature

PR #1727, opened 2022-11-28
Enable virtual assistant for all applications

YannanGao-gs:
## Summary

Fixes https://github.com/finos/legend-studio/issues/1712
Depends on https://github.com/finos/legend/pull/566

- [ ] Test(s) added

Legend Query:

https://user-images.githubusercontent.com/73408381/204555601-7b136199-ce4c-4da6-82a8-3123945eb223.mov

Legend Taxonomy:

https://user-images.githubusercontent.com/73408381/204555965-5b13198e-4ede-4115-811f-2e15d34dc6e6.mov

### DIFF
--- a/.changeset/spicy-vans-add.md
+++ b/.changeset/spicy-vans-add.md
@@ -1,0 +1,4 @@
+---
+'@finos/legend-application': minor
+---
+Enable virtual assistant for all applications ([#1712](https://github.com/finos/legend-studio/issues/1712)).
--- a/packages/legend-application-query-bootstrap/scripts/setup.js
+++ b/packages/legend-application-query-bootstrap/scripts/setup.js
@@ -77,6 +77,14 @@ export const setup = (outputDir) => {
               url: 'https://legend.finos.org/resource/studio/documentation/shared.json',
               simple: true,
             },
+            {
+              url: 'https://legend.finos.org/resource/studio/documentation/query.json',
+              simple: true,
+            },
+            {
+              url: 'https://legend.finos.org/resource/studio/documentation/dsl-service.json',
+              simple: true,
+            },
           ],
         },
         taxonomy: {
--- a/packages/legend-application-query/src/components/QuerySetup.tsx
+++ b/packages/legend-application-query/src/components/QuerySetup.tsx
@@ -30,6 +30,7 @@ import {
   MenuContentItemLabel,
   CheckIcon,
   MenuContentDivider,
+  AssistantIcon,
 } from '@finos/legend-art';
 import {
   getQueryParameters,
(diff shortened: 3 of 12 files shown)


.changeset/spicy-vans-add.md
```diff
@@ -0,0 +1,8 @@
+---
+'@finos/legend-application': patch
+'@finos/legend-application-query': patch
+'@finos/legend-application-studio': patch
+'@finos/legend-application-taxonomy': patch
+'@finos/legend-query-builder': patch
+---
+Enable virtual assistant for all applications ([#1712](https://github.com/finos/legend-studio/issues/1712)).
```
akphi, line 8: This changeset should only apply for `@finos/legend-application` and it should be a `minor` change

akphi, line 8: This changeset should only apply for `@finos/legend-application` and it should be a `minor` change. Let's split up the changeset properly. Thanks!

packages/legend-application-query-bootstrap/scripts/setup.js
```diff
@@ -77,6 +77,10 @@ export const setup = (outputDir) => {
               url: 'https://legend.finos.org/resource/studio/documentation/shared.json',
               simple: true,
             },
+            {
```
akphi, line 80: like said, ideally, let's merge https://github.com/finos/legend/pull/566 first then we can also add https://legend.finos.org/resource/studio/documentation/dsl-service.json or something like that here. We should add this to other setup.js files as well

packages/legend-application-query/src/components/QuerySetup.tsx
```diff
@@ -276,77 +277,99 @@ export const QuerySetupLandingPage = withQuerySetupLandingPageStore(
...
       );
     const showAllActionGroup = (): void => setupStore.setShowAllGroups(true);
-
+    const toggleAssistant = (): void =>
+      applicationStore.assistantService.toggleAssistant();
     useEffect(() => {
       setupStore.initialize(showAdvancedActions, showAllGroups, tagToFocus);
     }, [setupStore, showAdvancedActions, showAllGroups, tagToFocus]);
 
     return (
```
akphi, line 286: I have to warn you one small thing here is that there are children screens of query setup (i.e. when you click on one of the modes), so we probably missed the status bar there

packages/legend-query-builder/src/components/QueryBuilder.tsx
```diff
@@ -150,6 +153,20 @@ const QueryBuilderStatusBar = observer(
           >
             <HackerIcon />
           </button>
+          <button
```
akphi, line 156: This is fine, but I also have to warn you of the potential consequence of this, in the embedded query builder, we now have the button to hide virtual assistant, that would make this button showing up twice in the UI (including the one in Legend Studio status bar)

akphi (changes requested):

packages/legend-application-query/src/components/QuerySetup.tsx
YannanGao-gs: status bar was added in the BasicQuerySetup, which will make sure there will be one status bar in every single children screen. (also double-checked by rendering each child component one by one) :) 
Screenshot 2022-11-29 at 11 57 55 AM

akphi (approved):

PR #1753, opened 2022-12-09
Fix a bug with PackageableTreeNode not getting selected for dependency elements

gayathrir11:
## Summary

Fix bug when PackageableTreeNode not getting selected for dependency elements when dependency tree is expanded in the explorer tree

- [ ] Test(s) added

https://user-images.githubusercontent.com/87973126/206780757-dcc6f3e8-ec0c-430d-a56c-4ace8acb4d33.mp4

### DIFF
--- a/.changeset/forty-onions-pretend.md
+++ b/.changeset/forty-onions-pretend.md
@@ -1,0 +1,5 @@
+---
+'@finos/legend-application-studio': patch
+---
+
+Fix bug when opening dependency element does not open corresponding element tree node in the dependency explorer
--- a/packages/legend-application-studio/src/stores/shared/PackageTreeUtils.ts
+++ b/packages/legend-application-studio/src/stores/shared/PackageTreeUtils.ts
@@ -252,6 +252,11 @@ export const openNode = (
     openingNode = !openingNode ? node : openingNode;
     currentElement = currentElement.package;
   }
+  // Open the dependency root
+  const node = treeData.nodes.get(currentElement.path);
+  if (node) {
+    node.isOpen = currentElement instanceof Package;
+  }
   return openingNode;
 };
 


packages/legend-application-studio/src/stores/shared/PackageTreeUtils.ts
```diff
@@ -244,6 +245,13 @@ export const openNode = (
 ): PackageTreeNodeData | undefined => {
   let currentElement: PackageableElement | undefined = element;
   let openingNode: PackageTreeNodeData | undefined;
+  const dependencyRootNode = treeData.nodes.get('');
```
akphi, line 248: why don't we use `treeData.rootIds`, `get('')` seems a bit arbitrary

packages/legend-application-studio/src/stores/shared/PackageTreeUtils.ts
```diff
@@ -244,6 +249,17 @@ export const openNode = (
...
+  let dependencyRootNode;
+  if (rootId !== undefined) {
+    dependencyRootNode = treeData.nodes.get(rootId);
+  }
+  if (
+    dependencyRootNode?.label ===
+    ExplorerTreeRootPackageLabel.PROJECT_DEPENDENCY
+  ) {
+    dependencyRootNode.isOpen = true;
+  }
```
MauricioUyaguari, line 262: I think the bug arises from when it hits the dependencyNode in the while loop below and since the dependencyNode.package is null, it doesn't open that node. 
Can't we instead just open that node as well. Do a check after the while loop  to ensure that the currentElement Node is open even if the package is null. 

I don't like having dependency specific logic here.

.changeset/forty-onions-pretend.md
```diff
@@ -0,0 +1,5 @@
+---
+'@finos/legend-application-studio': patch
+---
+
+Fix a bug with `PackageableTreeNode` not getting selected for dependency elements.
```
MauricioUyaguari, line 5: Fix bug when opening dependency element does not open corresponding element tree node in the dependency explorer?

MauricioUyaguari (approved):

PR #1785, opened 2022-12-16
Show dependency project information in explorer panel

### LINKED ISSUE #1641: Feature request: Better way to view project dependencies
### Similar requests

### How are you using Studio?

Legend Studio

The current way of viewing dependencies in the explorer tree is not ergonomic in certain cases, for example, it's hard to tell which dependency an element is from

- [ ] We should refer to how apps like `IntelliJ` does this (see screenshot)
- [ ] The current mode has its merit, we should consider have a config in explorer panel to customize/toggle between modes

image

_No response_

_No response_

### Contribution

gayathrir11:
## Summary

closes #1641 

- [ ] Test(s) added

https://user-images.githubusercontent.com/87973126/208113164-125279a2-ae15-49a4-a075-51a4b524a4a0.mp4

*** Ctrl + P demo for dependency elements***

https://user-images.githubusercontent.com/87973126/209798542-76ea6a75-e6a4-4991-ba45-25945ad081b7.mp4

### DIFF
--- a/packages/legend-application-studio/src/components/editor/side-bar/Explorer.tsx
+++ b/packages/legend-application-studio/src/components/editor/side-bar/Explorer.tsx
@@ -65,8 +65,15 @@ import {
   getFileSystemChildNodes,
 } from '../../../stores/shared/FileSystemTreeUtils.js';
 import { FileSystemTree } from '../edit-panel/element-generation-editor/FileSystemViewer.js';
-import { generateViewEntityRoute } from '../../../stores/LegendStudioRouter.js';
-import { toTitleCase } from '@finos/legend-shared';
+import {
+  generateViewEntityRoute,
+  generateViewProjectByGAVRoute,
+} from '../../../stores/LegendStudioRouter.js';
+import {
+  guaranteeNonEmptyString,
+  guaranteeNonNullable,
+  toTitleCase,
+} from '@finos/legend-shared';
 import { flowResult } from 'mobx';
 import { useEditorStore } from '../EditorStoreProvider.js';
 import {
@@ -89,6 +96,11 @@ import { useApplicationStore } from '@finos/legend-application';
 import { PACKAGEABLE_ELEMENT_TYPE } from '../../../stores/shared/ModelClassifierUtils.js';
 import { useLegendStudioApplicationStore } from '../../LegendStudioBaseStoreProvider.js';
 import { queryClass } from '../edit-panel/uml-editor/ClassQueryBuilder.js';
+import { createViewSDLCProjectHandler } from '../../../stores/DependencyProjectViewerHelper.js';
+import {
+  MASTER_SNAPSHOT_ALIAS,
+  SNAPSHOT_VERSION_ALIAS,
+} from '@finos/legend-server-depot';
 
 const ElementRenamer = observer(() => {
   const editorStore = useEditorStore();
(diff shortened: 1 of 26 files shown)


packages/legend-graph/src/graph/helpers/DomainHelper.ts
```diff
@@ -230,11 +230,21 @@ export const isMainGraphElement = (
 export const isElementReadOnly = (element: PackageableElement): boolean =>
   !isMainGraphElement(element);
 
+// Each dependency project is considered as a root. Dependency project name is `${groupId}:${artifactId}`
+// So dependency name contains ':'. Here we are making the assumption that only dependecy project name
+// has this exception and we are making the check for dependency element on basis of that.
 export const isDependencyElement = (
```
MauricioUyaguari, line 236: Not really that excited for this check here. 
Could we maybe have pureModel been an input to this function as well and check if the root package of the element is one of the root packages in dependency manager of the pure model?

gayathrir11: makes sense. Will modify that

packages/legend-application-studio/src/components/editor/side-bar/Explorer.tsx
```diff
@@ -291,6 +295,55 @@ const ExplorerContextMenu = observer(
       (type: string): (() => void) =>
       (): void =>
         editorStore.newElementState.openModal(type, _package);
+    const isDependencyProjectRoot = (): boolean =>
```
MauricioUyaguari, line 298: Can't we just compare the node.package to see if its one of the packages of dependencyManager ? Don't like really comparing labels as in theory we can always change the label

MauricioUyaguari (changes requested):
For the video, could we show Ctlr + P on a dependent element and the element on the explorer panel should open

packages/legend-query-builder/src/stores/explorer/QueryFunctionsExplorerState.ts
```diff
@@ -139,7 +139,7 @@ const generateFunctionsExplorerTreeNodeChilrdren = (
 };
 
 export const getFunctionsExplorerTreeData = (
```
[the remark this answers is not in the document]
YannanGao-gs: I might be wrong.. but I think this change might not be right.
The original definition of root is `root = new Package(ROOT_PACKAGE_NAME.PROJECT_DEPENDENCY_ROOT);`. 
If there are two packages A, B from dependencies, root's children will be A,B.
```
root 
  -- A
        -- C
  -- B
        -- D
```

Now, the definition of root is `roots: Package[] = [];`, which means roots = [A, B] for the hierarchy above.
Then the logic below might not be equal. 
```
  root.children
    .slice()
    .filter((child) => !(child instanceof Unit))
    .filter(
      (child) =>
        child instanceof Package && validDisplayablePackageSet.has(child),
    )
```

The original root.children == [A, B], now root.children == C or D
which could be different for the case where there is a function just belongs to the root package.

packages/legend-graph/src/graph/helpers/DomainHelper.ts
```diff
@@ -230,11 +230,21 @@ export const isMainGraphElement = (
 export const isElementReadOnly = (element: PackageableElement): boolean =>
   !isMainGraphElement(element);
 
+// Each dependency project is considered as a root. Dependency project name is `${groupId}:${artifactId}`
```
MauricioUyaguari, line 233: can we remove this comment now?

packages/legend-query-builder/src/stores/explorer/QueryFunctionsExplorerState.ts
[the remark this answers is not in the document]
MauricioUyaguari: please meet and clarify whether this will be an issue or not thanks!

packages/legend-graph/src/graph/helpers/DomainHelper.ts
```diff
@@ -232,9 +232,16 @@ export const isElementReadOnly = (element: PackageableElement): boolean =>
...
-  ROOT_PACKAGE_NAME.PROJECT_DEPENDENCY_ROOT;
+  pureModel: PureModel,
+): element is PackageableElement => {
+  const rootPackage = returnUndefOnError(() => getElementRootPackage(element));
+  return (
+    (rootPackage?.name === ROOT_PACKAGE_NAME.PROJECT_DEPENDENCY_ROOT ||
+      (rootPackage &&
+        pureModel.dependencyManager.roots.includes(rootPackage))) ??
+    false
+  );
```
MauricioUyaguari, line 243: I don't understand the need for `?? false` in this boolean expression.

MauricioUyaguari (approved):

PR #1807, opened 2023-01-06
Service Test Editor UI Imp: Label options in the connection-id dropdown in the service test editor

YannanGao-gs:
## Summary

Label options in the connection-id dropdown in the service test editor will benefit users when there are identical connection ids from different runtimes.

- [ ] Test(s) added

Screenshot 2023-01-23 at 3 02 27 PM

### DIFF
--- a/.changeset/clean-lemons-grab.md
+++ b/.changeset/clean-lemons-grab.md
@@ -1,0 +1,5 @@
+---
+'@finos/legend-application-studio': patch
+---
+
+Label options in the connection-id dropdown in the service test editor.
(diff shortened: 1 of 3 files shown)


gayathrir11:
I think naming `IdentifiedConnectionWithRuntime` is a bit confusing because connections doesn't have a runtime. It's the other way around. Can we name it better? May be `RuntimeConnections` or something?
At other places I think we don't really have to mention that it is a connection from runtime. Why don't we add a note at the starting of the file regarding the assumption we make on all the connections are coming from a runtime as it is service?
I feel like we should just add a new field in `NewConnectionDataState` for runtime instead of introducing a type for this.
What do you think?

MauricioUyaguari:
@YannanGao-gs  
Don't really like the format of the label. 

Maybe we could have a label for the runtime the connection belongs to
and another label for the store it contains. 
We don't really need the connectionId again in the label since that is the main selection point.

PR #1810, opened 2023-01-10
fix regression introduced in #1693

gayathrir11:
## Summary

Incorrect change detection behaviour when we are re-directed to text mode due to graph build failure.

- [ ] Test(s) added

https://user-images.githubusercontent.com/87973126/211583349-07ea8a39-6d8e-40a5-9b8c-0f327e32b3cd.mp4

### DIFF
--- a/packages/legend-application-studio/src/stores/EditorStore.ts
+++ b/packages/legend-application-studio/src/stores/EditorStore.ts
@@ -347,11 +347,21 @@ export class EditorStore implements CommandRegistrar {
     if (graphEditor === GRAPH_EDITOR_MODE.GRAMMAR_TEXT) {
       // Stop change detection as we don't need the actual change detection in text mode
       this.changeDetectionState.stop();
-      this.changeDetectionState.computeLocalChangesInTextMode(
-        (yield this.graphManagerState.graphManager.pureCodeToEntities(
-          this.grammarTextEditorState.graphGrammarText,
-        )) as Entity[],
-      );
+      try {
+        yield flowResult(
+          this.changeDetectionState.computeLocalChangesInTextMode(
+            (yield this.graphManagerState.graphManager.pureCodeToEntities(
+              this.grammarTextEditorState.graphGrammarText,
+            )) as Entity[],
+          ),
+        );
+      } catch (error) {
+        assertErrorThrown(error);
+        this.applicationStore.log.warn(
+          LogEvent.create(GRAPH_MANAGER_EVENT.PARSING_FAILURE),
+          error,
+        );
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


packages/legend-application-studio/src/stores/sidebar-state/LocalChangesState.ts
```diff
@@ -822,9 +822,11 @@ export class TextLocalChangesState extends LocalChangesState {
   }
 
   *restartChangeDetection(): GeneratorFn<void> {
-    this.editorStore.changeDetectionState.computeLocalChangesInTextMode(
-      this.editorStore.changeDetectionState.workspaceLocalLatestRevisionState
-        .entities,
-    );
+    yield Promise.all([
```
akphi, line 825: do you need this `yield Promise.all()` if this is the only function you call? Also, looks like this function does not need to be a generator, can we change to have it as an `async` function?

packages/legend-application-studio/src/stores/EditorStore.ts
```diff
@@ -340,17 +340,23 @@ export class EditorStore implements CommandRegistrar {
     this.mode = val;
   }
 
-  *setGraphEditMode(graphEditor: GRAPH_EDITOR_MODE): GeneratorFn<void> {
+  *setGraphEditMode(
+    graphEditor: GRAPH_EDITOR_MODE,
+    entities?: Entity[],
```
akphi, line 345: I understand why we need to do this, but this seems really hacky, especially when one considers the name of the function :(

akphi (dismissed):

packages/legend-application-studio/src/stores/EditorStore.ts
```diff
@@ -347,11 +347,17 @@ export class EditorStore implements CommandRegistrar {
...
+      try {
+        yield flowResult(
+          this.changeDetectionState.computeLocalChangesInTextMode(
+            (yield this.graphManagerState.graphManager.pureCodeToEntities(
+              this.grammarTextEditorState.graphGrammarText,
+            )) as Entity[],
+          ),
+        );
+      } catch (error) {
+        assertErrorThrown(error);
```
MauricioUyaguari, line 359: Could we console this error as a warning. 
Also could you explain how this fixes the problem. Does it allow something else to happen after this method. 
My guess is we are now able to set the current revision entities (in the video case the dataspace and model), so that way when user recompiles it is compared it against those entities.

gayathrir11: yeah now it is able to set the entityIndexes. When user recompiles it compares against that entityIndexes.

MauricioUyaguari (approved):

PR #1823, opened 2023-01-18
minor changes to dependency explorer tree

gayathrir11:
## Summary

- [x] Modify dependency root label to `artifactId` instead of `groupId:artifactId`
- [x] FIx `Copy Link` option for dependency element

- [ ] Test(s) added

https://user-images.githubusercontent.com/87973126/213130327-923b03a9-e046-4f23-a50b-67624608c8e5.mp4

### DIFF
--- a/.changeset/fifty-swans-sniff.md
+++ b/.changeset/fifty-swans-sniff.md
@@ -1,0 +1,5 @@
+---
+'@finos/legend-application-studio': patch
+---
+
+Modify the label and fix the copy link functionality for dependependency tree elements
--- a/packages/legend-application-studio/src/components/editor/side-bar/Explorer.tsx
+++ b/packages/legend-application-studio/src/components/editor/side-bar/Explorer.tsx
@@ -91,6 +91,7 @@ import {
   isMainGraphElement,
   getFunctionSignature,
   getFunctionNameWithPath,
+  getElementRootPackage,
 } from '@finos/legend-graph';
 import { useApplicationStore } from '@finos/legend-application';
 import { PACKAGEABLE_ELEMENT_TYPE } from '../../../stores/shared/ModelClassifierUtils.js';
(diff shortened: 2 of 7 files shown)


.changeset/fifty-swans-sniff.md
```diff
@@ -0,0 +1,3 @@
+---
+'@finos/legend-application-studio': patch
```
YannanGao-gs, line 2: if it's possible, would you mind adding some descriptions in the changeset?

packages/legend-application-studio/src/stores/shared/PackageTreeUtils.ts
```diff
@@ -48,6 +48,10 @@ import {
   getElementRootPackage,
 } from '@finos/legend-graph';
 import { ExplorerTreeRootPackageLabel } from '../ExplorerTreeState.js';
+import { GAV_DELIMITER } from '@finos/legend-storage';
+
+const getArtifactIdFromProjectId = (projectId: string): string =>
```
MauricioUyaguari, line 53: We should be cautious here in case there is not always an artifactId. 
Legacy projectId may still occur. 
We can use the same logic of ProjectDependency. Is its not legacy dependency then use the logic here else show full project id.

packages/legend-application-studio/src/components/editor/side-bar/Explorer.tsx
```diff
@@ -283,20 +284,44 @@ const ExplorerContextMenu = observer(
     };
     const copyWorkspaceElementLink = (): void => {
```
MauricioUyaguari, line 285: can we add a `copy SDLC link` option as well to copy the sdlc project view in addition to this one. 
Thanks

MauricioUyaguari (changes requested):

MauricioUyaguari (approved):

PR #1833, opened 2023-01-20
Apply strict rules for association properties

### LINKED ISSUE #282: Feature request: Stricter association editor
## 🚀 Feature Request

Association is a powerful feature conceptually. However, it can be used to do fairly dangerous things, such as to extend system classes as it allows modification of immutable elements. As such, we should limit the association to only be limited to certain spaces, e.g. we cannot associate a class to classes from the `system`, or `project dependencies`.

- [ ] Assert this validation when building the graph, we should also incorporate this into `strict` mode, introduced by #1540 
- [ ] We should show warning in association editor
- [ ] In the `dispose` method of `classes`, let's also cleanup associations linkage: i.e. `propertiesFromAssociations`, etc.

gayathrir11:
## Summary

closes #282 

depends on finos/legend#593

- [ ] Test(s) added

https://user-images.githubusercontent.com/87973126/216938050-bb8aec4d-2b5e-4208-87bc-4947875575cc.mp4

### DIFF
--- a/.changeset/quick-books-switch.md
+++ b/.changeset/quick-books-switch.md
@@ -1,0 +1,6 @@
+---
+'@finos/legend-application-studio': patch
+'@finos/legend-graph': patch
+---
+
+Add warnings when associations reference system elements. ([#282](https://github.com/finos/legend-studio/issues/282))
--- a/packages/legend-application-studio/src/components/editor/edit-panel/uml-editor/AssociationEditor.tsx
+++ b/packages/legend-application-studio/src/components/editor/edit-panel/uml-editor/AssociationEditor.tsx
@@ -43,6 +43,7 @@ import {
   LongArrowRightIcon,
   PanelDropZone,
   Panel,
+  InfoCircleIcon,
 } from '@finos/legend-art';
 import { getElementIcon } from '../../../shared/ElementIconUtils.js';
 import { prettyCONSTName, guaranteeType } from '@finos/legend-shared';
@@ -92,6 +93,7 @@ import {
   buildElementOption,
   useApplicationNavigationContext,
   type PackageableElementOption,
+  LEGEND_APPLICATION_DOCUMENTATION_KEY,
 } from '@finos/legend-application';
 import { LEGEND_STUDIO_APPLICATION_NAVIGATION_CONTEXT_KEY } from '../../../../stores/LegendStudioApplicationNavigationContext.js';
 
@@ -113,6 +115,21 @@ const AssociationPropertyBasicEditor = observer(
       }
       return false;
     };
+    const isPropertyInvalid = (): boolean => {
+      const sytemAssociation =
+        editorStore.graphManagerState.graph.systemModel.getOwnNullableAssociation(
+          association.path,
+        );
+      if (
+        !sytemAssociation &&
+        editorStore.graphManagerState.graph.systemModel.getOwnNullableClass(
+          property.genericType.value.rawType.path,
+        )
+      ) {
+        return true;
+      }
+      return false;
+    };
     // Name
     const changeValue: React.ChangeEventHandler<HTMLInputElement> = (event) => {
       property_setName(property, event.target.value);
@@ -121,7 +138,9 @@ const AssociationPropertyBasicEditor = observer(
     const [isEditingType, setIsEditingType] = useState(false);
     // TODO: make this so that association can only refer to classes from the same graph space
     const propertyTypeOptions =
-      editorStore.graphManagerState.usableClasses.map(buildElementOption);
+      editorStore.graphManagerState.usableAssociationPropertyClasses.map(
+        buildElementOption,
+      );
     const propertyType = property.genericType.value.rawType;
     const propertyTypeName = getClassPropertyType(propertyType);
     const filterOption = createFilter({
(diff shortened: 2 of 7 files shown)


MauricioUyaguari:
@gayathrir11  As discussed offline, 
Let's do the following
1. Let's add some type of red error message under the property that contains a linkage to dependency classes. We want that info to have a link to a documentation explaining why. 
2. Add documentation to the legend page

We may need to work with @xannem to find a place in the documentation for this. We

packages/legend-graph/src/graphManager/protocol/pure/v1/transformation/pureGraph/to/helpers/V1_DomainBuilderHelper.ts
```diff
@@ -262,6 +264,23 @@ export const V1_buildAssociationProperty = (
     associatedProperty.type,
     `Association associated property 'type' field is missing`,
   );
+  const _class = context.graph.getMainGraphClass(associatedPropertyClassType);
+  const _association = context.graph.getOwnNullableAssociation(
```
MauricioUyaguari, line 268: I dont think this is correct. Will this work if the association is part of a dependency?

packages/legend-graph/src/graphManager/protocol/pure/v1/transformation/pureGraph/to/helpers/V1_DomainBuilderHelper.ts
```diff
@@ -262,6 +264,23 @@ export const V1_buildAssociationProperty = (
...
+     * See https://github.com/finos/legend-studio/issues/941
+     *
+     * @discrepancy graph-building
+     */
+    if (context.options?.strict) {
+      throw new GraphBuilderError(message);
+    }
+    context.log.warn(LogEvent.create(message));
+  }
   const associatedClass = context.resolveClass(associatedPropertyClassType);
```
MauricioUyaguari, line 284: Can't we move this logic as a helper method. takes the association, the property , pure model. Then check if assoication, property class are in the main graph.

packages/legend-graph/src/graphManager/GraphManagerState.ts
```diff
@@ -266,6 +267,9 @@ export class GraphManagerState extends BasicGraphManagerState {
       ...this.graph.ownClasses,
     ];
   }
+  get mainGraphClasses(): Class[] {
```
MauricioUyaguari, line 270: maybe to keep in line with the rest we should use something like `usableAssociationPropertyClasses`

.changeset/quick-books-switch.md
```diff
@@ -0,0 +1,6 @@
+---
+'@finos/legend-application-studio': patch
+'@finos/legend-graph': patch
+---
+
+Apply strict rules for association properties. ([#282](https://github.com/finos/legend-studio/issues/282))
```
MauricioUyaguari, line 6: Add warnings when associations reference system elements.

packages/legend-graph/src/graphManager/protocol/pure/v1/transformation/pureGraph/to/helpers/V1_DomainBuilderHelper.ts
```diff
@@ -262,6 +292,11 @@ export const V1_buildAssociationProperty = (
...
     `Association associated property 'type' field is missing`,
   );
+  validateAssociationProperty(
+    pureAssociation,
+    associatedPropertyClassType,
+    context,
+  );
   const associatedClass = context.resolveClass(associatedPropertyClassType);
   const property = V1_buildProperty(currentProperty, context, pureAssociation);
   associatedClass.value.propertiesFromAssociations.push(property);
```
MauricioUyaguari, line 302: Would it better to validate once the property has been built.
If the property type is in system then its invalid ?

packages/legend-application-studio/src/components/editor/edit-panel/uml-editor/AssociationEditor.tsx
```diff
@@ -113,6 +115,21 @@ const AssociationPropertyBasicEditor = observer(
       }
       return false;
     };
+    const isPropertyInvalid = (): boolean => {
+      const _association =
```
MauricioUyaguari, line 119: Why are we refetching the assoication if its defined above. 
I think its okay to show this error even on dependency association since a reference to system should `NEVER` happen unless the 2 classes and assoication are all in system.
So for this check i would just say do a check that assoication is not in system and that the proeprty type is in system. if that is both true than its invalid.

MauricioUyaguari (changes requested):
Few comments. 
@gayathrir11 could you also please (separate PR if needed) fix the drag and drop from explorer on association properties. It doesn't seem to be working.

packages/legend-graph/src/graphManager/protocol/pure/v1/transformation/pureGraph/to/helpers/V1_DomainBuilderHelper.ts
gayathrir11: I think we should do it before building the graph Because these validations are being made while we build the graph so it's better to check if it is not present system first and build the property

MauricioUyaguari (approved):

MauricioUyaguari:
@gayathrir11  lets add a test please. Thanks

Issue #1672, opened 2022-11-09
Minor improvements to productionize query 


MauricioUyaguari:
### Similar requests

- [X] I have searched and found no existing similar requests

### How are you using Studio?

Legend Query

### What problems are you trying to solve?

Minor improvements to productionize query can move to separate issues once work is started 

- [x] add owners to productionize query form
- [ ] add generate test data if possible to productionize query form
- [ ] ability register semi interactive service on Stand Alone Query 
- [ ] generate service URL pattern smartly (respecting the parameters, use service name for path)
- [ ] `bug` Dataspace explorer search not working properly. More invesitgation needed but `CovidDataspace` was created and when searched, metadata was returning the dataspace but query was failing to show option 

### Describe the solution you would like to see implemented

Separate tasks to different issues once tackled

### Contribution

- [ ] I would like to work on this feature

PR #1834, opened 2023-01-20
Support semi-interactive registration of services from Standalone query

gayathrir11:
## Summary

#1672 

- [ ] Test(s) added

https://user-images.githubusercontent.com/87973126/213707075-c743c181-efeb-48a5-95e5-1a3c65d96095.mp4

### DIFF
--- a/packages/legend-extension-dsl-service/src/components/studio/ServiceRegisterModal.tsx
+++ b/packages/legend-extension-dsl-service/src/components/studio/ServiceRegisterModal.tsx
@@ -14,32 +14,22 @@
  * limitations under the License.
  */
 
-import { useMemo, useState } from 'react';
+import { useState } from 'react';
 import { observer } from 'mobx-react-lite';
 import {
   ActionState,
   assertErrorThrown,
   assertNonEmptyString,
-  debounce,
-  type GeneratorFn,
   guaranteeNonNullable,
   LogEvent,
-  type PlainObject,
   uuid,
 } from '@finos/legend-shared';
-import {
-  User,
-  ProjectConfiguration,
-  useSDLCServerClient,
-  type SDLCServerClient,
-} from '@finos/legend-server-sdlc';
 import {
   generateServiceManagementUrl,
   LATEST_PROJECT_REVISION,
   LEGEND_STUDIO_APP_EVENT,
 } from '@finos/legend-application-studio';
 import { createServiceElement } from '../../stores/studio/QueryProductionizerStore.js';
-import { flowResult, flow } from 'mobx';
 import {
   CheckSquareIcon,
   clsx,
@@ -50,7 +40,7 @@ import {
   Panel,
   PanelFullContent,
   PanelLoadingIndicator,
-  RobotIcon,
+  RocketIcon,
   SquareIcon,
 } from '@finos/legend-art';
 import {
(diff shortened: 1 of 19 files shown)


MauricioUyaguari:
@beekemarie

MauricioUyaguari:
@gayathrir11 lets remove the dependency on sdlc-server on stand alone query. For owners we can just go back to simple inputs. 
Reason being is there are lot of implications of adding sdlc to query. We can look for the search users api to take an api in the future but for now lets just do that.

packages/legend-extension-dsl-service/src/components/studio/ServiceRegisterModal.tsx
```diff
@@ -0,0 +1,442 @@
...
+    const onClose = (): void => setShowRegisterServiceModal(false);
+    return (
+      <>
+        <button
+          className="query-editor__header__action btn--dark"
+          tabIndex={-1}
+          onClick={registerCurrentQuery}
+          title="Register query as service..."
+        >
+          <RobotIcon />
```
MauricioUyaguari, line 434: The main action here is register so let's change the icon to the rocket icon which symbolizes the register.

MauricioUyaguari (changes requested):
Looks good. 

Not a hard blocker (maybe separate PR), but could we update the register service modal in the studio service extension to look more like this. We can hide like the path there but we should have them have the same format/input order for consistency.

Main thing is removing the sdlc server. Basically adding a dependency to gitlab assumes that users on the stand alone will have access to gitlab.

packages/legend-extension-dsl-service/src/components/studio/ServiceRegisterModal.tsx
```diff
@@ -0,0 +1,376 @@
...
+} from '@finos/legend-graph';
+import type { ExistingQueryEditorStore } from '@finos/legend-application-query';
+import { ProjectData } from '@finos/legend-server-depot';
+
+const ServiceRegisterModal = observer(
+  (props: { editorStore: ExistingQueryEditorStore; onClose(): void }) => {
+    const { editorStore, onClose } = props;
+    const [registrationState] = useState(ActionState.create());
+    const [text, setText] = useState('');
+    const [serviceEnv, setServiceEnv] = useState<string | undefined>(undefined);
```
MauricioUyaguari, line 63: We should avoid defaulting to `null|undefined` is possible.
So please:
- activate service should be defaulted to `true`
- the service env should be defaulted to the first option
- owners should IF POSSIBLE. have the current user as the first owner.

packages/legend-extension-dsl-service/src/components/studio/ServiceRegisterModal.tsx
```diff
@@ -0,0 +1,376 @@
...
+          return;
+        }
+        try {
+          registrationState.inProgress();
+          const serverUrl = guaranteeNonNullable(
+            serverRegistrationOptions.find(
+              (option) => option.env === selectedEnvOption.value,
+            )?.managementUrl,
+          );
+          const versionInput = LATEST_PROJECT_REVISION;
```
MauricioUyaguari, line 147: why is the versionInput this?

packages/legend-extension-dsl-service/src/components/studio/ServiceRegisterModal.tsx
```diff
@@ -0,0 +1,376 @@
...
+          !servicePattern ||
+          !isServicePathValid ||
+          !isServiceUrlPatternValid ||
+          !selectedEnvOption
+        ) {
+          return;
+        }
+        try {
+          registrationState.inProgress();
+          const serverUrl = guaranteeNonNullable(
```
MauricioUyaguari, line 142: serverUrl should be the executionUrl.

packages/legend-extension-dsl-service/src/components/studio/ServiceRegisterModal.tsx
```diff
@@ -0,0 +1,376 @@
...
+              'You can now launch and monitor the operation of your service',
+            type: ActionAlertType.STANDARD,
+            actions: [
+              {
+                label: 'Launch Service',
+                type: ActionAlertActionType.PROCEED,
+                handler: (): void => {
+                  editorStore.applicationStore.navigator.visitAddress(
+                    generateServiceManagementUrl(
+                      guaranteeNonNullable(serverUrl),
```
MauricioUyaguari, line 194: here we use the managementURL

packages/legend-extension-dsl-service/src/components/studio/ServiceRegisterModal.tsx
```diff
@@ -0,0 +1,376 @@
...
+            currentQueryInfo.content,
+            currentQueryInfo.mapping,
+            currentQueryInfo.runtime,
+            editorStore.graphManagerState,
+          );
+          const serviceRegistrationResult =
+            await editorStore.graphManagerState.graphManager.registerService(
+              service,
+              editorStore.graphManagerState.graph,
+              project.groupId,
```
MauricioUyaguari, line 162: for groupId,artifactId, versionId. Why are we fetching the project again. 
The query already holds everything we need.
`editorStore.query.groupId, editorStore.query.artifactId, editorStore.query.versionId`
If the version is head then the input is undefined for the registerService function

packages/legend-extension-dsl-service/src/components/studio/ServiceRegisterModal.tsx
```diff
@@ -0,0 +1,376 @@
...
+              (option) => option.env === selectedEnvOption.value,
+            )?.managementUrl,
+          );
+          const versionInput = LATEST_PROJECT_REVISION;
+          registrationState.setMessage(`Registering service...`);
+          const service = await createServiceElement(
+            servicePath,
+            servicePattern,
+            owners,
+            currentQueryInfo.content,
```
MauricioUyaguari, line 153: I believe the `currentQueryInfo.content` here refers to the saved content of the query. 
We should be using the current query that the user is editing at the moment. Doesn't necesarrily mean it will be saved. Same with mapping, runtime.

PR #1883, opened 2023-02-07
Don't fetch mapping analysis if mapping hasn't changed

MauricioUyaguari:

- [ ] Test(s) added

### DIFF
--- a/.changeset/swift-rivers-kick.md
+++ b/.changeset/swift-rivers-kick.md
@@ -1,0 +1,5 @@
+---
+'@finos/legend-application-query': patch
+'@finos/legend-query-builder': patch
+'@finos/legend-graph': patch
+---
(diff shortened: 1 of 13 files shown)


akphi (approved):

MauricioUyaguari:
Overall
- icons overall don't make sense; switch over from icons to words (productionize, save, load query)
- change on aggregation name should change column name (i.e OLAP, groupBY)


Filter Panel

- Ensure drag and drop inside the operation works. Move In/Move out operators
- Fix Creating filter groups from explorer panel


Validation:
- Fix validation on parameters for existing constant values


Other:
- Right now can't copy over most dsb queries. Query Only accepts certain variance of projection. 
	- think about supporting more variance and saving variance into state 
- Result modifiers do not enforce order which can cause rewrite of query. (i.e project()->take()->filter() vs project()->filter()->take()) get generated to same query

Scaling:
- Query using huge amounts of JS look into
- etags leverage to save depot information 
- sending pmcd when executing via stand alone; we should cache this.
	- current solutions are engine to accept project coordinates; cache current pmcd session

Issue #317, opened 2021-07-05
RFC: Rework Modes


akphi:
## 💬 Request for Comments

Within Studio, we currently support 3 modes: `editor`, `viewer`, and `review`, each is somewhat being treated differently: e.g. `viewer` mode does not allow editing elements, `review` mode does not allow viewing and working with the element, etc. When `query builder` becomes standalone, we might need to yet support another mode. What we realize is,  these modes are not too far off from one another. In fact, we are trying to bring them closer in functionalities - see https://github.com/finos/legend-studio/issues/297.

In fact, we should consider all the flows and just make them different in a very small set of features. Instead of thinking that we should create a brand new mode without any functionality and then adding them one by one (like what we're trying to do in https://github.com/finos/legend-studio/issues/297), we probably should start using `EditorStore` and having ways to disable certain features if that's cleaner. i.e. starts from the full one - `editor mode` - and see what we can turn off

### Implementation Plan

- [ ] Viewer mode:
  - [ ] Disable saving/Review/Workspace Updater
  - [ ] Keep the LocalChange panel around, but then have a feature so that people can make changes to model and if they like, they can create a new workspace with these changes (almost like `git checkout -b <branch_name>` behaviour)
- [ ] Review mode:
  - [ ] Disable saving/review/Workspace updater
- [ ] Query builder viewer mode - this is the base for the standalone query builder
  - [ ] Disable editing everything, everything would be readonly (similar to the current viewer mode)

PR #1898, opened 2023-02-10
Refactor code in `EditorStore` to separate out logic for FormMode and TextMode

gayathrir11:

- [ ] Test(s) added


packages/legend-application-studio/src/stores/EditorStore.ts
```diff
@@ -169,8 +135,8 @@ export class EditorStore implements CommandRegistrar {
   sdlcState: EditorSDLCState;
   graphState: EditorGraphState;
   graphManagerState: GraphManagerState;
+  graphEditModeManagerState: GraphEditorModeManagerState;
```
akphi, line 138: im thinking of a more succinct name for this. Since calling `editorStore.graphEditModeManagerState.graphEditModeState` everywhere seems quite laborous, maybe what we should do is to squash `GraphEditorModeManagerState` into `EditorStore` and call `graphEditModeState` `graphEditorMode`, so we simply calling `editorStore.graphEditorMode`

packages/legend-application-studio/src/stores/GraphEditModeManagerState.ts
```diff
@@ -0,0 +1,1119 @@
...
+import {
+  graph_addElement,
+  graph_deleteElement,
+  graph_deleteOwnElement,
+  graph_dispose,
+  graph_renameElement,
+} from './shared/modifier/GraphModifierHelper.js';
+import { ElementEditorState } from './editor-state/element-editor-state/ElementEditorState.js';
+
+export abstract class GraphEditModeState {
```
akphi, line 68: --> `GraphEditorMode`

packages/legend-application-studio/src/stores/EditorStore.ts
akphi: eventually, we can rename/remove `EditorMode` and this should be the defacto `EditorMode` - https://github.com/finos/legend-studio/issues/317

packages/legend-application-studio/src/stores/GraphEditModeManagerState.ts
```diff
@@ -0,0 +1,1119 @@
...
+  ): GeneratorFn<void>;
+  abstract deleteElement(element: PackageableElement): GeneratorFn<void>;
+  abstract renameElement(
+    element: PackageableElement,
+    newPath: string,
+  ): GeneratorFn<void>;
+  abstract getCurrentGraphHash(): string | undefined;
+  abstract globalCompile(): GeneratorFn<unknown>;
+  abstract updateGraphAndApplication(entities: Entity[]): GeneratorFn<void>;
+  abstract get editMode(): GRAPH_EDITOR_MODE;
```
akphi, line 97: --> `mode` or `name`

gayathrir11:
I agree having source mode and target mode in switch mode currently doesn't make much sense as we know what to do when we are entering a garph edit mode. But when we have multi file support knowing source mode and target mode is useful as the set of operations we do is different.
Let us assume if we want to switch to form mode, the set of apis we call would be different when we are in single file text mode and multi file text mode. I don't want to do bunch of if-else here again to know what mode i am in currently. I hope this reason is not trivial.

packages/legend-extension-dsl-diagram/src/stores/studio/DiagramEditorState.ts
```diff
@@ -93,7 +93,10 @@ export class DiagramEditorClassViewEditorSidePanelState extends DiagramEditorSid
         (elementState) =>
           isType(elementState, ClassEditorState) &&
           elementState.element === classView.class.value,
-      ) ?? this.editorStore.createElementEditorState(classView.class.value),
+      ) ??
+        this.editorStore.tabManagerState.createElementEditorState(
```
akphi, line 97: here, I think it's best to avoid this weird usage, rather, let's just do

```ts
?? new ClassEditorState(this.editorStore, classView.class.value)
```

packages/legend-application-studio/src/stores/EditorTabManagerState.ts
```diff
@@ -53,11 +90,72 @@ export class EditorTabManagerState extends TabManagerState {
     );
   }
 
+  createElementEditorState(
```
akphi, line 93: probably safe to make this method `private`

packages/legend-application-studio/src/stores/EditorGraphState.ts
```diff
@@ -85,42 +80,28 @@ import {
...
-import { LegendStudioTelemetry } from './LegendStudioTelemetry.js';
 
 export enum GraphBuilderStatus {
   SUCCEEDED = 'SUCCEEDED',
   FAILED = 'FAILED',
   REDIRECTED_TO_TEXT_MODE = 'REDIRECTED_TO_TEXT_MODE',
 }
 
-export enum FormModeCompilationOutcome {
+export enum GraphEditorModeCompilationOutcome {
```
akphi, line 101:
```suggestion
export enum GraphCompilationOutcome {
```

packages/legend-application-studio/src/stores/EditorStore.ts
```diff
@@ -169,8 +137,8 @@ export class EditorStore implements CommandRegistrar {
   sdlcState: EditorSDLCState;
   graphState: EditorGraphState;
   graphManagerState: GraphManagerState;
+  graphEditorMode: GraphEditorMode;
```
akphi, line 140: the fact that we have `graphState: EditorGraphState` really burns me up, but I think for simplicity, we can leave them be for now, I guess when we do https://github.com/finos/legend-studio/issues/317 we can remove, `editorMode`, `mode` and `graphState`, for `EditorGraphState` after this change contains little state left, only serving as an organizer of methods

packages/legend-application-studio/src/stores/EditorStore.ts
```diff
@@ -1187,11 +932,100 @@ export class EditorStore implements CommandRegistrar {
     );
   }
 
-  changeLocalChangesState(): void {
-    if (this.isInFormMode) {
-      this.localChangesState = new FormLocalChangesState(this, this.sdlcState);
-    } else {
-      this.localChangesState = new TextLocalChangesState(this, this.sdlcState);
+  *switchModes(
```
akphi, line 935: What do we decide to do with this function, or are you still in the midst of addressing the comments?

akphi (changes requested):

packages/legend-application-studio/src/stores/GraphEditModeManagerState.ts
```diff
@@ -0,0 +1,1063 @@
...
+  graph_addElement,
+  graph_deleteElement,
+  graph_deleteOwnElement,
+  graph_dispose,
+  graph_renameElement,
+} from './shared/modifier/GraphModifierHelper.js';
+import { ElementEditorState } from './editor-state/element-editor-state/ElementEditorState.js';
+import { LegendStudioTelemetry } from './LegendStudioTelemetry.js';
+
+export abstract class GraphEditorMode {
```
akphi, line 70: I think it's better now to split this into 3 files

packages/legend-application-studio/src/stores/GraphEditModeManagerState.ts
```diff
@@ -0,0 +1,1063 @@
...
+            label: 'Stay',
+            default: true,
+            type: ActionAlertActionType.PROCEED,
+          },
+        ],
+      });
+    }
+  }
+
+  *cleanupBeforeEntering(fallbackOptions?: {
```
akphi, line 1047: why is this `fallbackOptions` not being shown as part of the `abstract` class contract?

packages/legend-application-studio/src/stores/EditorStore.ts
```diff
@@ -1187,11 +922,77 @@ export class EditorStore implements CommandRegistrar {
...
+        break;
+      }
+      case GRAPH_EDITOR_MODE.FORM: {
+        if (this.graphState.checkIfApplicationUpdateOperationIsRunning()) {
+          return;
+        }
+        try {
+          try {
+            yield flowResult(
+              this.getGraphEditorMode(
```
akphi, line 960: something here is still not right to me, I want the treatement here to be very simple and symmetrical. Say every time one changes the mode, we do:

```ts
const newMode = new ...();

// is this needed?
try {
  // prepare new state
  newMode.prepare(); // prepare the material needed to initialize() new state, or the necessary checks, if this fail, we do not allow leaving current mode since new state is not possible
} catch (err) {
  ... // if failure here, we probably should abort leaving mode?
}

try {
  currentMode.leave() // clean up --> this implies we need to set some state somewhere before starting a new mode, we should really minimize the logic here and offload to `initialize` as much as possible, also, there should not be failure happening
} catch (err) {
  ... // handle and potentially disallow leaving the mode
}

try {
  newMode.initialize() // enter new mode (bulk of the logic is here)
} catch (err) {
  ... // handle here, if failed to initialize new mode, we need a fall back mechanism to handle
  // maybe do this?
  newMode.fallback();
}
```

I think if we break things down like that, it would be much easier.

packages/legend-application-studio/src/stores/GraphEditModeManagerState.ts
akphi: see my comment in `EditorStore`; that is an idea worths exploring here, I think our interface here is still not generic enough.

akphi (changes requested):

packages/legend-application-studio/src/components/editor/Editor.tsx
```diff
@@ -204,10 +205,13 @@ export const Editor = withEditorStore(
                           : {})}
                       >
                         {(isResolvingConflicts || editable) &&
-                          editorStore.isInFormMode && <EditPanel />}
-                        {editable && editorStore.isInGrammarTextMode && (
-                          <GrammarTextEditor />
-                        )}
+                          editorStore.graphEditorMode.mode ===
```
akphi, line 208: not so satisfied with this, but we can deal with this later.

packages/legend-application-studio/src/components/editor/aux-panel/Problems.tsx
```diff
@@ -64,7 +58,8 @@ const ProblemItem = observer((props: { problem: Problem }) => {
         </div>
         {problem.sourceInformation && (
           <div className="auxiliary-panel__problem__source">
-            {editorStore.isInGrammarTextMode &&
+            {editorStore.graphEditorMode.mode ===
```
akphi, line 61: potentially if we have a place in rendering code layer to handle differentiating codepath, this is a good candidate to add there, together with all places we have to use `editorStore.graphEditorMode.mode` since although we decompose well at the state layer, our rendering is still messy

packages/legend-application-studio/src/stores/project-viewer/ProjectViewerStore.ts
```diff
@@ -472,33 +472,19 @@ export class ProjectViewerStore {
...
-            (yield this.editorStore.graphManagerState.graphManager.entitiesToPureCode(
-              entities,
-            )) as string;
-          yield flowResult(
-            this.editorStore.grammarTextEditorState.setGraphGrammarText(
-              editorGrammar,
-            ),
-          );
-        } catch {
+        yield flowResult(
```
akphi, line 475: note, here we're making assumption we're already in form mode, what if we're not?

packages/legend-application-studio/src/stores/GraphEditFormModeState.ts
```diff
@@ -0,0 +1,594 @@
...
+  graph_deleteElement,
+  graph_deleteOwnElement,
+  graph_dispose,
+  graph_renameElement,
+} from './shared/modifier/GraphModifierHelper.js';
+import { ElementEditorState } from './editor-state/element-editor-state/ElementEditorState.js';
+import { LegendStudioTelemetry } from './LegendStudioTelemetry.js';
+import { GraphEditorMode } from './GraphEditorMode.js';
+
+export class GraphEditFormModeState extends GraphEditorMode {
```
akphi, line 59: --> `FormGraphEditMode`, also rename the file

packages/legend-application-studio/src/stores/GraphEditSingleFileGrammarModeState.ts
```diff
@@ -0,0 +1,513 @@
...
+
+import { TextLocalChangesState } from './sidebar-state/LocalChangesState.js';
+import { GraphCompilationOutcome, type Problem } from './EditorGraphState.js';
+import { GRAPH_EDITOR_MODE, AUX_PANEL_MODE } from './EditorConfig.js';
+import { graph_dispose } from './shared/modifier/GraphModifierHelper.js';
+import { LegendStudioTelemetry } from './LegendStudioTelemetry.js';
+import { GraphEditorMode } from './GraphEditorMode.js';
+import { ElementEditorState } from './editor-state/element-editor-state/ElementEditorState.js';
+
+export class GraphEditSingleFileGrammarModeState extends GraphEditorMode {
```
akphi, line 51: --> `SingleFileGrammarGraphEditMode` (also rename the file)

packages/legend-application-studio/src/components/project-viewer/ProjectViewer.tsx
```diff
@@ -329,8 +330,10 @@ export const ProjectViewer = withEditorStore(
                       {...sideBarCollapsiblePanelGroupProps.remainingPanel}
                       minSize={300}
                     >
-                      {editorStore.isInFormMode && <EditPanel />}
-                      {editorStore.isInGrammarTextMode && <GrammarTextEditor />}
+                      {editorStore.graphEditorMode.mode ===
```
akphi, line 333: I'm thinking should we shift our thinking here a bit, technically, everything is still `EditPanel`, but we have 1 tab (non-closable) instead. What do you think? We can do the same for `Editor.tsx`, this is a clean approach since it's purely `state-driven`

packages/legend-application-studio/src/stores/EditorStore.ts
akphi: probably can move these 2 methods to inside of `EditorGraphState` now

packages/legend-application-studio/src/stores/EditorStore.ts
```diff
@@ -1187,11 +894,73 @@ export class EditorStore implements CommandRegistrar {
     );
   }
```
akphi, line 896: probably can move these 2 methods to inside of `EditorGraphState` now

packages/legend-application-studio/src/stores/EditorStore.ts
```diff
@@ -169,8 +134,8 @@ export class EditorStore implements CommandRegistrar {
   sdlcState: EditorSDLCState;
   graphState: EditorGraphState;
   graphManagerState: GraphManagerState;
```
akphi, line 136: Hm, in terms of organization at `EditorStore` level, I think we should move `EditorStore.graphEditorMode` to `EditorStore.graphState.mode`, I want us to start thinking about how we organize this. Maybe, what we should do is extending `GraphManagerState` and have `LegendStudioGraphManagerState` or something and put all graph handling logic of `EditorGraphState` in there, but colocating `GraphEditorMode` with `EditorGraphState` is sensible as it is what we had before

packages/legend-application-studio/src/stores/EditorStore.ts
```diff
@@ -1187,11 +894,73 @@ export class EditorStore implements CommandRegistrar {
...
+      isCompilationFailure?: boolean;
+      isGraphBuildFailure?: boolean;
+    },
+  ): GeneratorFn<void> {
+    switch (to) {
+      case GRAPH_EDITOR_MODE.GRAMMAR_TEXT: {
+        const graphEditorMode = new GraphEditSingleFileGrammarModeState(this);
+        try {
+          yield flowResult(
+            graphEditorMode.cleanupBeforeEntering(fallbackOptions),
```
akphi, line 909: why are we still calling different set of methods when switching mode? I felt like we can also following a fix procedure, i.e. simplifying this down to bare-bone mode-independent logic, i.e.

```plaintext
attempt to leave and cleanup current mode
check if we can enter target mode
initialize target mode
switch to new mode
```

packages/legend-application-studio/src/stores/EditorStore.ts
```diff
@@ -1187,11 +894,73 @@ export class EditorStore implements CommandRegistrar {
...
   }
 
-  changeLocalChangesState(): void {
-    if (this.isInFormMode) {
-      this.localChangesState = new FormLocalChangesState(this, this.sdlcState);
-    } else {
-      this.localChangesState = new TextLocalChangesState(this, this.sdlcState);
+  *switchModes(
+    to: GRAPH_EDITOR_MODE,
+    fallbackOptions?: {
```
akphi, line 899: I guess we can sleep on this for now, worst case, we have different set of methods: `handleGraphBuilderFailure()` and `handleCompilationFailure()`

packages/legend-application-studio/src/stores/EditorGraphState.ts
```diff
@@ -137,49 +118,39 @@ export class EditorGraphState {
   isInitializingGraph = false;
   isRunningGlobalCompile = false;
   isRunningGlobalGenerate = false;
-  isApplicationLeavingTextMode = false;
+  isApplicationLeavingGraphEditMode = false;
```
akphi, line 121: --> `isSwitchingMode`

akphi (changes requested):
left some minor comments, like last time, the one thing that gates me from fully accepting this change is the complexity in `switchMode()`; like any function of this type, if we can break it into consistent, simple interface function call that would be great, from my perspective, changing mode is an operation that is used by higher-level code that has no knowledge of the mode implementation, what we're doing here is practically leaking of mode-specific logic.

Now, the only 3 places we should allow this sort of thing to me are:
1. When switching mode since we have to explicitly mention the mode to switch to
2. When displaying the mode switcher button (maybe, right now we have binary-state button, but if we have a dropdown or selector, this can even go away)
3. When we initialize the mode in `EditorStore` or so, we are biased since we have to pick a default

packages/legend-application-studio/src/components/editor/Editor.tsx
akphi: see my comment below for `ProjectViewer`

packages/legend-application-studio/src/stores/project-viewer/ProjectViewerStore.ts
gayathrir11: this function gets called only when we initialize studio. So the mode we are in would be FORM

akphi (approved):
Like said, let's address comments here in a later PR, the organization benefit we got out of this PR is already huge. Thanks @gayathrir11!!

PR #1909, opened 2023-02-14
Only add properties with unique name and path in query builder

### LINKED ISSUE #1853: Bug: Query Parameters/constants windows flickering and inconsistent styling
### Similar issues

### How are you using Studio?

Legend Query

When in the query editor at a certain size the panels for Paramaters and Constants flicker constantly 

Kapture 2023-01-25 at 10 44 56

Other issues: 

- [x] Upon leaving text mode in query and clicking close, user gets no feedback from the screen so there should be a loading indicator of some sort so user does not repeatedly click. 
- [ ] When you right click 'add properties' in the class explorer tree to fetch structure, it should only check if there have not been existing columns with the same name+path and if so, do not automatically add them.
- [x] There should be a clear option (brush icon) for users to remove all columns in the fetch structure 
- [ ] When the constant is a date type the path displayed should change to just the name of the type of date. For example `meta::pure::functions::date::adjust`` should just be `yesterday`. 
Original: 
image
New:
image

### Steps to reproduce

_No response_

### Model data

_No response_

### Environment

_No response_

_No response_

### Contribution

xannem:
fixes #1853 

## Summary

- [x] Only add properties with unique name and path in query builder
Kapture 2023-02-14 at 10 36 12
- [x] Replace 'brush' icon with 'trash' icon for clarity 
image
- [x]  Allow humanize property of ID to be kept during prettify process

- [x] Test(s) added

### DIFF
--- a/packages/legend-query-builder/src/components/explorer/QueryBuilderExplorerPanel.tsx
+++ b/packages/legend-query-builder/src/components/explorer/QueryBuilderExplorerPanel.tsx
@@ -62,12 +62,16 @@ import {
   QueryBuilderExplorerTreeSubTypeNodeData,
   getQueryBuilderPropertyNodeData,
   getQueryBuilderSubTypeNodeData,
+  buildPropertyExpressionFromExplorerTreeNodeData,
 } from '../../stores/explorer/QueryBuilderExplorerState.js';
 import { useDrag } from 'react-dnd';
 import { QueryBuilderPropertyInfoTooltip } from '../shared/QueryBuilderPropertyInfoTooltip.js';
 import type { QueryBuilderState } from '../../stores/QueryBuilderState.js';
 import { flowResult } from 'mobx';
-import { prettyPropertyName } from '../../stores/QueryBuilderPropertyEditorState.js';
+import {
+  getPropertyChainName,
+  prettyPropertyName,
+} from '../../stores/QueryBuilderPropertyEditorState.js';
 import {
   type Type,
   type Multiplicity,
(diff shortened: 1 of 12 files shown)


YannanGao-gs (approved):

YannanGao-gs:
Idk if you need to add test for filtering node

packages/legend-shared/src/format/FormatterUtils.ts
```diff
@@ -56,6 +56,19 @@ export const TITLE_CASE_EXCEPTION_WORDS = [
   'up',
 ];
 
+export const prettyCONSTName = (value: string | undefined): string => {
+  if (!value) {
+    return '';
+  }
+  const containsId = value.slice(-2) === 'ID' || value.slice(-2) === 'Id';
```
MauricioUyaguari, line 63: lets move `ID` and `Id` to const. 
lets also move out `value.slice(-2)` as like valueSuffix before comparing to make the code easier to read

packages/legend-query-builder/src/components/explorer/QueryBuilderExplorerPanel.tsx
```diff
@@ -298,7 +304,36 @@ const QueryBuilderExplorerContextMenu = observer(
...
+            const propertyExpression =
+              buildPropertyExpressionFromExplorerTreeNodeData(
+                childNode,
+                queryBuilderState.explorerState,
+              );
+            if (
+              queryBuilderState.fetchStructureState.implementation instanceof
+              QueryBuilderTDSState
+            ) {
+              const currentProjectColumns =
```
MauricioUyaguari, line 318: can't you use the `.find` function instead of map and if the projection column is found to match your condition then you filter that out. Prevents you from having to use `.every` and going through all the projection columns.

MauricioUyaguari (changes requested):

packages/legend-query-builder/src/components/explorer/QueryBuilderExplorerPanel.tsx
xannem: True!

MauricioUyaguari (approved):

PR #1923, opened 2023-02-16
add more telemetry metrics

akphi:
- [x] bump dependencies
- [x] Report telemetry metrics:
  - [x] Legend Query: Create query
  - [x] Legend Query: Update/Save query


packages/legend-application-query/src/stores/QueryEditorStore.ts
```diff
@@ -246,6 +261,20 @@ export class QueryExportState {
         this.editorStore.applicationStore.notifySuccess(
           `Successfully updated query!`,
         );
+
+        LegendQueryTelemetry.logEvent_UpdateQuery(
```
MauricioUyaguari, line 265: @akphi
Right now this gives us created and updated numbers, but this could potential double count queries right. Should we also add a queryable way to do this from the server side to count the number of current saved queries?

PR #1916, opened 2023-02-15
Query builder fixes

xannem:
## Summary
- [x] Displays query title in query builder on top left corner of the window
Kapture 2023-02-15 at 15 36 12
image

- [x] When in the derived property editor, parameters that display should match the type of the derived property otherwise they should be hidden since the user will not be able to drag and drop the parameter to the derived property value. 
Kapture 2023-02-15 at 12 22 58

- [x] If there is already a created constant with the same variable name, show a validation error for user when they try to name the parameter. 
doNotAllowParameterDuplicateName

- [x] If the constant type is a date, rather than the label showing previously just `meta::pure::functions::date::adjust` a more descriptive title should appear for users. 

Originally: 
image

Now:
image

- [x] Styling fix to show that the 'export' button in query builder is able to be pressed when previously it appeared the same whether disabled or active.

Originally: 
image

Now:
image

## How did you test this change?

- [ ] Test(s) added
- [x] Manual testing (please provide screenshots/recordings)
- [ ] No testing (please provide an explanation)

PR #1934, opened 2023-02-20
Add tests for expected query builder behavior

xannem:
## Summary

Adds tests for #1916 and #1911 

- [x] Ascertains query builder sort icon in the project column has a label 
- [x] Shows humanized label for constants when user selects a constant with `Date` type `today`. 
- [x] Shows warning that Parameter Name Already Exists if constant with the same variable name already exists, and vice versa 

- [x] Test(s) added

### DIFF
--- a/packages/legend-query-builder/src/components/__tests__/QueryBuilderVariables.test.tsx
+++ b/packages/legend-query-builder/src/components/__tests__/QueryBuilderVariables.test.tsx
@@ -1,0 +1,208 @@
+/**
+ * Copyright (c) 2020-present, Goldman Sachs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { test, expect } from '@jest/globals';
+import {
+  waitFor,
+  fireEvent,
+  getByTitle,
+  act,
+  getByText,
+} from '@testing-library/react';
+import { TEST_DATA__simpleProjection } from '../../stores/__tests__/TEST_DATA__QueryBuilder_Generic.js';
+import { TEST_DATA__ModelCoverageAnalysisResult_ComplexRelational } from '../../stores/__tests__/TEST_DATA__ModelCoverageAnalysisResult.js';
+import TEST_DATA__ComplexRelationalModel from '../../stores/__tests__/TEST_DATA__QueryBuilder_Model_ComplexRelational.json';
+import { integrationTest } from '@finos/legend-shared';
+import {
+  create_RawLambda,
+  PrimitiveType,
+  stub_RawLambda,
+} from '@finos/legend-graph';
+import { QUERY_BUILDER_TEST_ID } from '../QueryBuilder_TestID.js';
+import { TEST__setUpQueryBuilder } from '../QueryBuilderComponentTestUtils.js';
+import { CUSTOM_DATE_PICKER_OPTION } from '../shared/CustomDatePicker.js';
+
+test(
+  integrationTest(
+    'Query builder shows validation error for constant name if existing duplicate parameter name',
+  ),
+  async () => {
+    const { renderResult, queryBuilderState } = await TEST__setUpQueryBuilder(
+      TEST_DATA__ComplexRelationalModel,
+      stub_RawLambda(),
+      'model::relational::tests::simpleRelationalMapping',
+      'model::MyRuntime',
+      TEST_DATA__ModelCoverageAnalysisResult_ComplexRelational,
+    );
+    await act(async () => {
+      queryBuilderState.initializeWithQuery(
+        create_RawLambda(
+          TEST_DATA__simpleProjection.parameters,
+          TEST_DATA__simpleProjection.body,
+        ),
+      );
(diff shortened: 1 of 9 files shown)


packages/legend-query-builder/src/components/__tests__/QueryBuilderVariables.test.tsx
```diff
@@ -0,0 +1,228 @@
...
+      TEST_DATA__ModelCoverageAnalysisResult_ComplexRelational,
+    );
+    await act(async () => {
+      queryBuilderState.initializeWithQuery(
+        create_RawLambda(
+          TEST_DATA__simpleProjection.parameters,
+          TEST_DATA__simpleProjection.body,
+        ),
+      );
+      queryBuilderState.setShowParametersPanel(true);
```
MauricioUyaguari, line 58: Can we show parameter/constant panel mirroring how the user would do it. 
I.e go to advanced options -> click show parameters/constants. 
We can even move those out to helper query builder test functions

packages/legend-query-builder/src/components/__tests__/QueryBuilderVariables.test.tsx
```diff
@@ -0,0 +1,228 @@
...
+      QUERY_BUILDER_TEST_ID.QUERY_BUILDER_CONSTANTS,
+    );
+    fireEvent.click(getByTitle(constantsPanel, 'Add Constant'));
+    await act(async () => {
+      if (!queryBuilderState.constantState.selectedConstant) {
+        return;
+      }
+      const varExpression =
+        queryBuilderState.constantState.selectedConstant.variable;
+      variableExpression_setName(varExpression, 'var_1');
```
MauricioUyaguari, line 75: Same here, can we change the input value of the input box rather than changing the state directly.

packages/legend-query-builder/src/components/__tests__/QueryBuilderVariables.test.tsx
```diff
@@ -0,0 +1,228 @@
...
+    await waitFor(() =>
+      renderResult.getByTestId(QUERY_BUILDER_TEST_ID.QUERY_BUILDER_PARAMETERS),
+    );
+    const parametersPanel = renderResult.getByTestId(
+      QUERY_BUILDER_TEST_ID.QUERY_BUILDER_PARAMETERS,
+    );
+
+    fireEvent.click(getByTitle(parametersPanel, 'Add Parameter'));
+
+    expect(
```
MauricioUyaguari, line 89: this is actually a bug. when we generate the name for parameters we check existing names to determine what we generate. We probably don't check include constants in this check. Could we fix that. and then for this test we should write out the duplicate and see the error

packages/legend-query-builder/src/components/__tests__/QueryBuilderVariables.test.tsx
```diff
@@ -0,0 +1,228 @@
...
+    fireEvent.click(getByTitle(constantsPanel, 'Add Constant'));
+
+    await act(async () => {
+      if (!queryBuilderState.constantState.selectedConstant) {
+        return;
+      }
+
+      const varExpression =
+        queryBuilderState.constantState.selectedConstant.variable;
+      variableExpression_setName(varExpression, 'var_1');
```
MauricioUyaguari, line 153: same comments as above

packages/legend-query-builder/src/components/__tests__/QueryBuilderVariables.test.tsx
```diff
@@ -0,0 +1,228 @@
...
+    const constantsPanel = renderResult.getByTestId(
+      QUERY_BUILDER_TEST_ID.QUERY_BUILDER_CONSTANTS,
+    );
+    fireEvent.click(getByTitle(constantsPanel, 'Add Constant'));
+
+    await act(async () => {
+      if (!queryBuilderState.constantState.selectedConstant) {
+        return;
+      }
+      queryBuilderState.constantState.selectedConstant.changeValSpecType(
```
MauricioUyaguari, line 202: can we pick the strict date as a user would from the dropdown.

MauricioUyaguari (changes requested):

packages/legend-query-builder/src/components/__tests__/QueryBuilderVariables.test.tsx
xannem: I think we don't check because we don't need to since when we call `generateEnumerableNameFromToken` for constant our prefix we put in front is `c_var` while for parameters it's `var` so there is never a conflict between c_var_1,etc. and var_1, etc. That's why I changed the constant name directly to be var_1 to cause this error

packages/legend-query-builder/src/components/__tests__/QueryBuilderVariables.test.tsx
xannem: Added a note

packages/legend-query-builder/src/components/__tests__/QueryBuilderVariables.test.tsx
xannem: I believe, same issue with accessing dropdown text

packages/legend-query-builder/src/components/__tests__/QueryBuilderVariables.test.tsx
```diff
@@ -0,0 +1,208 @@
...
+      TEST_DATA__ModelCoverageAnalysisResult_ComplexRelational,
+    );
+    await act(async () => {
+      queryBuilderState.initializeWithQuery(
+        create_RawLambda(
+          TEST_DATA__simpleProjection.parameters,
+          TEST_DATA__simpleProjection.body,
+        ),
+      );
+      //Note: Render result will not currently find the
```
MauricioUyaguari, line 57: indent this comment please. see other notes.

MauricioUyaguari (changes requested):
minor change

MauricioUyaguari (approved):

PR #1941, opened 2023-02-22
Add warning for users changing project's group ID or artifact ID

xannem:
## Summary

Kapture 2023-02-22 at 15 44 48

image

Before, users could just directly click and edit the inputs for group id or artifact id on the project config page. Now, users must click the edit icon to edit, and if they do a caution dialog will appear to make the potential consequences of their actions clearer for them: 

- [ ] Test(s) added

### DIFF
--- a/packages/legend-application-studio/src/components/editor/edit-panel/project-configuration-editor/ProjectConfigurationEditor.tsx
+++ b/packages/legend-application-studio/src/components/editor/edit-panel/project-configuration-editor/ProjectConfigurationEditor.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { useEffect } from 'react';
+import { useEffect, useState } from 'react';
 import { prettyCONSTName } from '@finos/legend-shared';
 import { observer } from 'mobx-react-lite';
 import {
@@ -27,11 +27,13 @@ import {
   CheckCircleIcon,
   ExclamationCircleIcon,
   Panel,
+  PencilIcon,
   PanelForm,
-  PanelFormTextField,
   CheckSquareIcon,
   SquareIcon,
   ExclamationTriangleIcon,
+  PanelFormSection,
+  PanelListItem,
 } from '@finos/legend-art';
 import { flowResult } from 'mobx';
 import {
@@ -42,6 +44,8 @@ import {
 } from '@finos/legend-server-sdlc';
 import { useEditorStore } from '../../EditorStoreProvider.js';
 import {
+  ActionAlertActionType,
+  ActionAlertType,
   DocumentationPreview,
   LEGEND_APPLICATION_DOCUMENTATION_KEY,
   useApplicationStore,
@@ -67,6 +71,47 @@ const ProjectStructureEditor = observer(
       flowResult(
         editorStore.projectConfigurationEditorState.updateToLatestStructure(),
       ).catch(applicationStore.alertUnhandledError);
+    };
+
+    const changeGroupId: React.ChangeEventHandler<HTMLInputElement> = (
+      event,
+    ) => {
+      const stringValue = event.target.value;
+      projectConfig.setGroupId(stringValue);
+    };
+
+    const changeArtifactId: React.ChangeEventHandler<HTMLInputElement> = (
+      event,
+    ) => {
+      const stringValue = event.target.value;
+      projectConfig.setArtifactId(stringValue);
+    };
+    const [blockingEditGav, setBlockingEditGav] = useState(true);
+
+    const allowEditingGav = (): void => {
+      if (blockingEditGav) {
+        editorStore.applicationStore.setActionAlertInfo({
+          message:
+            'If you are editing the group ID or artifact ID (gav), please keep in mind this means you are creating a new project which means that you will lose any previous project versions. Moreover, if your current project has dependencies, you will lose those dependencies until you release the version of your new project with its updated gav.',
+          type: ActionAlertType.CAUTION,
+          actions: [
+            {
+              label: 'Acknowledge and Proceed',
+              type: ActionAlertActionType.PROCEED_WITH_CAUTION,
+              handler: (): void => {
+                setBlockingEditGav(false);
+              },
+            },
+            {
+              label: 'Cancel',
+              type: ActionAlertActionType.PROCEED,
+              default: true,
+            },
+          ],
+        });
+      } else {
+        setBlockingEditGav(true);
+      }
     };
 
     return (
@@ -105,27 +150,87 @@ const ProjectStructureEditor = observer(
         </div>
 
         <PanelForm>
-          <PanelFormTextField
-            isReadOnly={isReadOnly}
-            value={projectConfig.groupId}
-            name="Group ID"
-            prompt="The domain for artifacts generated as part of the project build
-              pipeline and published to an artifact repository"
-            update={(value: string | undefined): void =>
-              projectConfig.setGroupId(value ?? '')
-            }
-          />
-          <PanelFormTextField
-            isReadOnly={isReadOnly}
-            value={projectConfig.artifactId}
-            name="Artifact ID"
-            prompt="The identifier (within the domain specified by group ID) for
+          <PanelFormSection>
+            <div className="panel__content__form__section__header__label">
+              Group ID
+            </div>
+            <div className="panel__content__form__section__header__prompt">
+              The domain for artifacts generated as part of the project build
+              pipeline and published to an artifact repository
+            </div>
+            <PanelListItem>
+              <div className="input-group">
+                <input
+                  className={clsx(
+                    'input input--dark input-group__input panel__content__form__section__input input--large',
+                  )}
+                  title={
+                    blockingEditGav && !isReadOnly
+                      ? 'Click pencil icon to edit group ID'
+                      : 'Group ID'
+                  }
+                  spellCheck={false}
+                  disabled={isReadOnly || blockingEditGav}
+                  value={projectConfig.groupId}
+                  onChange={changeGroupId}
+                />
+              </div>
+
+              <button
+                className={clsx(
+                  'panel__content__form__section__list__item__edit-btn',
+                  { 'icon--active': !blockingEditGav },
+                )}
+                disabled={isReadOnly}
+                onClick={() => allowEditingGav()}
+                tabIndex={-1}
+                title="Edit group ID"
+              >
+                <PencilIcon />
+              </button>
+            </PanelListItem>
+          </PanelFormSection>
+          <PanelFormSection>
+            <div className="panel__content__form__section__header__label">
+              Artifact ID
+            </div>
+            <div className="panel__content__form__section__header__prompt">
+              The identifier (within the domain specified by group ID) for
               artifacts generated as part of the project build pipeline and
-              published to an artifact repository"
-            update={(value: string | undefined): void =>
-              projectConfig.setArtifactId(value ?? '')
-            }
-          />
+              published to an artifact repository
+            </div>
+            <PanelListItem>
+              <div className="input-group">
+                <input
+                  className={clsx(
+                    'input input--dark input-group__input panel__content__form__section__input input--large',
+                  )}
+                  title={
+                    blockingEditGav && !isReadOnly
+                      ? 'Click pencil icon to edit artifact ID'
+                      : 'Artifact ID'
+                  }
+                  spellCheck={false}
+                  disabled={isReadOnly || blockingEditGav}
+                  value={projectConfig.artifactId}
+                  onChange={changeArtifactId}
+                />
+              </div>
+
+              <button
+                className={clsx(
+                  'panel__content__form__section__list__item__edit-btn',
+                  { 'icon--active': !blockingEditGav },
+                )}
+                disabled={isReadOnly}
+                title="Edit artifact ID"
+                onClick={() => allowEditingGav()}
+                tabIndex={-1}
+              >
+                <PencilIcon />
+              </button>
+            </PanelListItem>
+          </PanelFormSection>
         </PanelForm>
       </div>
     );
--- a/packages/legend-art/style/base/_panel.scss
+++ b/packages/legend-art/style/base/_panel.scss
@@ -203,6 +203,10 @@
   .input--small {
     max-width: 45rem;
   }
+
+  .input--large {
+    width: 45rem;
+  }
 }
 
 .panel__list-selector {
--- a/packages/legend-art/style/components/_icon.scss
+++ b/packages/legend-art/style/components/_icon.scss
@@ -24,6 +24,11 @@
   font-family: Raleway, sans-serif;
   font-weight: 900;
   user-select: none;
+
+  &--active svg,
+  &--active svg:hover {
+    color: var(--color-yellow-100);
+  }
 }
 
 .icon--flat-data {
--- a/.changeset/great-apricots-kick.md
+++ b/.changeset/great-apricots-kick.md
@@ -1,0 +1,5 @@
+---
+'@finos/legend-application-studio': patch
+---
+
+Add warning for users changing project's group ID or artifact ID
--- a/.changeset/plentiful-pigeons-pillage.md
+++ b/.changeset/plentiful-pigeons-pillage.md
@@ -1,0 +1,3 @@
+---
+'@finos/legend-art': patch
+---


packages/legend-application-studio/src/components/editor/edit-panel/project-configuration-editor/ProjectConfigurationEditor.tsx
```diff
@@ -105,27 +154,87 @@ const ProjectStructureEditor = observer(
...
-            update={(value: string | undefined): void =>
-              projectConfig.setGroupId(value ?? '')
-            }
-          />
-          <PanelFormTextField
-            isReadOnly={isReadOnly}
-            value={projectConfig.artifactId}
-            name="Artifact ID"
-            prompt="The identifier (within the domain specified by group ID) for
+          <PanelFormSection>
```
akphi, line 157: in terms of UX, what I think makes sense to me is we do the block on `Update Config` action, this is the time when the change happens. If we need people to be really cautious about their changing the `gav` coordinates, we can have a add warning to the input, with potentially link to documentation

packages/legend-art/style/base/_panel.scss
```diff
@@ -203,6 +203,10 @@
   .input--small {
     max-width: 45rem;
   }
+
+  .input--large {
```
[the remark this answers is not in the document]
xannem: This one is meant to force it to have 45rem since if the element is under <PanelListItem> the width will automatically be small. But I will change naming from `--large` :D

packages/legend-art/style/components/_icon.scss
```diff
@@ -24,6 +24,14 @@
   font-family: Raleway, sans-serif;
   font-weight: 900;
   user-select: none;
+
+  &--left {
```
akphi, line 28: let's remove these

packages/legend-art/style/base/_panel.scss
```diff
@@ -203,6 +203,10 @@
   .input--small {
     max-width: 45rem;
   }
+
+  .input--small--set {
```
akphi, line 207: I'm as hesitant to add this class as much as the struggle you have to name it, so maybe let's try something like setting the input field to `width: 100%` and used together with `input--small`

packages/legend-application/src/components/ActionAlert.tsx
```diff
@@ -93,7 +93,7 @@ const ActionAlertContent = observer((props: { info: ActionAlertInfo }) => {
                 autoFocus={Boolean(action.default)}
                 // since this is a text button, no need for tooltip
               >
-                {action.label}
+                {action.label} {action.content}
```
akphi, line 96: like said, let's not do this.

packages/legend-application-studio/src/components/editor/edit-panel/project-configuration-editor/ProjectConfigurationEditor.tsx
```diff
@@ -360,13 +438,106 @@ export const ProjectConfigurationEditor = observer(() => {
...
-      flowResult(configState.updateConfigs()).catch(
-        applicationStore.alertUnhandledError,
+
+  const updateGavConfigs = (): void => {
+    const documentationEntry =
+      applicationStore.documentationService.getDocEntry(
+        LEGEND_STUDIO_DOCUMENTATION_KEY.QUESTION_HOW_TO_UPDATE_PROJECT_GAV_COORDINATES,
       );
+
+    const actions = documentationEntry
```
akphi, line 448: a little too verbose, I think we can do like before and for the list of actions, we can add a condition in front of it and filter out by `.filter(Boolean)`

packages/legend-application-studio/src/components/editor/edit-panel/project-configuration-editor/ProjectConfigurationEditor.tsx
```diff
@@ -69,6 +76,32 @@ const ProjectStructureEditor = observer(
...
+    };
+
+    const changeArtifactId: React.ChangeEventHandler<HTMLInputElement> = (
+      event,
+    ) => {
+      const stringValue = event.target.value;
+      projectConfig.setArtifactId(stringValue);
+    };
+
+    const changedArtifactId =
```
akphi, line 93:
```suggestion
    const isArtifactIdChanged =
```

packages/legend-application-studio/src/components/editor/edit-panel/project-configuration-editor/ProjectConfigurationEditor.tsx
```diff
@@ -69,6 +76,32 @@ const ProjectStructureEditor = observer(
...
+      projectConfig.artifactId !==
+      editorStore.tabManagerState.getCurrentEditorState(
+        ProjectConfigurationEditorState,
+      ).originalConfig.artifactId;
+
+    const changedGroupId =
```
akphi, line 99:
```suggestion
    const isGroupdIdChanged =
```

packages/legend-application-studio/src/components/editor/edit-panel/project-configuration-editor/ProjectConfigurationEditor.tsx
```diff
@@ -105,27 +138,72 @@ const ProjectStructureEditor = observer(
...
+              pipeline and published to an artifact repository
+            </div>
+            <PanelListItem>
+              <div className="input-group">
+                <input
+                  className={clsx(
+                    'input input--dark input-group__input panel__content__form__section__input input--small--set',
+                    { 'input--caution': changedGroupId },
+                  )}
+                  title={'Group ID'}
```
akphi, line 156:
```suggestion
                  title="Group ID"
```

packages/legend-application-studio/src/components/editor/edit-panel/project-configuration-editor/ProjectConfigurationEditor.tsx
```diff
@@ -105,27 +138,72 @@ const ProjectStructureEditor = observer(
...
+              published to an artifact repository
+            </div>
+            <PanelListItem>
+              <div className="input-group">
+                <input
+                  className={clsx(
+                    'input input--dark input-group__input panel__content__form__section__input input--small--set',
+                    { 'input--caution': changedArtifactId },
+                  )}
+                  title={'Artifact ID'}
```
akphi, line 190:
```suggestion
                  title="Artifact ID"
```

packages/legend-application/src/stores/AlertService.ts
```diff
@@ -29,11 +29,13 @@ export interface ActionAlertInfo {
   title?: string;
   message: string;
   prompt?: string;
+  content?: React.ReactElement;
```
akphi, line 32: we can change `title?: React.ReactNode` right?

packages/legend-application-studio/src/components/editor/edit-panel/project-configuration-editor/ProjectConfigurationEditor.tsx
```diff
@@ -360,13 +438,106 @@ export const ProjectConfigurationEditor = observer(() => {
...
+          {
+            label: 'Cancel',
+            type: ActionAlertActionType.PROCEED,
+            default: true,
+          },
+        ];
+
+    editorStore.applicationStore.setActionAlertInfo({
+      message:
+        'Please read these instructions before you edit groupid or artifactid (gav) as changing them can have downstream impact. Be aware you will lose any previous project versions. Moreover, if your current project has dependant projects you can break those too if you do not change the gav in a controlled way.',
```
akphi, line 513:
```suggestion
        'Please be cautious that modifying group ID or artifact ID (GAV coordinates) can potentially have a big downstream impact. Be aware that the project will lose all previous versions; also, any dependant projects can break too if the coordinates are not changed in a controlled way',
```

akphi: we should also add a prompt here: if `Boolean(documentationEntry)` we can show: `Please see the instructions for more guidance`

packages/legend-application-studio/src/components/editor/edit-panel/project-configuration-editor/ProjectConfigurationEditor.tsx
```diff
@@ -360,13 +438,106 @@ export const ProjectConfigurationEditor = observer(() => {
...
+      type: ActionAlertType.CAUTION,
+      actions: actions,
     });
   };
+  const updateConfigs = (): void => {
+    const changedArtifactId =
```
akphi, line 519: why do we repeat these checks here instead of just using them from the component variables? you can potentially make this a `computed` getter in the `ProjectConfigurationEditorState`

akphi (changes requested):

akphi:
@xannem by the way, for the input warning, should we add an exclamation point on the right inside the input box, wdyt?

packages/legend-art/style/base/_panel.scss
xannem: Yes width:100% will still not change input and keep it at min length since its inside PanelListItem component so that both items in one row, but I added it as `&--full` hopefully that makes more sense?

packages/legend-application/src/stores/AlertService.ts
xannem: You mean changing label to be `React.ReactNode`? Title belongs to modal header I believe. I was thinking if we change label though, almost every existing label will have to be text changed to `label: <>text here</>` which seems less readable because in almost every case label now and later may be string.

xannem:
> @xannem by the way, for the input warning, should we add an exclamation point on the right inside the input box, wdyt?

I feel like it might be a bit busy if there is exclamation point in input warning since we already have documentation icon that pops up to the right of the input box so then there would be two icon-like items here if that's what you meant

image

I could switch the question icon to use exclamation icon instead though?

akphi:
> > @xannem by the way, for the input warning, should we add an exclamation point on the right inside the input box, wdyt?
> 
> I feel like it might be a bit busy if there is exclamation point in input warning since we already have documentation icon that pops up to the right of the input box so then there would be two icon-like items here if that's what you meant
> 
> image
> 
> I could switch the question icon to use exclamation icon instead though?

Yep agree, looks too cluttered

packages/legend-application-studio/src/components/editor/edit-panel/project-configuration-editor/ProjectConfigurationEditor.tsx
```diff
@@ -360,13 +430,97 @@ export const ProjectConfigurationEditor = observer(() => {
...
-      flowResult(configState.updateConfigs()).catch(
-        applicationStore.alertUnhandledError,
+
+  const updateGavConfigs = (): void => {
+    const documentationEntry =
+      applicationStore.documentationService.getDocEntry(
+        LEGEND_STUDIO_DOCUMENTATION_KEY.QUESTION_HOW_TO_DEFINE_A_CONSTRAINT,
       );
+
+    const actions = [
```
akphi, line 440: let's inline this

packages/legend-art/style/base/_button.scss
```diff
@@ -58,6 +58,14 @@
   @include flexCenter;
 }
 
+.btn--right {
```
akphi, line 61: can we not do this? this is not generic style

packages/legend-application-studio/src/components/editor/edit-panel/project-configuration-editor/ProjectConfigurationEditor.tsx
```diff
@@ -105,27 +128,74 @@ const ProjectStructureEditor = observer(
...
+                  disabled={isReadOnly}
+                  value={projectConfig.groupId}
+                  onChange={changeGroupId}
+                />
+
+                {isProjectGavChanged && (
+                  <DocumentationLink
+                    className="panel__content__form__section__list__item__edit btn--right"
+                    documentationKey={
+                      // QUESTION_HOW_TO_DEFINE_A_CONSTRAINT
```
akphi, line 157: can we remove these comments?

packages/legend-application-studio/src/components/editor/edit-panel/project-configuration-editor/ProjectConfigurationEditor.tsx
```diff
@@ -360,13 +430,97 @@ export const ProjectConfigurationEditor = observer(() => {
...
+        handler: (): void => {
+          editorStore.localChangesState.alertUnsavedChanges((): void => {
+            flowResult(configState.updateConfigs()).catch(
+              applicationStore.alertUnhandledError,
+            );
+          });
+        },
+      },
+      documentationEntry && {
+        label: <>View Instructions</>,
```
akphi, line 481: you don't need `<>` around, let's use `string`

packages/legend-application-studio/src/components/editor/edit-panel/project-configuration-editor/ProjectConfigurationEditor.tsx
```diff
@@ -360,13 +430,97 @@ export const ProjectConfigurationEditor = observer(() => {
...
+      applicationStore.documentationService.getDocEntry(
+        LEGEND_STUDIO_DOCUMENTATION_KEY.QUESTION_HOW_TO_DEFINE_A_CONSTRAINT,
       );
+
+    const actions = [
+      {
+        label: (
+          <>
+            Acknowledge and Proceed
+            {documentationEntry && (
```
akphi, line 445: why don't just use the `DocumentationLink` component?

packages/legend-art/style/base/_panel.scss
```diff
@@ -202,6 +202,10 @@
 
   .input--small {
     max-width: 45rem;
+
+    &--full {
```
akphi, line 206: looks confusing, why not just have `.input--full { width: 100% }`?

akphi (changes requested):

akphi (changes requested):
just minor changes

packages/legend-application-studio/src/components/editor/edit-panel/project-configuration-editor/ProjectConfigurationEditor.tsx
xannem: It's for the `event.preventDefault();` line, since I wasn't sure how we could pass in an event within the `DocumentationLink` component

packages/legend-application-studio/src/components/editor/edit-panel/project-configuration-editor/ProjectConfigurationEditor.tsx
```diff
@@ -360,13 +431,81 @@ export const ProjectConfigurationEditor = observer(() => {
...
+              applicationStore.navigator.visitAddress(documentationEntry.url);
+            }
+          },
+        },
+        {
+          label: 'Cancel',
+          type: ActionAlertActionType.PROCEED,
+        },
+      ].filter(isNonNullable),
+      ...(Boolean(documentationEntry) && {
```
akphi, line 492: let's move this up

packages/legend-application-studio/src/components/editor/edit-panel/project-configuration-editor/ProjectConfigurationEditor.tsx
```diff
@@ -360,13 +431,81 @@ export const ProjectConfigurationEditor = observer(() => {
...
+    editorStore.applicationStore.setActionAlertInfo({
+      message:
+        'Please be cautious that modifying group ID or artifact ID (GAV coordinates) can potentially have a big downstream impact. Be aware that the project will lose all previous versions; also, any dependant projects can break too if the coordinates are not changed in a controlled way.',
+      type: ActionAlertType.STANDARD,
+      actions: [
+        {
+          label: (
+            <>
+              Acknowledge and Proceed
+              {documentationEntry && (
```
akphi, line 450: don't need this check as this is part of the logic of `DocumentationKey`

akphi (approved):

packages/legend-art/style/base/_input.scss
```diff
@@ -55,7 +56,21 @@
...
+  }
+
+  &:focus {
+    border: 0.1rem solid var(--color-yellow-200) !important;
+  }
+}
+
 .input-group {
+  @include flexCenter;
+
```
gayathrir11, line 73: @xannem please be careful while making such changes. Because components in legend-art are used widely in other packages we have to make sure we are not breaking anything existing by making such styling changes. If you want this css at particular editor  I would suggest  you to customize it at editor level.

PR #1968, opened 2023-03-01
Improve ability to use negative numbers in filter & post-filter panel for query builder

### LINKED ISSUE #1980: Query builder filters with negative numbers reset to zero
Currently, when users use the filter panel in the query builder and input negative numbers, although running the query retrieves the correct results at first, the negative number defaults back to '0' when users click 'Edit Query in PURE' or when they save the query and attempt to load it back again. 

This same problem exists for the post-filter panel as well.

https://user-images.githubusercontent.com/73408381/222758891-f3025919-52de-490a-9404-620f4681ce54.mov

Kapture 2023-03-06 at 15 27 29

_Originally posted by @YannanGao-gs in https://github.com/finos/legend-studio/issues/1968#issuecomment-1453693644_

xannem:
## Summary

fixes #1980

Previously, the only way to include a negative number in the filter panel was for users to manually move the cursor to the beginning of the '0' number, press '-', and then move the cursor after the 0, type the number they wanted to be negative, and then delete the 0. This is very unintuitive for the user and a strange method they have to follow in order to write negative numbers. 

To resolve this, we change the input type to be `text` to allow the usage of `.` and `-` to show for the input while the user is typing, but when they click away we change it to be a number to keep our expected behavior and also improve UX for users trying to filter with negative numbers. 

Kapture 2023-03-01 at 15 49 22

Below is a video of the way originally to get negative numbers (pressing the negative sign if you have '0' will result in no activity) 
Kapture 2023-03-01 at 15 57 54

- [ ] Test(s) added

### DIFF
--- a/.changeset/common-starling-yodels.md
+++ b/.changeset/common-starling-yodels.md
@@ -1,0 +1,5 @@
+---
+'@finos/legend-query-builder': patch
+---
+
+Add tests for expected behavior for entry of negative numbers
--- a/packages/legend-graph/src/graph/MetaModelConst.ts
+++ b/packages/legend-graph/src/graph/MetaModelConst.ts
@@ -30,6 +30,7 @@ export const PARSER_SECTION_MARKER = '###';
 export const FUNCTION_SIGNATURE_MULTIPLICITY_INFINITE_TOKEN = 'MANY';
 export const DIRECTORY_PATH_DELIMITER = '/';
 export const SOURCE_INFORMATION_PROPERTY_KEY_SUFFIX = 'sourceInformation';
+export const MINUS_STRING = 'minus';
 
 export const RESERVERD_PACKAGE_NAMES = ['$implicit'];
 
--- a/packages/legend-query-builder/src/components/shared/BasicValueSpecificationEditor.tsx
+++ b/packages/legend-query-builder/src/components/shared/BasicValueSpecificationEditor.tsx
@@ -361,13 +361,67 @@ const NumberPrimitiveInstanceValueEditor = observer(
       resetValue,
       setValueSpecification,
     } = props;
+
+    const DECIMAL_POINT = '.';
+    const NEGATIVE_SIGN = '-';
+
+    // Note: Although we want to allow the user to type
+    // '-' and '.' for negative and decimal numbers respectively,
+    // we want to check for when user leaves the input to not
+    // leave the number value at an invalid state (i.e. if
+    // user just left '-' without an actual number)
+    const onBlur = (): void => {
+      if (valueSpecification.values[0] === NEGATIVE_SIGN) {
+        instanceValue_setValue(valueSpecification, 0, 0);
+        setValueSpecification(valueSpecification);
+      }
+      const strValue = valueSpecification.values[0] as string;
+      const strLength = strValue.length;
+      if (strLength > 0 && strValue.at(-1) === DECIMAL_POINT) {
+        instanceValue_setValue(
+          valueSpecification,
+          strValue.substring(0, strLength - 1),
+          0,
+        );
+        setValueSpecification(valueSpecification);
+      }
+    };
+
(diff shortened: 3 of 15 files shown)


YannanGao-gs:
I think there might be a bug. See video below

https://user-images.githubusercontent.com/73408381/222758891-f3025919-52de-490a-9404-620f4681ce54.mov

MauricioUyaguari:
Tests ? :)

xannem:
> I think there might be a bug. See video below

whoa you're right I checked it this is preexisting thing already from how we handle negative nums in filter good catch I will go add new issue for that

.changeset/common-starling-yodels.md
```diff
@@ -0,0 +1,5 @@
+---
+'@finos/legend-query-builder': patch
+---
+
+Add tests for expected behavior for entry of negative numbers
```
akphi, line 5: don't add changelog when you just add a test please.

packages/legend-graph/src/graph/MetaModelConst.ts
```diff
@@ -30,6 +30,7 @@ export const PARSER_SECTION_MARKER = '###';
 export const FUNCTION_SIGNATURE_MULTIPLICITY_INFINITE_TOKEN = 'MANY';
 export const DIRECTORY_PATH_DELIMITER = '/';
 export const SOURCE_INFORMATION_PROPERTY_KEY_SUFFIX = 'sourceInformation';
+export const MINUS_STRING = 'minus';
```
akphi, line 33: this is handled in query-builder, there's a list of supported functions over there, please move it there accordingly, and it should be called `QUERY_BUILDER_SUPPORTED_FUNCTIONS.MINUS` instead

packages/legend-query-builder/src/components/shared/BasicValueSpecificationEditor.tsx
```diff
@@ -361,13 +361,82 @@ const NumberPrimitiveInstanceValueEditor = observer(
...
+      }
+
       let inputVal = isInteger
         ? parseInt(event.target.value, 10)
         : parseFloat(event.target.value);
       inputVal = isNaN(inputVal) ? 0 : inputVal;
-      instanceValue_setValue(valueSpecification, inputVal, 0);
+
+      valueIsNegativeSymbol
+        ? instanceValue_setValue(valueSpecification, NEGATIVE_SIGN, 0)
```
akphi, line 438: this expression doesn't make any sense to me

packages/legend-query-builder/src/components/shared/BasicValueSpecificationEditor.tsx
```diff
@@ -361,13 +361,82 @@ const NumberPrimitiveInstanceValueEditor = observer(
       resetValue,
       setValueSpecification,
     } = props;
+
```
akphi, line 364: @MauricioUyaguari I'm a little upset how complicated this gets. I think this is not the only place and really the pain is due to us sing mobx to control the graph and the state of the component, maybe we need to think of another way to do this

akphi (changes requested):

packages/legend-query-builder/src/stores/fetch-structure/tds/post-filter/QueryBuilderPostFilterStateBuilder.ts
```diff
@@ -178,6 +179,24 @@ const processPostFilterTree = (
   postFilterState: QueryBuilderPostFilterState,
   parentPostFilterNodeId: string | undefined,
 ): void => {
+  // NOTE: This checks if the expression is a simple function expression of Minus
```
akphi, line 182: Felt like this can be handled at another layer right when we process simple function expression, since here what you're doing is just to take care of them at filter/post-filter level. I'll see what I can do here

akphi (approved):

PR #2005, opened 2023-03-14
Save parameter value state when you switch to text mode and back in query builder

### LINKED ISSUE #1982: Feature request: Save parameter value state when you switch to text mode and back in query builder
### Similar requests

### How are you using Legend?

Legend Studio

When you have parameters in a query and you provided a value for them for execution The state of these parameter values is not preserved when you go to text mode and come back to the query builder state.

We need to preserve the state of paramter values entered by user when they enter text mode and come back to query state

_No response_

_No response_

### Contribution

gayathrir11:
## Summary

closes #1982

- [x] Test(s) added

https://user-images.githubusercontent.com/87973126/224931628-d08fbb9f-4bfc-4dfe-a789-f210ac725a1a.mp4

### DIFF
--- a/packages/legend-art/src/testMocks/MonacoEditorMockUtils.tsx
+++ b/packages/legend-art/src/testMocks/MonacoEditorMockUtils.tsx
@@ -70,6 +70,7 @@ export const MockedMonacoEditorInstance = {
   onDidChangeCursorPosition: createMock(),
   onDidFocusEditorText: createMock(),
   onDidBlurEditorText: createMock(),
+  onDidFocusEditorWidget: createMock(),
 };
 
 export const MockedMonacoEditorAPI = {
@@ -86,6 +87,7 @@ export const MockedMonacoEditorAPI = {
   setModelLanguage: createMock(),
   defineTheme: createMock(),
   EndOfLinePreference: MockedMonacoEditorEndOfLinePreference,
+  removeAllMarkers: createMock(),
 };
 
 export enum MockedMonacoLanguageCompletionItemKind {
--- a/packages/legend-query-builder/src/components/__tests__/QueryBuilderVariables.test.tsx
+++ b/packages/legend-query-builder/src/components/__tests__/QueryBuilderVariables.test.tsx
@@ -28,7 +28,7 @@ import {
 } from '../../stores/__tests__/TEST_DATA__QueryBuilder_Generic.js';
 import { TEST_DATA__ModelCoverageAnalysisResult_ComplexRelational } from '../../stores/__tests__/TEST_DATA__ModelCoverageAnalysisResult.js';
 import TEST_DATA__ComplexRelationalModel from '../../stores/__tests__/TEST_DATA__QueryBuilder_Model_ComplexRelational.json';
-import { integrationTest } from '@finos/legend-shared';
+import { createMock, integrationTest } from '@finos/legend-shared';
 import {
   create_RawLambda,
   PrimitiveType,
@@ -38,6 +38,10 @@ import {
 import { QUERY_BUILDER_TEST_ID } from '../QueryBuilder_TestID.js';
 import { TEST__setUpQueryBuilder } from '../QueryBuilderComponentTestUtils.js';
 import { CUSTOM_DATE_PICKER_OPTION } from '../shared/CustomDatePicker.js';
+import {
+  MockedMonacoEditorInstance,
+  MockedMonacoEditorAPI,
+} from '@finos/legend-art';
 
 test(
   integrationTest(
@@ -307,6 +311,9 @@ test(
       );
     });
     await waitFor(() =>
+      renderResult.getByTestId(QUERY_BUILDER_TEST_ID.QUERY_BUILDER),
+    );
+    await waitFor(() =>
       renderResult.getByTestId(QUERY_BUILDER_TEST_ID.QUERY_BUILDER_PARAMETERS),
     );
     const parameterPanel = renderResult.getByTestId(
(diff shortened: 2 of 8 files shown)


packages/legend-query-builder/src/components/__tests__/QueryBuilderVariables.test.tsx
```diff
@@ -287,3 +287,54 @@ test(integrationTest('Query builder parameter default values'), async () => {
...
+    );
+    expect(getByText(parameterPanel, 'var_1')).not.toBeNull();
+    fireEvent.click(renderResult.getByText('Run Query'));
+    let executeDialog = await waitFor(() => renderResult.getByRole('dialog'));
+    expect(getByText(executeDialog, 'Set Parameter Values'));
+
+    const parameterValue = getByText(executeDialog, 'var_1');
+
+    // Here we mimic the toggling to text mode by calling the rebuildWithQuery function.
+    await act(async () => {
```
xannem, line 323: Would it be possible to mimic clicking the button by title 'View Query in Pure' and then clicking Close button?

packages/legend-query-builder/src/stores/QueryBuilderStateBuilder.ts
```diff
@@ -608,14 +609,26 @@ export const processParameters = (
...
+        checkIfEquivalent(ps.parameter, parameter),
+      ) === undefined
+    ) {
+      const parameterState = new LambdaParameterState(
+        parameter,
+        queryBuilderState.observableContext,
+        queryBuilderState.graphManagerState.graph,
+      );
+      parameterState.mockParameterValue();
+      queryParameterState.addParameter(parameterState);
```
MauricioUyaguari, line 630: We could be altering the query with this logic as we could move around how we add the parameter states to querybuilder.parametersState. 
i.e `test: String[1]. name:String[1]` -> (text mode edit) `test2: string[1], name:String[1]`
With this logic i believe we would rewrite the query parameter to `name: String[1], test2: String[1]` when we change querystate to lambda. 
Can't we just for do the `forEach` on each parameter and if we find a similar one i.e `checkIfEquivalent` in the old state we use that one and add that to the parameter state, if not we create a new one and add?

packages/legend-query-builder/src/components/__tests__/QueryBuilderVariables.test.tsx
MauricioUyaguari: lets try this, we might face issues with Monaco editor and the mocking of functions, if so lets add that to the comments on why we don't mock monaco and the network calls. I imagine something will become too expensive :)

MauricioUyaguari (changes requested):

xannem (approved):

packages/legend-query-builder/src/components/__tests__/QueryBuilderVariables.test.tsx
```diff
@@ -287,3 +291,67 @@ test(integrationTest('Query builder parameter default values'), async () => {
...
+    MockedMonacoEditorInstance.onDidFocusEditorWidget.mockReturnValue(null);
+
+    // Here we mimic the toggling to text mode.
+    fireEvent.click(renderResult.getByTitle('View Query in Pure'));
+    const lambdaEditor = await waitFor(() => renderResult.getByRole('dialog'));
+    const MOCK__globalCompileInFormModeFn = createMock();
+    queryBuilderState.graphManagerState.graphManager.pureCodeToLambda =
+      MOCK__globalCompileInFormModeFn;
+    MOCK__globalCompileInFormModeFn.mockResolvedValue(
+      create_RawLambda(param1Lambda.parameters, param1Lambda.body),
```
MauricioUyaguari, line 342: should we enhance the test to actually change the lambda parameters to mimic the change we are doing here ?

MauricioUyaguari (commented):
will merge but can you look into enhancing the test a bit more. Thank!

MauricioUyaguari (approved):

packages/legend-art/src/testMocks/MonacoEditorMockUtils.tsx
```diff
@@ -86,6 +87,7 @@ export const MockedMonacoEditorAPI = {
   setModelLanguage: createMock(),
   defineTheme: createMock(),
   EndOfLinePreference: MockedMonacoEditorEndOfLinePreference,
+  removeAllMarkers: createMock(),
```
akphi, line 90: why are we changing this, is this problem with rebasing?

packages/legend-query-builder/src/stores/QueryBuilderState.ts
```diff
@@ -301,7 +303,7 @@ export abstract class QueryBuilderState implements CommandRegistrar {
     this.resultState = resultState;
   }
 
-  resetQueryContent(): void {
+  resetQueryBodyContent(): void {
```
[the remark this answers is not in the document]
akphi: As discussed, let's revert most of the changes, (also remember to revert the `readonly` changes you made up the top of this file). We want to clone the param state (which you did), but want to be a little more explicit at this level, we can modify `rebuildWithQuery` to take advanced options to preserve the param states. Same goes for `resetQueryResult` in https://github.dev/finos/legend-studio/pull/1984

akphi (changes requested):

packages/legend-art/src/testMocks/MonacoEditorMockUtils.tsx
gayathrir11: I was trying to monaco in the test so i needed these functions

Issue #2006, opened 2023-03-14
Feature request: Improvements to the query import modal


gayathrir11:
### Implementation plan

| Query load flows   |  Rename | Preview | Delete | Last modified date | Filter by Class | Filter by dataspace | Filter by mine |
| -------------------- | ---------- | -------- | ------- | --------------------- | --------------- | --------------------  | --------------- |
|   Service execution editor                            |        ❎        |        ✔️       |      ❎      |           ✔️                      |             ❎            |      ❎                          |                  ✔️     |
| Load existing query (setup)  | ✔️ | ✔️  |✔️  | ✔️  |  ❎ |  ❎  | ✔️  |
| Productionize query (setup)  |  ❎| ✔️  | ❎ | ✔️  |  ❎ |  ❎  | ✔️  |
| Load existing query (query editor)  | ✔️ | ✔️  |✔️  | ✔️  |  ❎ |  ❎  | ✔️  |
| Load existing query (from dataspace)  | ✔️ | ✔️  |✔️  | ✔️  | ✔️  | ✔️   | ✔️  |

***Engine work***
- [x] Modify search endpoint to return last modified date for a query finos/legend-engine#1535
- [x] Support searching queries by query id finos/legend-engine#1609
     
***General design of import modal***

- Filter Options

    - Mine only
    
- Advanced options
    
     - Preview
     
***Studio Tasks***

- [x] Send `classPath` as tagged value when we save/update queries so that we can filter via class #2018 
- [x] Add a generic load query importer
- [x] Customize load query importer for all workflows via extensions
- [x] Limit to search to list of top 10 queries
- [x] Make the minimum search text to be 3 letters
- [x] Leverage `localStorage` to store list of 10 query ids most used by the user (update the list stored whenever user tries to access a new query. Use concept of caching here)
- [ ] Add metrics around the operations getting added (THINK)

akphi:
| Query load flows                     | Rename | Preview | Delete | Last modified date | Filter by Class | Filter by dataspace | Filter by mine |  | SearchModified                      | View | Free |                                    |
| ------------------------------------ | ------ | ------- | ------ | ------------------ | --------------- | ------------------- | -------------- |  | ----------------------------------- | ---- | ---- | ---------------------------------- |
| Service execution editor             | ❎      | ✔️      | ❎      | ✔️                 | ❎               | ❎                   | ✔️             |  | x                                   | x    | x    | <- should be scoped to the project |
[...]

PR #2018, opened 2023-03-16
Modify Query to have classPath as tagged value when we save it via different query workflows

gayathrir11:
## Summary

#2006

- [ ] Test(s) added

### DIFF
--- a/.changeset/little-flies-unite.md
+++ b/.changeset/little-flies-unite.md
@@ -1,0 +1,5 @@
+---
+'@finos/legend-extension-dsl-data-space': patch
+---
+
+Modify `Query` to have `classPath` as tagged value when we save it via dataspace workflow
--- a/packages/legend-extension-dsl-data-space/src/DSL_DataSpace_Const.ts
+++ b/packages/legend-extension-dsl-data-space/src/DSL_DataSpace_Const.ts
@@ -15,7 +15,7 @@
  */
 
 // NOTE: these are non-compilable profile and tag that we come up with for query
-// so that it records the dataSpace and class it is created from
+// to store extra metadata
 export const QUERY_PROFILE_PATH = 'meta::pure::profiles::query';
 export const QUERY_PROFILE_TAG_DATA_SPACE = 'dataSpace';
 export const QUERY_PROFILE_TAG_CLASS = 'class';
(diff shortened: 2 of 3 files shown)


.changeset/little-flies-unite.md
```diff
@@ -0,0 +1,6 @@
+---
+'@finos/legend-extension-dsl-data-space': patch
```
xannem, line 2: Not sure if this should be separated into a different changeset or not 😓

xannem (dismissed):

packages/legend-graph/src/graphManager/action/query/QueryUtils.ts
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+import { QueryTaggedValue } from './Query.js';
+
+// NOTE: these are non-compilable profile and tag that we come up with for query
+// so that it records the class it is created from
+export const QUERY_PROFILE_PATH = 'meta::pure::profiles::query';
+export const QUERY_PROFILE_TAG_CLASS = 'class';
+
+export const createQueryClassTaggedValue = (
```
akphi, line 24: I think this is specific to query created from data-space, can we move this to where we handle the same synthetic tagged values logic for data-space query?

packages/legend-application-query/src/stores/QueryEditorStore.ts
```diff
@@ -634,6 +635,11 @@ export class MappingQueryCreatorStore extends QueryEditorStore {
         query.groupId = this.groupId;
         query.artifactId = this.artifactId;
         query.versionId = this.versionId;
+        query.taggedValues = [
+          createQueryClassTaggedValue(
```
akphi, line 639: I actually think we don't want to do this analytics for query not created from a dataspace :( let's not do that.

.changeset/little-flies-unite.md
akphi: I think, the change in this PR should only happen in `dsl-data-space`

akphi (changes requested):

packages/legend-extension-dsl-data-space/src/DSL_DataSpace_Const.ts
```diff
@@ -15,6 +15,7 @@
  */
 
 // NOTE: these are non-compilable profile and tag that we come up with for query
-// so that it records the dataSpace it is created from
+// so that it records the dataSpace and class it is created from
```
akphi, line 18:
```suggestion
// to store extra metadata
```

akphi (changes requested):

akphi (approved):

PR #2037, opened 2023-03-22
Fix: allow query ui results to render data for columns with dot notation

xannem:
Originally, query results were not returning data in the special case where a column name had contained a '.' 

```

    executionResult.result = {
      columns: ['This Column Has .Dot', 'Normal Column'],
      rows: [
        {
          values: ['valueOne', 'valueTwo'],
        },
      ],
    };
```
would have shown up as 
image.
This is due to how ag-grid-table options render it, more detail found: https://www.ag-grid.com/javascript-data-grid/column-definitions/#suppress-field-dot-notation.

Fixed so now data will render successfully and not show up as empty cells even though values exist.

- [ ] Test(s) added

### DIFF
--- a/.changeset/debonair-ducks-dwaddle.md
+++ b/.changeset/debonair-ducks-dwaddle.md
@@ -1,0 +1,5 @@
+---
+'@finos/legend-query-builder': patch
+---
+
+Fix for allowing column names that include '.' to render data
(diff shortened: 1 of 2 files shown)


YannanGao-gs (approved):

akphi (approved):

MauricioUyaguari:
@xannem  Test?

PR #2053, opened 2023-03-28
Complex pure prop mapping

hardikmaheshwari:
## Summary
Remove assertion for non-nullable target set for complex property. We are relaxing this in backend
Ref: https://github.com/finos/legend-engine/pull/1577

- [ ] Test(s) added


MauricioUyaguari:
@hardikmaheshwari have we checked that the analyze mapping returns the properties of the complex properties as true so this works with query builder out of the box. 
Additionally, does the mapping editor, behave if this mapping is loaded. I believe we rely on the target being there for rendering but maybe not.

hardikmaheshwari:
> @hardikmaheshwari have we checked that the analyze mapping returns the properties of the complex properties as true so this works with query builder out of the box. Additionally, does the mapping editor, behave if this mapping is loaded. I believe we rely on the target being there for rendering but maybe not.

@MauricioUyaguari Thanks for looking, yes I loaded the mapping editor and tested it out. It was failing earlier because of asserts which I am removing. 
As for query builder you are right analyze mapping won't return the nested properties but it might be a bit tricky to handle cyclic models from the backend so leaving that for now

MauricioUyaguari (approved):

PR #2069, opened 2023-04-03
fix diagram simplify path when property path overlays on classes

MauricioUyaguari:
## Summary

- [ ] bug fix

- [ ] Test(s) added

### DIFF
--- a/packages/legend-extension-dsl-diagram/src/components/studio/__tests__/ClassDiagramPreview.test.tsx
+++ b/packages/legend-extension-dsl-diagram/src/components/studio/__tests__/ClassDiagramPreview.test.tsx
@@ -68,3 +68,105 @@ test(integrationTest('Class diagram preview shows up properly'), async () => {
     renderResult.getByTestId(DSL_DIAGRAM_TEST_ID.CLASS_DIAGRAM_PREVIEW),
   );
 });
+
+const TEST_DATA__dummyModelWithDiagram = [
+  {
+    path: 'model::Person',
+    content: {
+      _type: 'class',
+      name: 'Person',
+      superTypes: [],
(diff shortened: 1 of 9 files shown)


packages/legend-extension-dsl-diagram/src/components/studio/__tests__/ClassDiagramPreview.test.tsx
```diff
@@ -68,3 +68,105 @@ test(integrationTest('Class diagram preview shows up properly'), async () => {
...
+        },
+      ],
+      generalizationViews: [],
+      package: 'model',
+    },
+    classifierPath: 'meta::pure::metamodel::diagram::Diagram',
+  },
+];
+
+test(integrationTest('Diagram editor shows up properly'), async () => {
```
akphi, line 160: Should we move this test to graph builder for DSL diagram instead? and rename it to the more specific case it's trying to test

packages/legend-extension-dsl-diagram/src/graphManager/protocol/pure/v1/transformation/pureGraph/V1_DSL_Diagram_GraphBuilderHelper.ts
```diff
@@ -92,6 +95,19 @@ export const V1_buildClassView = (
   return view;
 };
 
+const _relationshipView_simplifyPathWithErrorHandling = (
```
akphi, line 98: I think the thing we should say here is this is an optimization we do when building the diagram; not something that will render the diagram non-compilable, so we shouldn't break.

akphi (approved):

PR #2084, opened 2023-04-06
improve performance of running all mapping legacy tests

MauricioUyaguari:
## Summary 
- [x] running all mapping tests is causing memory issues on large projects. A major reason is we currently transform the graph for each test. We are changing this to transform the graph once and use that same pmcd to run each test. Additionally this https://github.com/finos/legend-studio/pull/2087 will help as well when using large dependency models. This is a temproary fix as the long term solution is to move over to the new testable framework where Engine does all the work and just returns a list of test results. 

- [ ] Test(s) added

### DIFF
--- a/packages/legend-graph/src/graphManager/AbstractPureGraphManager.ts
+++ b/packages/legend-graph/src/graphManager/AbstractPureGraphManager.ts
@@ -85,6 +85,7 @@ import type {
 import type { PackageableRuntime } from '../graph/metamodel/pure/packageableElements/runtime/PackageableRuntime.js';
 import type { GraphDataOrigin } from '../graph/GraphDataOrigin.js';
 import type { GraphData } from './GraphData.js';
+import type { DEPRECATED__MappingTest } from '../graph/metamodel/pure/packageableElements/mapping/DEPRECATED__MappingTest.js';
 
 export interface TEMPORARY__EngineSetupConfig {
   env: string;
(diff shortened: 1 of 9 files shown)


akphi (approved):

packages/legend-graph/src/graphManager/AbstractPureGraphManager.ts
```diff
@@ -402,6 +403,19 @@ export abstract class AbstractPureGraphManager {
     report?: GraphManagerOperationReport,
   ): Promise<ExecutionResult>;
 
+  abstract DEPREACTED_runLegacyMappingTests(
```
akphi, line 406: Typo and lets do double underscores

akphi (changes requested):

packages/legend-graph/src/graphManager/protocol/pure/v1/V1_PureGraphManager.ts
```diff
@@ -2365,6 +2387,80 @@ export class V1_PureGraphManager extends AbstractPureGraphManager {
...
+          mapping,
+          pureModelContext,
+          options,
+          report,
+        ),
+      ),
+    );
+  }
+
+  private async DEPREACTED_runLegacyMappingTest(
```
akphi, line 2421: TODO: use double underscore

akphi (approved):

PR #2086, opened 2023-04-06
Validation on constant and parameter names in query

xannem:
## Summary
Checks for a valid string when users enter parameter or constant names in query: 
- [x] Disables creation of constant or parameter name if invalid 
image

- [x] Does not save renamed/updated state of constant or parameter name if invalid

Kapture 2023-04-10 at 15 19 20

- [x] Warns user if they try to navigate away from constant/parameter name window after renaming it to something invalid 
Kapture 2023-04-11 at 14 01 24

- [x] Test(s) added

### DIFF
--- a/packages/legend-art/src/panel/Panel.tsx
+++ b/packages/legend-art/src/panel/Panel.tsx
@@ -17,7 +17,7 @@
 import { capitalize, prettyCONSTName, toTitleCase } from '@finos/legend-shared';
 import { clsx } from 'clsx';
 import { observer } from 'mobx-react-lite';
-import { forwardRef } from 'react';
+import { forwardRef, useState } from 'react';
 import { CheckSquareIcon, SquareIcon } from '../icon/Icon.js';
 import { generateSimpleDIVComponent } from '../utils/ComponentCreatorUtils.js';
 
@@ -175,6 +175,107 @@ export const PanelFormSection = generateSimpleDIVComponent(
   'panel__content__form__section',
 );
 
+export const PanelFormValidatedTextField = forwardRef<
+  HTMLInputElement,
+  {
+    name?: string;
+    value: string | undefined;
+    update: (value: string | undefined) => void;
+    validateInput: (input: string) => string | undefined;
+    onValidationFailure?:
+      | ((validationIssue: string) => void | string | undefined)
+      | undefined;
+    prompt?: string | React.ReactNode;
+    placeholder?: string;
+    isReadOnly?: boolean;
+    className?: string | undefined;
+    errorMessageClassname?: string | undefined;
+    inputType?: string | undefined;
+    darkMode?: boolean;
+    fullWidth?: boolean;
+  }
+>(function PanelFormValidatedTextField(props, ref) {
+  const {
+    name,
+    value,
+    update,
+    prompt,
+    errorMessageClassname,
+    placeholder,
+    validateInput,
+    onValidationFailure,
+    isReadOnly,
+    className,
+    darkMode,
+    fullWidth,
+    inputType,
+  } = props;
+
+  const displayValue = value ?? '';
+
+  const [inputValue, setInputValue] = useState(displayValue);
+
+  const changeValue: React.ChangeEventHandler<HTMLInputElement> = (event) => {
+    const stringValue = event.target.value;
+    const updatedValue = stringValue ? stringValue : undefined;
+
+    setInputValue(updatedValue ?? '');
+
+    const validationErrorMessage = validateInput(updatedValue ?? '');
+    if (validationErrorMessage === undefined) {
+      update(updatedValue);
+    } else {
+      if (onValidationFailure !== undefined) {
+        onValidationFailure(validationErrorMessage);
+      }
+    }
+  };
+
+  return (
+    <PanelFormSection>
+      {name && (
+        <div className="panel__content__form__section__header__label">
+          {capitalize(name)}
+        </div>
+      )}
+      {prompt && (
+        <div className="panel__content__form__section__header__prompt">
+          {prompt}
+        </div>
+      )}
+      <div className="input-group">
+        <input
+          className={clsx(
+            'input input-group__input panel__content__form__section__input',
+            className,
+            { 'input--dark': darkMode ? darkMode : true },
+            { 'input--small': !fullWidth },
+          )}
+          ref={ref}
+          type={inputType ? inputType : 'text'}
+          spellCheck={false}
+          disabled={isReadOnly}
+          placeholder={placeholder}
+          value={inputValue}
+          onChange={changeValue}
+        />
+        {validateInput(inputValue) && (
+          <div
+            className={clsx(
+              'panel__content__form__section__input-group__error-message input-group__error-message',
+              { 'input--small': !fullWidth },
+              errorMessageClassname,
+            )}
+          >
+            {validateInput(inputValue)}
+          </div>
+        )}
+      </div>
+      {validateInput(inputValue) !== undefined && <PanelDivider />}
+    </PanelFormSection>
+  );
+});
+
 export const PanelFormTextField = forwardRef<
   HTMLInputElement,
   {
@@ -204,6 +305,7 @@ export const PanelFormTextField = forwardRef<
     fullWidth,
     inputType,
   } = props;
+
   const displayValue = value ?? '';
   const changeValue: React.ChangeEventHandler<HTMLInputElement> = (event) => {
     const stringValue = event.target.value;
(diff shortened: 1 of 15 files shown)


packages/legend-query-builder/src/components/QueryBuilderConstantExpressionPanel.tsx
```diff
@@ -86,6 +89,8 @@ const QueryBuilderConstantExpressionEditor = observer(
         );
     const validationMessage = !variableName
       ? `Constant Name can't be empty`
+      : validateValidString(variableName) === false
```
akphi, line 92: the one thing we need to add here is we must block the saving, right now we only show the warning, but users are still able to create/edit the variable

xannem: makes sense!

packages/legend-query-builder/src/components/QueryBuilderConstantExpressionPanel.tsx
```diff
@@ -105,6 +114,12 @@ const QueryBuilderConstantExpressionEditor = observer(
       constantState.setValueSpec(valSpec);
     };
 
+    const saveConstantName = (): void => {
+      if (!validationMessage) {
+        variableExpression_setName(varExpression, constantNameValue);
+      }
```
gayathrir11, line 120: Do we want to show a warning to the user after they exit the constant panel that we haven't saved the constant name they have updated?

packages/legend-query-builder/src/components/__tests__/QueryBuilderParametersPanel.test.tsx
```diff
@@ -45,7 +45,7 @@ import {
 
 test(
   integrationTest(
-    'Query builder shows validation error for constant name if existing duplicate parameter name',
+    'Query builder shows validation error for parameter name if existing duplicate constant name',
```
gayathrir11, line 48: why do we change this? Don't we have validations on constant name if there is an existing paramter name anymore?

packages/legend-query-builder/src/components/QueryBuilderConstantExpressionPanel.tsx
xannem: That's true, might be good safeguard. I added now an update where if user navigates away and its invalid, then we will have a caution action alert that lets them know 👍

packages/legend-shared/src/CommonUtils.ts
```diff
@@ -224,6 +224,14 @@ export const shallowStringify = (object: unknown): string =>
       : val,
   );
 
+export const validateValidString = (input: string): boolean => {
```
akphi, line 227: we can do one-liner here

akphi: let's rename this to `isValidIdentifier` and it could be in `formatterUtils` or probably even better place is `@finos/legend-graph/src/MetaModelUtil`, potentially, this can take a flag to accept `$` or not

packages/legend-shared/src/CommonUtils.ts
```diff
@@ -224,6 +224,14 @@ export const shallowStringify = (object: unknown): string =>
...
+  if (input.match(/^[a-z][\w\d]*$/)) {
```
akphi, line 228: I thought `\w` includes `\d`? - Also I think we diverge here, Pure/Engine allows `$` after the first char <---- let's make it optional check for us

packages/legend-query-builder/src/components/QueryBuilderParametersPanel.tsx
```diff
@@ -123,9 +114,52 @@ const VariableExpressionEditor = observer(
     const changeMultiplicity = (val: MultiplicityOption): void => {
       lambdaParameterState.changeMultiplicity(varState, val.value);
     };
+    const [parameterNameValue, setParameterNameValue] = useState(varState.name);
```
akphi, line 117: hm, I think finding out in the alert is a bit late, why don't we do it the service pattern editor way? <----- potentially we can componentize that as well

packages/legend-art/src/components/panel/Panel.tsx
```diff
@@ -181,6 +181,7 @@ export const PanelFormTextField = forwardRef<
     name: string;
     value: string | undefined;
     update: (value: string | undefined) => void;
+    updateBlur?: () => void;
```
akphi, line 184: don't think we need this for now, see me other comments in the component

akphi (changes requested):

packages/legend-art/src/panel/Panel.tsx
```diff
@@ -175,6 +175,107 @@ export const PanelFormSection = generateSimpleDIVComponent(
...
+    validateInput,
+    onValidationFailure,
+    isReadOnly,
+    className,
+    darkMode,
+    fullWidth,
+    inputType,
+  } = props;
+
+  const displayValue = value ?? '';
```
akphi, line 214: don't think we need this

packages/legend-art/src/panel/Panel.tsx
```diff
@@ -175,6 +175,107 @@ export const PanelFormSection = generateSimpleDIVComponent(
...
+
+  const [inputValue, setInputValue] = useState(displayValue);
+
+  const changeValue: React.ChangeEventHandler<HTMLInputElement> = (event) => {
+    const stringValue = event.target.value;
+    const updatedValue = stringValue ? stringValue : undefined;
```
akphi, line 220: don't think we need this either, `event.target.value` is always non-nullable

packages/legend-art/src/panel/Panel.tsx
```diff
@@ -175,6 +175,107 @@ export const PanelFormSection = generateSimpleDIVComponent(
...
+
+    setInputValue(updatedValue ?? '');
+
+    const validationErrorMessage = validateInput(updatedValue ?? '');
+    if (validationErrorMessage === undefined) {
```
akphi, line 225: validation can be done as part of the normal logic outside of this function, as it automatically gets re-triggered as the `setState` is called

packages/legend-art/src/panel/Panel.tsx
```diff
@@ -175,6 +175,107 @@ export const PanelFormSection = generateSimpleDIVComponent(
...
+    darkMode,
+    fullWidth,
+    inputType,
+  } = props;
+
+  const displayValue = value ?? '';
+
+  const [inputValue, setInputValue] = useState(displayValue);
+
+  const changeValue: React.ChangeEventHandler<HTMLInputElement> = (event) => {
```
akphi, line 218: Just a comment really, I'm not sure yet what's the best way to handle it, but with this, when the value is changed elsewhere, we won't be able to update the form value (so is the problem of using `useState` in general)

packages/legend-query-builder/src/components/QueryBuilderConstantExpressionPanel.tsx
```diff
@@ -84,12 +84,29 @@ const QueryBuilderConstantExpressionEditor = observer(
...
-      ? 'Constant Name Already Exists'
-      : undefined;
+
+    const [hasFailedValidation, setFailedValidation] = useState<boolean>(false);
+
+    const getValidationMessage = (
+      constantInput: string,
+    ): string | undefined => {
+      const possibleMessage = !constantInput
+        ? `Constant Name can't be empty`
```
akphi, line 94:
```suggestion
        ? `Constant name can't be empty`
```

packages/legend-query-builder/src/components/QueryBuilderParametersPanel.tsx
```diff
@@ -214,7 +229,7 @@ const VariableExpressionEditor = observer(
             {isCreating && (
               <ModalFooterButton
                 text="Create"
-                inProgress={Boolean(validationMessage)}
+                inProgress={hasFailedValidation}
```
akphi, line 232: why this, instead of `disabled`?

packages/legend-query-builder/src/components/QueryBuilderConstantExpressionPanel.tsx
```diff
@@ -181,7 +198,7 @@ const QueryBuilderConstantExpressionEditor = observer(
             {isCreating && (
               <ModalFooterButton
                 text="Create"
-                inProgress={Boolean(validationMessage)}
+                inProgress={hasFailedValidation}
```
akphi, line 201: why this, instead of `disabled`?

akphi (approved):

PR #2202, opened 2023-05-04
rename query fix

xannem:
## Summary

- [x] Removing `rename query` from additional options and using behavior of double-clicking title instead
- [x] Fixing regression bug of not having query title displayed 
- [x] Changing searching validation for duplicate query name to only include queries that are created by user 

Kapture 2023-05-04 at 15 43 23

- [ ] Test(s) added

### DIFF
--- a/packages/legend-application-query/src/components/QueryEditor.tsx
+++ b/packages/legend-application-query/src/components/QueryEditor.tsx
@@ -43,7 +43,7 @@ import {
   clsx,
 } from '@finos/legend-art';
 import { observer } from 'mobx-react-lite';
-import { Fragment, useEffect, useMemo, useRef } from 'react';
+import { Fragment, useEffect, useMemo, useRef, useState } from 'react';
 import {
   type MappingQueryCreatorPathParams,
   type ExistingQueryEditorPathParams,
@@ -199,6 +199,15 @@ const QueryEditorHeaderContent = observer(
     const renameState = editorStore.renameState;
     const applicationStore = useLegendQueryApplicationStore();
 
+    const renameRef = useRef<HTMLInputElement>(null);
+    const [showQueryNameEditInput, setShowQueryNameEditInput] =
+      useState<boolean>(false);
+
+    useEffect(() => {
+      renameRef.current?.focus();
+      renameRef.current?.select();
+    }, [showQueryNameEditInput]);
+
     // actions
     const openQueryLoader = (): void => {
       editorStore.queryLoaderState.setQueryLoaderDialogOpen(true);
@@ -236,8 +245,6 @@ const QueryEditorHeaderContent = observer(
       );
     };
 
-    const renameRef = useRef<HTMLInputElement>(null);
-
     const updateQuery = applicationStore.guardUnhandledError(
       async (): Promise<void> => {
         try {
@@ -252,7 +259,7 @@ const QueryEditorHeaderContent = observer(
               }),
             ),
           );
-          renameRef.current?.select();
+          setShowQueryNameEditInput(true);
         } catch {
           // do nothing
         }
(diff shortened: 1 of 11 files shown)


packages/legend-query-builder/style/shared/_query-loader.scss
```diff
@@ -243,6 +244,7 @@
 
     &__author__icon {
       svg {
+        display: flex;
```
akphi, line 247: this looks weird, don't think we ever need to set flex to the `svg`

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -540,15 +541,6 @@ const QueryEditorHeaderContent = observer(
             disabled={editorStore.isViewProjectActionDisabled}
             content={
               <MenuContent>
-                {/* TODO: remove this and move it to inline editing directly on the title */}
```
akphi, line 543: It's okay to leave this here I think, let's remove the comment

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -416,6 +416,7 @@ const QueryEditorHeaderContent = observer(
           </div>
         ) : (
           <div
+            onDoubleClick={updateQuery}
```
akphi, line 419: nice! But in terms of UX, we need visual hints that this is possible, using tooltip or some CSS border, or altering the `cursor: pointer`

gayathrir11 (approved):

packages/legend-application-query/style/_query-editor.scss
```diff
@@ -61,6 +61,10 @@
         padding: 0.5rem;
         margin-right: 10%;
       }
+
+      &__title:hover {
```
akphi, line 65: usually, when one sees something like this, it means you can set `cursor: pointer` on the base CSS class

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -384,15 +391,24 @@ const QueryEditorHeaderContent = observer(
...
-                  onKeyDown={onEnter}
+                  onKeyDown={(event) => {
+                    if (event.code === 'Enter') {
+                      event.stopPropagation();
+                      setShowQueryNameEditInput(false);
+                      onEnter(event);
+                    } else if (event.code === 'Escape') {
+                      setShowQueryNameEditInput(false);
+                      event.stopPropagation();
+                      renameState.renameQueryState.reset();
```
akphi, line 407: Hitting `Esc` should do select all, isntead of erasing everything - `renameRef.current?.select();`

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -408,16 +424,17 @@ const QueryEditorHeaderContent = observer(
...
                 <CheckIcon />
               </button>
             </PanelListItem>
           </div>
         ) : (
           <div
+            onDoubleClick={updateQuery}
             className="query-editor__header__content__main query-editor__header__content__title"
-            title="Query title"
+            title="Double-click to rename Query"
```
akphi, line 437:
```suggestion
            title="Double-click to rename query"
```

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -408,16 +424,17 @@ const QueryEditorHeaderContent = observer(
                   'btn--icon__caution': isExistingQueryName,
                 })}
                 onClick={renameQuery}
-                title="Rename Query"
+                title="Rename query"
               >
                 <CheckIcon />
```
akphi, line 429: You know, I think we can do a little better, we don't need the check button, on `Enter`, we will save, I think this is a fairly clear interaction

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -384,15 +391,24 @@ const QueryEditorHeaderContent = observer(
...
+                      setShowQueryNameEditInput(false);
+                      event.stopPropagation();
+                      renameState.renameQueryState.reset();
+                      renameState.editorStore.setRenameState(undefined);
+                    }
+                  }}
                   value={renameState.queryName}
-                  placeholder="Search"
                 />
                 {isExistingQueryName && (
```
akphi, line 413: Did we leverage the API you added in https://github.com/finos/legend-engine/pull/1669 ?

packages/legend-query-builder/src/stores/QueryLoaderState.ts
```diff
@@ -246,7 +246,7 @@ export class QueryLoaderState {
         name,
       )) as Query;
       this.onQueryRenamed?.(query);
-      this.applicationStore.notificationService.notify(
+      this.applicationStore.notificationService.notifySuccess(
```
akphi, line 249: TODO: just making a comment here, probably would be great if we can bring the same duplicated name warning here

akphi (changes requested):

packages/legend-query-builder/src/stores/QueryLoaderState.ts
xannem: Makes sense, added 👍

akphi (approved):

PR #2205, opened 2023-05-05
Make call to /auth/authorize blocking

ivan-kyosev-gs:
## Summary

This PR makes the call to /auth/authorize on legend-sdlc block.

Previously, when landing on the studio page for the first time, if the /auth/authorized endpoint returned 'false', the 'window.location.href' redirect to  /auth/authorize would run alongside other API calls to legend-sdlc. This was causing failures when legend-sdlc was set to use OIDC for identity delegation.

Now the call to /auth/authorize resolves without making any further requests to the server, before the redirect is complete.

Manual testing - I was using a debugger connected to legend-sdlc to see what endpoints are called in which order when loading the root page.


packages/legend-application-studio/src/stores/LegendStudioBaseStore.ts
```diff
@@ -182,12 +182,14 @@ export class LegendStudioBaseStore {
       this.isSDLCAuthorized =
         (yield this.sdlcServerClient.isAuthorized()) as boolean;
       if (!this.isSDLCAuthorized) {
-        this.applicationStore.navigationService.navigator.goToAddress(
-          SDLCServerClient.authorizeCallbackUrl(
-            this.applicationStore.config.sdlcServerUrl,
-            this.applicationStore.navigationService.navigator.getCurrentAddress(),
-          ),
-        );
+        yield new Promise(() => {
```
akphi, line 185: I don't think this wrapping of promise is doing anything, I wonder what are we trying to do here really?

Do we want to have a timeout/a wait? if so we can do something similar to a Thread.sleep - i.e. await setTimeout... in JS

Since we're going to a new page anyway and that's end of life of the current app, should we just simply call `return;` right here?

packages/legend-application-studio/src/stores/LegendStudioBaseStore.ts
```diff
@@ -188,6 +190,7 @@ export class LegendStudioBaseStore {
             this.applicationStore.navigationService.navigator.getCurrentAddress(),
           ),
         );
+        return false;
```
akphi, line 193: can we add extended documentation here on why we do it like this, and why do we need to do this.

could be here, or could be near line 117. Thanks!

ivan-kyosev-gs: I reworked this to make it cleaner

akphi (approved):

akphi (approved):

akphi (changes requested):
Looks like you need a changeset

akphi:
> Looks like you need a changeset - you can leave its message empty though

I meant a changeset with the changed package but you dont need to provide any message

akphi (changes requested):

.changeset/wild-geese-cry.md
```diff
@@ -0,0 +1,2 @@
+---
```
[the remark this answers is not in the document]
ivan-kyosev-gs: Running "yarn changeset" generated the file like this. I needed to manually fill it in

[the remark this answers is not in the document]
akphi: `yarn changeset` only works on your git staging area

.changeset/wild-geese-cry.md
```diff
@@ -0,0 +1,6 @@
...
+'@finos/legend-application-studio': patch
+'@finos/legend-server-sdlc': patch
+---
+
+Stop making sdlc server calls after a redirect to /api/auth/authorize has started
```
akphi, line 6:
```suggestion
```

akphi (approved):

PR #2213, opened 2023-05-07
support for default parameter values in saved queries

MauricioUyaguari:

- [ ] Test(s) added

https://github.com/finos/legend-studio/assets/29690824/0e272f7b-c0e8-48d7-9ce7-0ea81a0827ed

### DIFF
--- a/packages/legend-application-query/src/stores/QueryEditorStore.ts
+++ b/packages/legend-application-query/src/stores/QueryEditorStore.ts
@@ -51,10 +51,12 @@ import {
   Mapping,
   type Runtime,
   type Service,
+  type ValueSpecification,
   createGraphBuilderReport,
   LegendSDLC,
   QuerySearchSpecification,
   toLightQuery,
+  QueryParameterValue,
 } from '@finos/legend-graph';
 import {
   EXTERNAL_APPLICATION_NAVIGATION__generateStudioProjectViewUrl,
@@ -194,6 +196,7 @@ export class QueryCreatorState {
       const query = (yield this.editorStore.buildQueryForPersistence(
         new Query(),
         rawLambda,
+        queryBuilderState.getCurrentParameterValues(),
         config,
       )) as Query;
       query.name = this.queryName;
@@ -356,6 +359,7 @@ export abstract class QueryEditorStore {
   async buildQueryForPersistence(
     query: Query,
     rawLambda: RawLambda,
+    parameters: Map<string, ValueSpecification> | undefined,
     config: QueryPersistConfiguration | undefined,
   ): Promise<Query> {
     try {
@@ -379,6 +383,30 @@ export abstract class QueryEditorStore {
       query.content =
         await this.graphManagerState.graphManager.lambdaToPureCode(rawLambda);
       config?.decorator?.(query);
+      // any error in default parameters we will ignore and only log
+      query.defaultParameterValues = undefined;
+      try {
+        if (parameters?.size) {
+          const defaultParameterValues =
+            await this.graphManagerState.graphManager.valueSpecificationsToPureCode(
+              parameters,
+            );
+          query.defaultParameterValues = Array.from(
+            defaultParameterValues.entries(),
+          ).map(([key, val]) => {
+            const queryParam = new QueryParameterValue();
+            queryParam.name = key;
+            queryParam.content = val;
+            return queryParam;
+          });
+        }
+      } catch (error) {
+        assertErrorThrown(error);
+        this.applicationStore.logService.error(
+          LogEvent.create(LEGEND_QUERY_APP_EVENT.GENERIC_FAILURE),
+          error,
+        );
+      }
       return query;
     } catch (error) {
       assertErrorThrown(error);
@@ -751,7 +779,6 @@ export class ServiceQueryCreatorStore extends QueryEditorStore {
 export class ExistingQueryUpdateState {
   readonly editorStore: ExistingQueryEditorStore;
   readonly updateQueryState = ActionState.create();
-
   queryRenamer = false;
   saveModal = false;
   updateDiffState: QueryBuilderDiffViewState | undefined;
@@ -804,6 +831,7 @@ export class ExistingQueryUpdateState {
       const query = (yield this.editorStore.buildQueryForPersistence(
         this.editorStore.query ?? new Query(),
         rawLambda,
+        queryBuilderState.getCurrentParameterValues(),
         config,
       )) as Query;
       query.name = queryName ?? query.name;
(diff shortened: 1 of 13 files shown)


packages/legend-graph/src/graph-manager/AbstractPureGraphManager.ts
```diff
@@ -317,6 +317,16 @@ export abstract class AbstractPureGraphManager {
     pretty?: boolean,
   ): Promise<Map<string, string>>;
 
+  abstract valueSpecsToPureCode(
```
akphi, line 320:
```suggestion
  abstract valueSpecificationsToPureCode(
```

packages/legend-graph/src/graph-manager/AbstractPureGraphManager.ts
```diff
@@ -317,6 +317,16 @@ export abstract class AbstractPureGraphManager {
...
+    lambdas: Map<string, ValueSpecification>,
+    pretty?: boolean,
+  ): Promise<Map<string, string>>;
+
+  abstract pureCodeToValueSpecs(
```
akphi, line 325:
```suggestion
  abstract pureCodeToValueSpecifications(
```

packages/legend-graph/src/graph-manager/action/query/Query.ts
```diff
@@ -29,6 +29,11 @@ export class QueryStereotype {
   stereotype!: string;
 }
 
+export class QueryParameterValue {
```
akphi, line 32: I think we should rename this to `QueryDefaultParameterValue` to be clearer
```suggestion
export class QueryDefaultParameterValue {
```

packages/legend-graph/src/graph-manager/protocol/pure/v1/engine/query/V1_Query.ts
```diff
@@ -23,6 +23,21 @@ import {
   V1_taggedValueModelSchema,
 } from '../../transformation/pureProtocol/serializationHelpers/V1_DomainSerializationHelper.js';
 
+export class V1_QueryParameterValue {
```
akphi, line 26:
```suggestion
export class V1_QueryDefaultParameterValue {
```

packages/legend-application-query/src/stores/QueryEditorStore.ts
```diff
@@ -379,6 +383,30 @@ export abstract class QueryEditorStore {
       query.content =
         await this.graphManagerState.graphManager.lambdaToPureCode(rawLambda);
       config?.decorator?.(query);
+      // any error in default parameters we will ignore and only log
+      query.defaultParameterValues = undefined;
```
akphi, line 387: The biggest concern I have here is, should we make this implied default behavior or this should be `a checkbox` when users save/saveAs their queries. What do you think?

packages/legend-query-builder/src/stores/QueryBuilderState.ts
```diff
@@ -371,6 +374,20 @@ export abstract class QueryBuilderState implements CommandRegistrar {
     this.setRuntimeValue(val);
   }
 
+  getParameterExecutionValues(): Map<string, ValueSpecification> | undefined {
```
akphi, line 377: First time I read this, I find it confusing since what does it really contrast to? let's simplify the name to `getCurrentParameterValues`

packages/legend-query-builder/src/stores/QueryBuilderState.ts
```diff
@@ -444,12 +461,42 @@ export abstract class QueryBuilderState implements CommandRegistrar {
...
+    defaultParameterValues?: Map<string, ValueSpecification>,
+  ): void {
+    this.rebuildWithQuery(query, {
+      defaultParameterValues,
+    });
     this.resetQueryResult();
     this.changeDetectionState.initialize(query);
   }
 
+  buildInitialParameterValues(
```
akphi, line 475: please inline this function, it's not being used any where else, and the file is already super-cluttered.

packages/legend-query-builder/src/stores/QueryBuilderState.ts
```diff
@@ -444,12 +461,42 @@ export abstract class QueryBuilderState implements CommandRegistrar {
...
+    const res = new Map<string, QueryBuilderParameterValue>();
+    if (defaultParameterValues?.size) {
+      Array.from(defaultParameterValues.entries()).forEach(([k, v]) => {
+        res.set(k, {
+          variable: k,
+          value: v,
+        });
+      });
+      return res;
+    } else if (preserveParameterValues) {
```
akphi, line 488: I would think `preserveParameterValues` take higher precedence than `defaultParameterValues`, here are we saying, for example I'm running my query A
- A has default param: x = 1
- We run A set x = 2
- We go to text-mode, when we come back to form mode, x = 1

packages/legend-query-builder/src/stores/QueryBuilderState.ts
```diff
@@ -444,12 +461,42 @@ export abstract class QueryBuilderState implements CommandRegistrar {
...
+      this.parametersState.parameterStates.forEach((ps) => {
+        res.set(ps.parameter.name, {
+          variable: ps.parameter,
+          value: ps.value,
+        });
+      });
+      return res;
+    }
+    return undefined;
```
akphi, line 497: Also, I would expect to logic here to be merging existing one with default ones where the former override, for example I'm running my query A

- A has default param: x = 1, y = 2
- We run A set x = 2, so we have x = 2, y = 2

akphi (changes requested):

packages/legend-application-query/src/stores/QueryEditorStore.ts
MauricioUyaguari: i think for now implied behaviour. we can look to maybe add it to the save modals but for me it should be like a hidden feature for now.

packages/legend-query-builder/src/stores/QueryBuilderState.ts
MauricioUyaguari: Fair point. 
My assumption was that either or would be set but not both. Will change

packages/legend-query-builder/src/stores/QueryBuilderState.ts
MauricioUyaguari: no i dont think we should merge. 
The expectation is really one or the other. The initial load of the query will have either defaultParameters or nothing. 
Each rebuild after has a defined state build and will use that.

packages/legend-application-query/src/stores/QueryEditorStore.ts
akphi: TODO: as dicussed, I think we should have something like a checkbox in the save/saveas modal to let people decide if they want to persist the parameters' values as default values or not

akphi (approved):

PR #2227, opened 2023-05-11
Utilize the cancelUserExecution API to cancel query execution both on the server's and client's side

YannanGao-gs:
## Summary

Utilize the cancelUserExecution API to cancel query execution both on the server's and client's side

- [ ] Test(s) added

### DIFF
--- a/packages/legend-application-studio/src/components/editor/editor-group/mapping-editor/DEPRECATED__MappingTestEditor.tsx
+++ b/packages/legend-application-studio/src/components/editor/editor-group/mapping-editor/DEPRECATED__MappingTestEditor.tsx
@@ -734,10 +734,11 @@ export const DEPRECATED__MappingTestEditor = observer(
     const runTest = applicationStore.guardUnhandledError(() =>
       flowResult(testState.runTest()),
     );
-    const cancelTest = (): void => {
-      testState.setIsRunningTest(false);
-      testState.setTestRunPromise(undefined);
-    };
+
+    const cancelTest = applicationStore.guardUnhandledError(() =>
+      flowResult(testState.cancelTest()),
+    );
+
     const executionPlanState = testState.executionPlanState;
     const generatePlan = applicationStore.guardUnhandledError(() =>
       flowResult(testState.generatePlan(false)),
--- a/packages/legend-application-studio/src/stores/editor/editor-state/element-editor-state/service/ServiceExecutionState.ts
+++ b/packages/legend-application-studio/src/stores/editor/editor-state/element-editor-state/service/ServiceExecutionState.ts
@@ -466,6 +466,9 @@ export abstract class ServicePureExecutionState extends ServiceExecutionState {
     execution: PureExecution,
   ) {
     super(editorStore, serviceEditorState, execution);
+    makeObservable(this, {
+      cancelQuery: flow,
+    });
 
     this.execution = execution;
     this.queryState = new ServicePureExecutionQueryState(
@@ -618,15 +621,14 @@ export abstract class ServicePureExecutionState extends ServiceExecutionState {
       this.editorStore.applicationStore.telemetryService,
     );
 
+    let promise;
     try {
       this.isRunningQuery = true;
-
       const stopWatch = new StopWatch();
       const report = reportGraphAnalytics(
         this.editorStore.graphManagerState.graph,
       );
-
-      const promise = this.editorStore.graphManagerState.graphManager.runQuery(
+      promise = this.editorStore.graphManagerState.graphManager.runQuery(
         this.getExecutionQuery(),
         this.selectedExecutionContextState?.executionContext.mapping.value,
         this.selectedExecutionContextState?.executionContext.runtime,
@@ -640,7 +642,6 @@ export abstract class ServicePureExecutionState extends ServiceExecutionState {
         },
         report,
       );
-
       this.setQueryRunPromise(promise);
       const result = (yield promise) as ExecutionResult;
       if (this.queryRunPromise === promise) {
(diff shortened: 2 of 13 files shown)


packages/legend-application-studio/src/stores/editor/editor-state/element-editor-state/service/ServiceExecutionState.ts
```diff
@@ -661,17 +661,32 @@ export abstract class ServicePureExecutionState extends ServiceExecutionState {
...
-      this.editorStore.applicationStore.notificationService.notifyError(error);
+      // Don't notify users about success or failure
+      if (this.queryRunPromise === promise) {
+        assertErrorThrown(error);
+        this.editorStore.applicationStore.logService.error(
+          LogEvent.create(GRAPH_MANAGER_EVENT.EXECUTION_FAILURE),
+          error,
+        );
+        this.editorStore.applicationStore.notificationService.notifyError(
+          error,
```
gayathrir11, line 672: You made a comment here that we are not gonna user about the error then we should probably remove this?

MauricioUyaguari:
@YannanGao-gs  lets please clean up this PR a bit. i see a lot of small things that should like the changeset not being accurate, as well as some of the catch with // do nothing comment that we shouldnt need. This should be done before we send out to reviewers. 
Also explain why we are keeping some of the old code. Thanks!

packages/legend-application-studio/src/stores/editor/editor-state/element-editor-state/service/ServiceExecutionState.ts
YannanGao-gs: Heyy, we don't wanna notify users about any failure or success from the query that is being canceled. 
However, we should notify users about any failure from the current query that is being executed.
Therefore, ` if (this.queryRunPromise === promise) ` is used to identify if this failure is from the canceled query or executed query.

gayathrir11 (approved):

packages/legend-query-builder/src/components/QueryBuilderResultPanel.tsx
```diff
@@ -531,16 +531,20 @@ export const QueryBuilderResultPanel = observer(
...
         );
       }
+      resultState.pressedRunQuery.complete();
     };
     const cancelQuery = (): void => {
+      resultState.pressedRunQuery.complete();
       resultState.setIsRunningQuery(false);
       queryBuilderState.resultState.setQueryRunPromise(undefined);
+      flowResult(resultState.cancelQuery()).catch((error) => {
+        // Don't notify users about success or failure
```
MauricioUyaguari, line 546: log it

packages/legend-query-builder/src/stores/QueryBuilderResultState.ts
```diff
@@ -260,20 +261,31 @@ export class QueryBuilderResultState {
...
-        LogEvent.create(GRAPH_MANAGER_EVENT.EXECUTION_FAILURE),
-        error,
-      );
-      this.queryBuilderState.applicationStore.notificationService.notifyError(
-        error,
-      );
+      // When the query is canceled on the server's side by calling the API, there will be an execution failure error.
+      // We don't notify users about this success or failure.
+      // However, we should notify users about any failure from the current query that is being executed.
+      if (this.queryRunPromise === promise) {
```
MauricioUyaguari, line 267: // When user cancels the query by calling the `cancelQuery` api, it will throw an exeuction failure error. 
// For now, we don't want to notify users about this failure. Therefore we check to ensure the promise is still the same one.
// When cancelled the query, we set the `queryRunPromise` as undefined.

MauricioUyaguari (changes requested):
small comments

packages/legend-query-builder/src/components/QueryBuilderResultPanel.tsx
MauricioUyaguari: i dont see the logging of this?

packages/legend-application-studio/src/components/editor/editor-group/mapping-editor/DEPRECATED__MappingTestEditor.tsx
```diff
@@ -737,6 +737,9 @@ export const DEPRECATED__MappingTestEditor = observer(
     const cancelTest = (): void => {
       testState.setIsRunningTest(false);
```
MauricioUyaguari, line 738: we had said we would move the logic into cancel function and log the error.

MauricioUyaguari (changes requested):
comments

packages/legend-query-builder/src/components/__tests__/QueryBuilderFetchStructure.test.tsx
```diff
@@ -771,3 +771,30 @@ test(
...
+
+  await act(async () => {
+    queryBuilderState.initializeWithQuery(
+      create_RawLambda(undefined, TEST_DATA__simpleProjection.body),
+    );
+    const tdsState = guaranteeType(
+      queryBuilderState.fetchStructureState.implementation,
+      QueryBuilderTDSState,
+    );
+    tdsState.setShowWindowFuncPanel(true);
```
MauricioUyaguari, line 792: why do we need to show the window func. or get the tdtState?

MauricioUyaguari (changes requested):
comments

packages/legend-application-studio/src/components/editor/editor-group/mapping-editor/DEPRECATED__MappingTestEditor.tsx
```diff
@@ -735,8 +737,13 @@ export const DEPRECATED__MappingTestEditor = observer(
       flowResult(testState.runTest()),
     );
     const cancelTest = (): void => {
-      testState.setIsRunningTest(false);
-      testState.setTestRunPromise(undefined);
+      flowResult(testState.cancelTest()).catch((error) => {
```
MauricioUyaguari, line 740: lets inline the try, catch logic within the `cancelTest` method and the other cancel methods

MauricioUyaguari (approved):

PR #2291, opened 2023-06-02
fix a problem with fully-interactive service registration resolving the correct service

akphi:
This is an adjustment we need to make in corresponding to changes made in #2286


MauricioUyaguari:
https://tenor.com/view/dog-meme-gif-25326537

PR #2302, opened 2023-06-06
misc. cleanup

xannem:
## Summary

misc. cleanups 

- [x]  Rename `Global Test Runner` to `Test Runner`, put tabs bar below the top bar
- [x] Upon entering the renaming query interface text does not jerk but stays still, rounded border and margins consistent 
- [x] Query light mode edits 
- [x] add components for PanelHeaders 
- [x] Add PanelContentLists simple DIV component 

- [ ] Test(s) added


packages/legend-shared/src/format/FormatterUtils.ts
```diff
@@ -74,6 +76,9 @@ export const prettyCONSTName = (value: string | undefined): string => {
   if (!value) {
     return '';
   }
+  if (EXCEPTION_ACRONYMS.includes(value)) {
```
akphi, line 79: I don't think we should do this

xannem: Removed and also removed usage of `prettyCONSTName` in query builder dropdown 👍

akphi (approved):

Issue #2311, opened 2023-06-07
Feature request: Isolate DND actions when rearranging or dropping 


xannem:
### Similar requests

- [X] I have searched and found no existing similar requests

### What problems are you trying to solve?

Users are able to drag and drop properties to rearrange their order when building a query in the property fetch panel. In addition, they are also able to drag and drop properties from the property fetch panel to the filter panel. 

(Demonstrated below when moving `Cases` to the filter panel, the order is also shifted in the property fetch panel) 
![Kapture 2023-06-07 at 11 44 18](https://github.com/finos/legend-studio/assets/41248956/e1acb48b-ad2d-4f1a-affe-083004afe0b5)

This can cause inadvertent actions by the user when they are only trying to drag a property to a different panel and may lead to unintentional rearranging. 

### Describe the solution you would like to see implemented

We can change the rearranging of properties to only happen **after** the user has dropped their property and not rearrange while on hover. This way the destination of the property (which panel it ends up in) will limit the action that is permitted so it follows the user's intention. 

### Describe alternatives you have considered

We could guard the rearranging of properties to only happen when the user is holding down a key as well like `ctrl` so that the general drag and drop of properties can be unaffected. However, this would not be as intuitive and require more action on the user's part compared to the first solution.

_No response_

### Contribution

- [x] I would like to work on this feature

PR #2330, opened 2023-06-13
Limit DND rearrange option in query and misc. DND filter edits

xannem:
## Summary
Fixes https://github.com/finos/legend-studio/issues/2311 

- [x] Isolates the rearranging action to occur only after user drops column so that they do not accidentally rearrange column orders when trying to drag column to filter panel.
- [x] Adds more space for the droppable area of filter and post-filter panel 
        image

- [x] Reserves space for scrollbar in panels so that there is no jerking of the display when user drags and drops items 
- [x] Only shows droppable filter logical groups on hovering on the droppable area for less visual clutter
         Kapture 2023-06-15 at 10 19 45

(Isolating rearranging action example below, when dropping column Cases to filter panel the column order is not rearranged in the fetch structure panel, and is only rearranged when the column is dropped in the fetch structure panel)
Kapture 2023-06-15 at 16 33 54

 

- [ ] Test(s) added


gayathrir11 (approved):

akphi (changes requested):
@xannem TODO: please add the fix for re-arrange and I'll review one more round

akphi (approved):
TODO: for TDS projection column re-arrange: there are known limitations to the current drop location indicator that happens when drag-and-dropping the first/last item. I think we get the right spirit but we should go back and smoothen out those edge cases at some point

xannem:
> TODO: for TDS projection column re-arrange: there are known limitations to the current drop location indicator that happens when drag-and-dropping the first/last item. I think we get the right spirit but we should go back and smoothen out those edge cases at some point

Issue added: https://github.com/finos/legend-studio/issues/2380

PR #2343, opened 2023-06-16
Remove Binary Datatype from dropdown list

hardikmaheshwari:
## Summary
Binary Type is deprecated and should not be a part of the dropdown. Unfortunately we can't remove it entirely for backward compatibiity

- [ ] Test(s) added

### DIFF
--- a/.changeset/tasty-snakes-bow.md
+++ b/.changeset/tasty-snakes-bow.md
@@ -1,0 +1,6 @@
+---
+'@finos/legend-query-builder': patch
+'@finos/legend-graph': patch
+---
+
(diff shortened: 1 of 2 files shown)


MauricioUyaguari:
@hardikmaheshwari we dont need to do any of these changes. Studio still needs to work if there is a binary property defined. We want to just remove it from our dropdown. 
The only places we need to change is the method `usableClassPropertyTypes` in 
`GraphManagerState`. 
We can add binary to this filter 
https://github.com/finos/legend-studio/blob/master/packages/legend-graph/src/graph-manager/GraphManagerState.ts#L165

PR #2355, opened 2023-06-21
use existing query metadata when saving to new query

MauricioUyaguari:

- [ ] Test(s) added

### DIFF
--- a/packages/legend-application-query/src/components/QueryEditor.tsx
+++ b/packages/legend-application-query/src/components/QueryEditor.tsx
@@ -88,10 +88,7 @@ import { QUERY_EDITOR_TEST_ID } from '../__lib__/LegendQueryTesting.js';
 const CreateQueryDialog = observer(() => {
   const editorStore = useQueryEditorStore();
   const createQueryState = editorStore.queryCreatorState;
-  const close = (): void => {
-    createQueryState.setShowCreateModal(false);
-    createQueryState.editorStore.setExistingQueryName(undefined);
-  };
+  const close = (): void => createQueryState.close();
   const applicationStore = useApplicationStore();
   const create = (): void => {
     flowResult(createQueryState.createQuery()).catch(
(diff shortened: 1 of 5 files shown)


packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -84,103 +84,107 @@ import { QUERY_DOCUMENTATION_KEY } from '../application/LegendQueryDocumentation
...
-  const createQueryState = editorStore.queryCreatorState;
-  const close = (): void => {
-    createQueryState.setShowCreateModal(false);
-    createQueryState.editorStore.setExistingQueryName(undefined);
-  };
-  const applicationStore = useApplicationStore();
-  const create = (): void => {
-    flowResult(createQueryState.createQuery()).catch(
-      applicationStore.alertUnhandledError,
+const CreateQueryDialog = observer(
```
akphi, line 89: See my comment in `QueryCreatorState`, I think we pass in the query eagerly than passing in the component here

akphi (changes requested):

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -435,6 +435,9 @@ const QueryEditorHeaderContent = observer(
 
     const handleQuerySaveAs = (): void => {
       editorStore.queryCreatorState.setShowCreateModal(true);
+      if (editorStore instanceof ExistingQueryEditorStore) {
```
akphi, line 438: why don't we set this when we initialize the state for `ExistingQueryEditorStore` instead of doing this here every time user hits the `Save As` button?

akphi (changes requested):

akphi (approved):

packages/legend-application-query/src/components/QueryEditor.tsx
akphi: Let's be slightly cleaner, let's create 2 methods: `openModal` and `close`

```ts
open(originalQuery?: Query | undefined)
close()
```

akphi (approved):

PR #2360, opened 2023-06-22
query filter & post-filter dropTargetConnector

xannem:
## Summary

Adding new `dropTargetConnector` to post-filter and filter panel in query so that users can continuously DND 

Kapture 2023-06-22 at 11 12 04

- [ ] Test(s) added

### DIFF
--- a/packages/legend-query-builder/src/components/filter/QueryBuilderFilterPanel.tsx
+++ b/packages/legend-query-builder/src/components/filter/QueryBuilderFilterPanel.tsx
@@ -898,6 +898,35 @@ export const QueryBuilderFilterPanel = observer(
       [handleDrop],
     );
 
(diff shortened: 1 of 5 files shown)


packages/legend-query-builder/src/components/filter/QueryBuilderFilterPanel.tsx
```diff
@@ -898,6 +898,35 @@ export const QueryBuilderFilterPanel = observer(
       [handleDrop],
     );
 
+    const [{ addFilterDragOver }, filterTargetConnector] = useDrop<
```
akphi, line 901: I believe `dropTargetConnector` is a special ref function, so you can do something like `dropTargetConnector(refForBlankNode)` - so no need to repeat code like this

akphi (changes requested):

akphi (approved):

PR #2373, opened 2023-06-26
Add new query option 

xannem:
## Summary
add button to create new query 
Kapture 2023-06-26 at 14 37 19

- [x] Test(s) added

### DIFF
--- a/.changeset/silent-stilts-shush.md
+++ b/.changeset/silent-stilts-shush.md
@@ -1,0 +1,5 @@
+---
+'@finos/legend-query-builder': patch
+---
+
+Add option to create new query in query header actions
--- a/.changeset/silent-stilts-shush.md
+++ b/.changeset/silent-stilts-shush.md
@@ -1,5 +1,3 @@
 ---
 '@finos/legend-query-builder': patch
 ---
-
-Add option to create new query in query header actions
--- a/packages/legend-application-query/src/components/QueryEditor.tsx
+++ b/packages/legend-application-query/src/components/QueryEditor.tsx
@@ -420,13 +420,12 @@ const QueryEditorHeaderContent = observer(
     };
 
     const newQuery = (): void => {
-      console.log('svpp');
       editorStore.queryBuilderState?.resetQueryResult();
       editorStore.queryBuilderState?.resetQueryContent();
       editorStore.queryBuilderState?.setisNewQuery(true);
 
       if (editorStore instanceof ExistingQueryEditorStore) {
-        editorStore.setLightName('');
+        editorStore.setLightQueryName('');
       }
     };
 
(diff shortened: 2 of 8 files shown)


.changeset/silent-stilts-shush.md
```diff
@@ -0,0 +1,5 @@
+---
+'@finos/legend-query-builder': patch
+---
+
+Add option to create new query in query header actions
```
gayathrir11, line 5: this changeset description would be more relevant to legend-application-query. legend-query-builder doesn't really have the concept of creating new query

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -414,6 +419,41 @@ const QueryEditorHeaderContent = observer(
       }
     };
 
+    const newQuery = (): void => {
+      console.log('svpp');
```
gayathrir11, line 423: can you please remove this

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -500,11 +540,21 @@ const QueryEditorHeaderContent = observer(
...
             </div>
           </Button>
-
+          <Button
+            className="query-editor__header__action btn--dark"
+            disabled={editorStore.isPerformingBlockingAction}
+            onClick={cautionNewQuery}
+            title="New query"
+          >
+            <FileEmptyIcon />
```
gayathrir11, line 549: Not sure about the icon choice here

packages/legend-application-query/src/stores/QueryEditorStore.ts
```diff
@@ -965,6 +966,12 @@ export class ExistingQueryEditorStore extends QueryEditorStore {
     this.query = val;
   }
 
+  setLightName(val: string): void {
```
gayathrir11, line 969: I thinksetlightQueryName would be more more appropriate name for this function

.changeset/silent-stilts-shush.md
xannem: True!

packages/legend-application-query/src/components/QueryEditor.tsx
xannem: me neither, @akphi?

akphi (changes requested):
I think the right thing to do here is to find a way to reset the `QueryEditorStore`. Your current approach is not clean in that it hacks the `ExistingQueryEditorStore`, and the URL is not updated properly.

If we have a way to swap the `ExistingQueryEditorStore` with a `MappingQueryCreatorStore`, and preserve the `GraphManagerState` that would be ideal because we save time by not rebuilding the graph.

But another challenge is that we need to be adaptive here since the query is created from a dataspace, i.e. when we swap out the `QueryEditorStore`, we need to create `DataSpaceQueryCreatorStore` not `MappingQueryCreatorStore`, and we must not re-initialize the graph

I think this is a design-limitation on our end, @MauricioUyaguari and I will see what we should go about here, but thank you so much for attempting it!

PR #2378, opened 2023-06-27
query results table improvements

xannem:
## Summary

- [x] When filtering a cell now, you don't have to select the cell first and then right click to bring up the filter actions; the filter actions will directly correspond to the selection or cell you're hovering over 
Kapture 2023-06-27 at 12 15 16
- [x] Originally when editing preview limit, users would have defaulted to 1 when pressing backspace (for example if you wanted to change preview limit from 1000 to 2000, as soon as you pressed the backspace of 1 it would default back to 1. To get around this you would have had to move your cursor, do 12000 and then backspace the 1 so that it was a valid number). Instead PR updates this so that the preview limit only becomes a valid entry (1) when the user exits the input box to make it expected behavior/easier for them to input the value they would like 
Kapture 2023-06-27 at 12 18 16

- [ ] Test(s) added

### DIFF
--- a/packages/legend-query-builder/src/stores/QueryBuilderResultState.ts
+++ b/packages/legend-query-builder/src/stores/QueryBuilderResultState.ts
@@ -37,7 +37,6 @@ import {
   buildRawLambdaFromLambdaFunction,
   reportGraphAnalytics,
   extractExecutionResultValues,
-  TDSExecutionResult,
 } from '@finos/legend-graph';
 import { buildLambdaFunction } from './QueryBuilderValueSpecificationBuilder.js';
 import { DEFAULT_TAB_SIZE } from '@finos/legend-application';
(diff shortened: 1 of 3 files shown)


packages/legend-query-builder/src/stores/QueryBuilderResultState.ts
```diff
@@ -187,6 +187,15 @@ export class QueryBuilderResultState {
     return this.executionResult.result.rows[rowIndex]?.values ?? [''];
   };
 
+  findIsCoordinatesSelected = (
```
akphi, line 190:
```suggestion
  isCoordinatesSelected = (
```

also, this kind of function can easily be put in the renderer

akphi: let's move all these 4 `find...` functions into component

akphi (changes requested):

akphi (approved):

PR #2430, opened 2023-07-14
Handle exists logic correctly in query

gayathrir11:

- [ ] Test(s) added

***Multiple conditions in single exists***

https://github.com/finos/legend-studio/assets/87973126/a946def6-269e-41f3-b7ae-7739e2fcb45b

***Multiple exists in single group operation***

https://github.com/finos/legend-studio/assets/87973126/38a815ec-b260-4766-9704-0419b5a6e687

***Nested exists***

https://github.com/finos/legend-studio/assets/87973126/622a08c3-24fe-48d0-8a90-534e6ee606bb

***Nested grouping inside exists***

https://github.com/finos/legend-studio/assets/87973126/a8d7d956-4555-4a6c-a7ba-3a3de0c88937

### DIFF
--- a/packages/legend-query-builder/src/stores/filter/QueryBuilderFilterStateBuilder.ts
+++ b/packages/legend-query-builder/src/stores/filter/QueryBuilderFilterStateBuilder.ts
@@ -115,33 +115,191 @@ const processFilterTree = (
       expression.parametersValues[0],
       AbstractPropertyExpression,
     );
-    assertTrue(
+    if (
       propertyExpression.func.value.multiplicity.upperBound === undefined ||
-        propertyExpression.func.value.multiplicity.upperBound > 1,
-      `Can't process filter expression: exists is only supported for propertyExpression with multiplicity greater than 1`,
-    );
-    const existsNode = new QueryBuilderFilterTreeExistsNodeData(
-      parentFilterNodeId,
-    );
-    const lambdaFunctionInstance = guaranteeType(
-      expression.parametersValues[1],
-      LambdaFunctionInstanceValue,
-      `Can't process filter expression: only supports exists with second paramter as LambdaFunctionInstanceValue`,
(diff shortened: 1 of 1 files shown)


packages/legend-query-builder/src/stores/filter/QueryBuilderFilterStateBuilder.ts
```diff
@@ -70,6 +106,200 @@ const processFilterTree = (
...
+        SimpleFunctionExpression,
+      );
+      existsNode.setPropertyExpression(propertyExpression);
+      existsNode.lambdaParameterName =
+        lambdaFunction.functionType.parameters[0]?.name;
+      filterState.nodes.set(existsNode.id, existsNode);
+      processFilterTree(filterExpression, filterState, existsNode.id);
+      filterState.addNodeFromNode(existsNode, parentNode);
+    } else {
+      // Build property chain if we have exists filter() and the multiplicity of the property
```
MauricioUyaguari, line 145: are we rewriting the users query here ? 
Seems like we are flattening their exists lambda if they don't have it. 
Not sure if we should do this. The creation of the exists lambda should be flatten as we are guiding users to do this but if the user's query has it in this format i dont think we should override their query.

gayathrir11: yes we are rewriting user's query here. Seems like it's something we do support in query today. We rewrite the exists for users if the multiplicity of property is less than or equal to 1. Rationale @akphi gave me for this is it is helpful when people change their model multiplicity.

MauricioUyaguari: will keep it as if for now.

MauricioUyaguari (approved):

PR #2501, opened 2023-08-11
feat: support for auto-insert package suggestions in Studio text mode

mihir1021:
## Summary

In Studio text mode, when users encounter a compilation error in which an element's type can't be found, they must manually resolve it. This process involves searching for definition(s) of the problematic element, determining which path to use, and then copy pasting the selected path before the problematic element.

Upon a can't find element error now, users are prompted with a set of candidate elements that match the problematic element under a new Suggestions tab that pops up when Engine returns a code fix suggestion and remains as long as the suggestion is active (this update is made in this PR). Each candidate row contains the candidate's full path, source information, and an arrow button that will take the user to the candidate's definition within the code editor on click. Once the user determines which candidate they are referring to, they can click on the candidate's row in order to automatically insert the candidate's full path in front of the problematic element. This feature enables the user to resolve such errors in one click!

- [ ] Test(s) added


akphi:
closing in favor of #2502

PR #2543, opened 2023-08-23
Default to only showing tables in the source dropdown when drag and drop store to mapping source

YannanGao-gs:
## Summary

Default to only showing tables in the source dropdown when drag and drop store to mapping source

- [ ] Test(s) added

before: 

https://github.com/finos/legend-studio/assets/73408381/b0b051d8-5221-41b4-95d5-407e920bbec4

After: 

https://github.com/finos/legend-studio/assets/73408381/73183e5e-b8cb-44f8-b596-485511db158e

### DIFF
--- a/packages/legend-application-studio/src/components/editor/editor-group/mapping-editor/InstanceSetImplementationEditor.tsx
+++ b/packages/legend-application-studio/src/components/editor/editor-group/mapping-editor/InstanceSetImplementationEditor.tsx
@@ -101,6 +101,9 @@ export const InstanceSetImplementationSourceExplorer = observer(
     const { setImplementation, isReadOnly } = props;
     const editorStore = useEditorStore();
     const applicationStore = useApplicationStore();
+    const [sourceElementToFilter, setSourceElementToFilter] = useState<
+      PackageableElement | undefined
+    >(undefined);
     const mappingEditorState =
       editorStore.tabManagerState.getCurrentEditorState(MappingEditorState);
     const instanceSetImplementationState =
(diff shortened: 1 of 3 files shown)


packages/legend-application-studio/src/components/editor/editor-group/mapping-editor/InstanceSetImplementationEditor.tsx
```diff
@@ -101,6 +101,8 @@ export const InstanceSetImplementationSourceExplorer = observer(
     const { setImplementation, isReadOnly } = props;
     const editorStore = useEditorStore();
     const applicationStore = useApplicationStore();
+    const [relationalMappingStrictMode, setRelationalMappingStrictMode] =
```
MauricioUyaguari, line 104: dont think this naming or the fact we reference relational in the instance set implementation. 
Could we think of another way to do this please. 
Maybe have a filter function here with the Mapping Source as input ?

YannanGao-gs: Screenshot 2023-08-24 at 10 33 03 AM
As the screenshot shows, this flag is used when composing the dropdown. The flag is only set to true when the source is of type database

YannanGao-gs: I updated to code to pass in the sourceElement which could be used to compose the dropdown

MauricioUyaguari (approved):

PR #2579, opened 2023-09-09
Persist seconds field when zero seconds are provided for DateTime field in filter and post-filter values

### LINKED ISSUE #2580: Feature request: Persist seconds field when zero seconds are provided for DateTime field in filter and post-filter values
### Similar requests

### How are you using Legend?

Legend Studio

Persist seconds field when zero seconds are provided for DateTime field in filter and post-filter values

Currently when user don't specify seconds for DateTime we don't throw any error and when user selects zero seconds it is removed by HTML input when returning the changed value. We should append zero seconds when HTML removes it 

_No response_

_No response_

### Contribution

- [x] I would like to work on this feature

gayathrir11:
## Summary

closes #2580 

- [x] Test(s) added

***Selecting zero seconds from editor***

https://github.com/finos/legend-studio/assets/87973126/9b0331d8-a1b6-4be0-9484-2236c178b401

### DIFF
--- a/packages/legend-query-builder/src/components/shared/CustomDatePicker.tsx
+++ b/packages/legend-query-builder/src/components/shared/CustomDatePicker.tsx
@@ -840,22 +840,21 @@ const AbsoluteTimeValueSpecificationEditor: React.FC<{
   const updateAbsoluteTimeValue: React.ChangeEventHandler<HTMLInputElement> = (
     event,
   ) => {
+    //
+    const value = new Date(event.target.value).getUTCSeconds()
+      ? event.target.value
+      : `${event.target.value}:00`;
     if (valueSpecification instanceof SimpleFunctionExpression) {
       setValueSpecification(
         buildPrimitiveInstanceValue(
           graph,
           PRIMITIVE_TYPE.DATETIME,
-          event.target.value,
+          value,
           observerContext,
         ),
       );
     } else {
-      instanceValue_setValue(
-        valueSpecification,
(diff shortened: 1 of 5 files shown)


MauricioUyaguari:
I know the initial task was to throw an error but I am a little concerned here of impact here.

Could we separate this into 2 PRs. I am good with the changes in custom picker where we add the seconds when its datetime. I think thats the main change here thats needed alongside the relevant tests. 

The other PR could be on throwing an error and showing unsupported mode. We should think if we want to do this. 
Does the engine compiler throw an error currently if the user does this? Main Concern is if user has a query with this filter already, what should our approach be in this case? maybe add the seconds ? 

We can discuss that, but to prevent hold up on the main task lets go with the other changes.

packages/legend-query-builder/src/components/shared/CustomDatePicker.tsx
```diff
@@ -840,22 +840,21 @@ const AbsoluteTimeValueSpecificationEditor: React.FC<{
   const updateAbsoluteTimeValue: React.ChangeEventHandler<HTMLInputElement> = (
     event,
   ) => {
+    //
+    const value = new Date(event.target.value).getUTCSeconds()
```
MauricioUyaguari, line 844: wont this always return 0, even they have specified the seconds ? 
`new Date(event.target.value).getUTCSeconds() `

MauricioUyaguari (approved):

PR #2589, opened 2023-09-13
Fix default milestoning dates getting added as parameters when they are defined as constants

gayathrir11:

- [x] Test(s) added

https://github.com/finos/legend-studio/assets/87973126/9627b1d9-7227-428b-a125-a061afc34849

### DIFF
--- a/packages/legend-query-builder/src/stores/__tests__/QueryBuilderBuildLambdaRoundtrip.test.tsx
+++ b/packages/legend-query-builder/src/stores/__tests__/QueryBuilderBuildLambdaRoundtrip.test.tsx
@@ -139,6 +139,8 @@ import {
 } from './TEST_DATA__QueryBuilder_Calendar.js';
 import { DEFAULT_LIMIT } from '../QueryBuilderResultState.js';
 import TEST_DATA__QueryBuilder_Model_SimpleRelational from '../../stores/__tests__/TEST_DATA__QueryBuilder_Model_SimpleRelational.json';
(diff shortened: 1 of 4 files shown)


packages/legend-query-builder/src/stores/__tests__/QueryBuilderBuildLambdaRoundtrip.test.tsx
```diff
@@ -668,6 +674,12 @@ const cases: RoundtripTestCase[] = [
     TEST_DATA__lambda_isOperatorFilterForDate,
     undefined,
   ],
+  [
+    'Milestoning',
+    milestoningCtx,
```
MauricioUyaguari, line 679: lets give this test a better description

packages/legend-query-builder/src/stores/milestoning/QueryBuilderMilestoningState.ts
```diff
@@ -155,6 +155,9 @@ export class QueryBuilderMilestoningState implements Hashable {
     if (
       !this.queryBuilderState.parametersState.parameterStates.find(
         (p) => p.variableName === parameterName,
+      ) &&
```
MauricioUyaguari, line 158: lets also look anywhere else in the code base where we look for variables to ensure we have looked inside the constant states as well.

gayathrir11: I don't think we have this kind of logic anywhere else. Checked it too. Milestoning is an excpetion because we magically create parameters for them to use in the query

MauricioUyaguari (approved):

PR #2612, opened 2023-09-19
fix typeahead search for milestoned properties

gayathrir11:

- [x] Test(s) added

### DIFF
--- a/packages/legend-query-builder/src/stores/QueryBuilderTypeaheadHelper.ts
+++ b/packages/legend-query-builder/src/stores/QueryBuilderTypeaheadHelper.ts
@@ -28,6 +28,25 @@ import {
   matchFunctionName,
   PrimitiveType,
   type ObserverContext,
+  Multiplicity,
+  getMilestoneTemporalStereotype,
+  extractElementNameFromPath,
+  CollectionInstanceValue,
+  GenericTypeExplicitReference,
+  GenericType,
+  CORE_PURE_PATH,
+  FunctionType,
+  LambdaFunction,
+  PackageableElementExplicitReference,
+  PropertyExplicitReference,
+  getAllClassDerivedProperties,
+  VariableExpression,
+  buildRawLambdaFromLambdaFunction,
+  type Type,
+  INTERNAL__UnknownValueSpecification,
+  V1_transformRawLambda,
+  V1_GraphTransformerContextBuilder,
+  V1_serializeRawValueSpecification,
 } from '@finos/legend-graph';
 import {
   getNullableFirstEntry,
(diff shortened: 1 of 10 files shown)


packages/legend-query-builder/src/stores/QueryBuilderTypeaheadHelper.ts
```diff
@@ -54,6 +69,57 @@ const initializeQueryBuilderState = (
...
+  let currentExpression: ValueSpecification | undefined = propertyExpression;
+  while (currentExpression instanceof AbstractPropertyExpression) {
+    // check if the property is milestoned
+    if (
+      currentExpression.func.value.genericType.value.rawType instanceof Class &&
+      currentExpression.func.value._OWNER._generatedMilestonedProperties
+        .length !== 0 &&
+      currentExpression.func.value instanceof DerivedProperty &&
+      !currentExpression.func.value._OWNER.derivedProperties.includes(
+        currentExpression.func.value,
```
MauricioUyaguari, line 85: can you explain when this last condition will be true?

packages/legend-query-builder/src/stores/QueryBuilderTypeaheadHelper.ts
```diff
@@ -54,6 +69,57 @@ const initializeQueryBuilderState = (
...
+    ) {
+      const parameterValue = createSupportedFunctionExpression(
+        QUERY_BUILDER_SUPPORTED_FUNCTIONS.NOW,
+        PrimitiveType.DATETIME,
+      );
+      const parameterValues =
+        currentExpression.parametersValues.length === 2
+          ? [parameterValue]
+          : [parameterValue, parameterValue];
+      functionExpression_setParametersValues(
```
MauricioUyaguari, line 96: i have a major concern here
arent we replacing the parameter value of this derived property in real time. so we are changing the actual value of the user's query when we create the typeahead query. is that correct?

packages/legend-query-builder/src/stores/QueryBuilderTypeaheadHelper.ts
gayathrir11: When the property is milestoned and it is actually a derived property.

packages/legend-query-builder/src/stores/QueryBuilderTypeaheadHelper.ts
gayathrir11: I missed this. Thanks for pointing this!! This would actually modify the existing values. I think will create a new AbstractPropertyExpression instead of modifying it.

gayathrir11: Updated the test cases to catch such scenarios

MauricioUyaguari (approved):

PR #2648, opened 2023-09-27
Update dev console with feature to set dev protocol version

AFine-gs:
Verified that the new toggle sets and clears the dev protocol on client version property in execute input
image

### DIFF
--- a/.changeset/yellow-parents-wash.md
+++ b/.changeset/yellow-parents-wash.md
@@ -1,0 +1,6 @@
+---
+'@finos/legend-application-studio': patch
+'@finos/legend-graph': patch
+---
+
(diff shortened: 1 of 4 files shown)


.changeset/yellow-parents-wash.md
```diff
@@ -0,0 +1,6 @@
+---
+'@finos/legend-application-studio': patch
+'@finos/legend-graph': patch
```
YannanGao-gs, line 3: could you please split these two changesets? 
please keep description of the changeset that contains '@finos/legend-graph': patch empty

MauricioUyaguari (approved):

PR #2677, opened 2023-10-06
Protocol and metamodel changes for Mastery Runtime

agshubh-97:

- [ ] Test(s) added


.changeset/big-tables-hang.md
```diff
@@ -0,0 +1,5 @@
+---
+'@finos/legend-extension-dsl-mastery': patch
+---
+
+Protocol and Metamodel changes for Mastery Runtime
```
gayathrir11, line 5: Can you please modify this `Add support for Mastery Runtime`

packages/legend-extension-dsl-mastery/src/graph-manager/protocol/pure/v1/model/packageableElements/mastery/V1_DSL_Mastery_Runtime.ts
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Hashable, hashArray } from '@finos/legend-shared';
+import { MASTERY_HASH_STRUCTURE } from '../../../../../../../graph/DSL_Mastery_HashUtils';
+import { V1_PackageableElement } from '@finos/legend-graph';
+
+export abstract class V1_Runtime
+  extends V1_PackageableElement
```
gayathrir11, line 22: Isn't it called `MasteryRuntime`? If that's the case then can you please rename it to `MasteryRuntime` everywhere

packages/legend-extension-dsl-mastery/src/graph-manager/protocol/pure/v1/transformation/pureProtocol/V1_DSL_Mastery_ProtocolHelper.ts
```diff
@@ -925,3 +926,50 @@ export const V1_masterRecordDefinitionModelSchema = (
...
+      ).V1_getExtraRuntimeProtocolSerializers?.() ?? [],
+  );
+  for (const serializer of extraRuntimeSerializers) {
+    const runtimeJson = serializer(protocol, plugins);
+    if (runtimeJson) {
+      return runtimeJson;
+    }
+  }
+  throw new UnsupportedOperationError(
+    `Can't serialize runtime Json: no compatible serializer available from plugins`,
```
gayathrir11, line 951: Please modify this to `Can't serialize mastery runtime: no compatible serializer available from plugins

gayathrir11 (changes requested):
Please fix the pipelines and also some tests in `legend-manual-tests` and DSL_Mastery_Roundtrip.test.ts

gayathrir11:
I don't think you have made changes to `DSL_Mastery_PureProtocolProcessorPlugin.ts`

packages/legend-extension-dsl-mastery/src/graph-manager/protocol/pure/v1/model/packageableElements/mastery/V1_DSL_Mastery_Runtime.ts
agshubh-97: The convention that we've used in the rest of the protocols is the same as this one. So, stuck to V1_Runtime. Shall I switch to MasteryRuntime?

agshubh-97:
> I don't think you have made changes to `DSL_Mastery_PureProtocolProcessorPlugin.ts`

I've added changes in DSL_Mastery_PureProtocolProcessorPlugin_Extension as Runtime is abstract and did not have any implementations. Please correct me if I'm missing something.

packages/legend-extension-dsl-mastery/src/graph-manager/protocol/pure/v1/model/packageableElements/mastery/V1_DSL_Mastery_Runtime.ts
gayathrir11: For protocol models can you modify it to `V1_MasteryRuntime` and for metamodels `MasteryRuntime`

gayathrir11:
> > I don't think you have made changes to `DSL_Mastery_PureProtocolProcessorPlugin.ts`
> 
> I've added changes in DSL_Mastery_PureProtocolProcessorPlugin_Extension as Runtime is abstract and did not have any implementations. Please correct me if I'm missing something.

Please take a look at how other functions defined in the interface are getting called. You will need to do something similar so that when any element of type `MasteryRuntime` is found it can be routed to implementations of the function when you implement the plugin

PR #2684, opened 2023-10-06
end to end workflow II - Enable query from connection workflow

YannanGao-gs:
## Summary

end to end workflow II - Enable query from connection workflow

- [ ] Test(s) added

https://github.com/finos/legend-studio/assets/73408381/ec8e4795-8a48-4995-a951-0d84cc454aaf


gayathrir11 (changes requested):

packages/legend-application-studio/src/stores/editor/editor-state/element-editor-state/connection/DatabaseBuilderState.ts
```diff
@@ -350,12 +354,6 @@ export class DatabaseSchemaExplorerState {
   }
 
   get resolveDatabasePackageAndName(): [string, string] {
-    if (!this.isCreatingNewDatabase) {
```
MauricioUyaguari, line 353: why are we removing this. if we are not creating a new db then the package should be there.

packages/legend-application-studio/src/stores/editor/editor-state/element-editor-state/connection/DatabaseBuilderState.ts
```diff
@@ -799,61 +797,77 @@ export class DatabaseSchemaExplorerState {
     }
   }
 
-  *updateDatabase(): GeneratorFn<void> {
+  // this method just updates database
+  *updateDatabase(): GeneratorFn<Database> {
+    this.isUpdatingDatabase = true;
```
MauricioUyaguari, line 802: again why are we removing this ?

packages/legend-application-studio/src/stores/editor/editor-state/element-editor-state/connection/DatabaseBuilderState.ts
```diff
@@ -799,61 +797,77 @@ export class DatabaseSchemaExplorerState {
...
+    const schemas = Array.from(
+      guaranteeNonNullable(this.treeData).nodes.values(),
+    )
+      .map((schemaNode) => {
+        if (schemaNode instanceof DatabaseSchemaExplorerTreeSchemaNodeData) {
+          return schemaNode.schema;
+        }
+        return undefined;
+      })
+      .filter(isNonNullable);
```
MauricioUyaguari, line 824: dont understand why we need this as the generation should do this

packages/legend-application-studio/src/stores/editor/editor-state/element-editor-state/connection/DatabaseModelBuilderState.ts
```diff
@@ -122,11 +128,9 @@ export class DatabaseModelBuilderState {
...
       for (const entity of entities) {
         let entityChangeType: EntityChangeType;
-        if (
-          this.editorStore.graphManagerState.graph.getNullableElement(
-            entity.path,
-          ) === undefined
-        ) {
+        const graph =
+          this.assistanceLightGraph ?? this.editorStore.graphManagerState.graph;
+        if (graph.getNullableElement(entity.path) === undefined) {
```
MauricioUyaguari, line 133: i dont get the point of this `assistanceLightGraph` the only place its used is here. we don't use it below when loading the entities into the graph. What exactly is it you are trying to do in context of this state.

packages/legend-application-studio/src/stores/editor/sidebar-state/end-to-end-workflow/GlobalEndToEndFlowState.ts
```diff
@@ -0,0 +1,598 @@
...
+        this.databaseBuilderWizardState,
+        'databaseBuilderState must exist before building model from database',
+      ).schemaExplorerState.database,
+      false,
+    );
+    this.databaseModelBuilderState.assistanceLightGraph = this.graph;
+  }
+
+  *buildDatabase(): GeneratorFn<void> {
+    const databaseBuilderWizardState = guaranteeNonNullable(
```
MauricioUyaguari, line 213: are we catching any of these errors? we risk going throwing errors in the UI here right?

packages/legend-application-studio/src/stores/editor/sidebar-state/end-to-end-workflow/GlobalEndToEndFlowState.ts
```diff
@@ -0,0 +1,598 @@
...
+        getNonNullableEntry(this.graph.mappings, 0),
+        this.graph.classes,
+      )[0],
+    );
+    const config = new QueryBuilderConfig();
+    const queryBuilderState = new ClassQueryBuilderState(
+      this.globalEndToEndWorkflowState.editorStore.applicationStore,
+      this.globalEndToEndWorkflowState.editorStore.graphManagerState,
+      config,
+    );
```
MauricioUyaguari, line 536: lets set up some tme here to go walk me through this file

packages/legend-application-studio/src/stores/editor/editor-state/element-editor-state/connection/DatabaseBuilderState.ts
YannanGao-gs: I didn't remove this method.  I just separate part of it to a new method and rename the original one..

packages/legend-application-studio/src/stores/editor/editor-state/element-editor-state/connection/DatabaseBuilderState.ts
YannanGao-gs: To support uses to click the 'back' button and this is still editable

packages/legend-application-studio/src/stores/editor/editor-state/element-editor-state/connection/DatabaseModelBuilderState.ts
YannanGao-gs: There is a comment in my code `'This assistanceLightGraph is used to build database models if database doesn't exist in editorStore.graphManagerState.graph'`
For this flow, database is not in the graph yet so I have to use an assistanceLightGraph that contains this database otherwise you will see the error that complaining no such db

packages/legend-application-studio/src/stores/editor/editor-state/element-editor-state/connection/DatabaseBuilderState.ts
YannanGao-gs: without code below,  this.database.package and this.database.name are undefined

YannanGao-gs:
> any reason why there are no tests here?

thought tests will be added in phase 2

packages/legend-application-studio/src/stores/editor/editor-state/element-editor-state/connection/DatabaseModelBuilderState.ts
MauricioUyaguari: i saw the comment still dont understand this. if the light graph is defined you use but when calling the 
`this.editorStore.graphState.loadEntityChangesToGraph` below doesn't that load entities into the main graph?
Also we should be refactoring these states to make sense in isolation of your stepper workflow. 
Therefore if we may consider having a graph input for this state if it make sense.

packages/legend-application-studio/src/stores/editor/editor-state/element-editor-state/connection/DatabaseBuilderState.ts
MauricioUyaguari: did you look into how it would affect where its already used? dont think thats a valid reason, the change should make sense in isolation of where its used but how it makes sense in `DatabaseBuilderState`

packages/legend-application-studio/src/stores/editor/editor-state/element-editor-state/connection/DatabaseBuilderState.ts
MauricioUyaguari: why?

packages/legend-application-studio/src/stores/editor/editor-state/element-editor-state/connection/DatabaseModelBuilderState.ts
YannanGao-gs: there is a mistake. saveModels() is not used. will remove the changes. 
this graph is used in previewDatabaseModels

packages/legend-application-studio/src/stores/editor/editor-state/element-editor-state/connection/DatabaseBuilderState.ts
YannanGao-gs: For the original feature, the difference would be the target path is always editable, which kind of makes senes to me..
if users want to create a new db with a new path name, now this is doable.
If users just want to update the original db, then just keep the value as what it is before

YannanGao-gs:
closed by https://github.com/finos/legend-studio/pull/2705

PR #2692, opened 2023-10-12
Add query workflow information as a part of telemetry data we send to analytics

gayathrir11:

- [ ] Test(s) added

### DIFF
--- a/packages/legend-query-builder/src/__lib__/QueryBuilderTelemetryHelper.ts
+++ b/packages/legend-query-builder/src/__lib__/QueryBuilderTelemetryHelper.ts
@@ -21,13 +21,16 @@ import {
   QUERY_BUILDER_FILTER_EVENT,
   QUERY_BUILDER_POST_FILTER_EVENT,
 } from './QueryBuilderEvent.js';
+import type { QueryStateInfo } from '../stores/QueryBuilderState.js';
 
 type QueryExecution_TelemetryData = GraphManagerOperationReport & {
   dependenciesCount: number;
+  state?: QueryStateInfo | undefined;
 };
 type QueryMappingModelCoverageAnalysis_TelemetryData =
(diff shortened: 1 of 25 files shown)


packages/legend-query-builder/src/__lib__/QueryBuilderTelemetryHelper.ts
```diff
@@ -24,10 +24,12 @@ import {
 
 type QueryExecution_TelemetryData = GraphManagerOperationReport & {
   dependenciesCount: number;
+  state?: object | undefined;
```
MauricioUyaguari, line 27: Don't like this type being an object here. Is this suppose to be the querySource or querySDLC info here ? 
Can we create a type here that makes sense. or would we reach some circular dependencies?

packages/legend-application-studio/src/stores/editor/EditorStore.ts
```diff
@@ -337,6 +350,24 @@ export class EditorStore implements CommandRegistrar {
     this.showSearchElementCommand = val;
   }
 
+  getSourceInfo(): ProjectQuerySDLC | undefined {
+    if (this.mode === EDITOR_MODE.VIEWER) {
+      return {
+        projectId: this.sdlcState.activeProject.projectId,
```
MauricioUyaguari, line 356: should we maybe put this in the EditorMode class ? 
This would break in showcaseeditorMode as well as when viewer mode is using gav coordinates right???

MauricioUyaguari: Lets be careful using using methods that throw here. most likely we should safely call this method and return undefined if we can't calculate source

MauricioUyaguari (changes requested):

MauricioUyaguari (approved):

PR #2696, opened 2023-10-13
Unlock Ag grid enterprise features in query

gayathrir11:

- [ ] Test(s) added


packages/legend-query-builder/src/graph-manager/QueryBuilderConfig.ts
```diff
@@ -23,9 +23,15 @@ export class QueryBuilderConfig {
...
 
+  /**
+   * This flag is to enable enterprise features of ag-grid
+   */
+  TEMPORARY__enableAdvancedMode = false;
+
   static readonly serialization = new SerializationFactory(
     createModelSchema(QueryBuilderConfig, {
       TEMPORARY__disableQueryBuilderChat: optional(primitive()),
+      TEMPORARY__enableAdvancedMode: optional(primitive()),
```
MauricioUyaguari, line 34: lets do `enableAdancedGridMode`

packages/legend-query-builder/src/components/QueryBuilderResultPanel.tsx
```diff
@@ -740,6 +777,31 @@ const QueryBuilderGridResult = observer(
     const { executionResult, queryBuilderState } = props;
 
     const resultState = queryBuilderState.resultState;
+    const isAdvancedModeEnabled = queryBuilderState.isAdvancedModeEnabled;
+    const colDefs = executionResult.result.columns.map(
```
MauricioUyaguari, line 781: for this could we create two separate functions for advanced and simplle definitions. would be cleaner than having flags in each config

akphi (dismissed):

packages/legend-query-builder/src/components/QueryBuilderResultPanel.tsx
```diff
@@ -764,31 +862,35 @@ const QueryBuilderGridResult = observer(
         >
           <DataGrid
```
MauricioUyaguari, line 863: could we something similar here and separate the component 
something like `isAdvancedMode ? <datagrid options={advanced} : <datagrid options={normal}`.

packages/legend-query-builder/src/stores/fetch-structure/QueryBuilderFetchStructureState.ts
```diff
@@ -64,6 +64,7 @@ export class QueryBuilderFetchStructureState {
       }
       case FETCH_STRUCTURE_IMPLEMENTATION.GRAPH_FETCH: {
         this.queryBuilderState.setIsCalendarEnabled(false);
+        this.queryBuilderState.setIsAdvancedModeEnabled(false);
```
MauricioUyaguari, line 67: is there logic to turn this back when we toggle back to tds?

gayathrir11: the mode will be turned off when they toggle back to tds. When they are in graph fetch space advanced mode button would be disabled, when they switch back to tds the button would be editable again

MauricioUyaguari (approved):

PR #2697, opened 2023-10-13
fix the bug where DnD from projection column with a special lambda name generates wrong pure code

YannanGao-gs:
## Summary

fix the bug where DnD from projection column with a special lambda name generates wrong pure code

- [x] Test(s) added

before: 

https://github.com/finos/legend-studio/assets/73408381/9d368e20-4215-46ae-928e-471ab94368c6

after: 

https://github.com/finos/legend-studio/assets/73408381/15b0a359-c2df-49d4-8723-dd0358e29784

### DIFF
--- a/packages/legend-query-builder/src/components/filter/QueryBuilderFilterPanel.tsx
+++ b/packages/legend-query-builder/src/components/filter/QueryBuilderFilterPanel.tsx
@@ -76,6 +76,7 @@ import {
   assertErrorThrown,
   assertTrue,
   debounce,
+  deepClone,
   generateEnumerableNameFromToken,
   getNullableFirstEntry,
   guaranteeNonNullable,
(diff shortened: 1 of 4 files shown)


packages/legend-query-builder/src/components/filter/QueryBuilderFilterPanel.tsx
```diff
@@ -247,9 +273,9 @@ export const buildFilterTreeWithExists = (
       const parentPropertyChainIndex = existsLambdaPropertyChains.findIndex(
         (p) =>
           p instanceof AbstractPropertyExpression &&
-          p.func.value ===
-            targetDropNode.propertyExpressionState.propertyExpression.func
-              .value &&
+          p.func.value.hashCode ===
+            targetDropNode.propertyExpressionState.propertyExpression.func.value
+              .hashCode &&
```
gayathrir11, line 278: I don't think we should use hashcode to check the equality. Do you know where the inequality is happening. I remember that it used to work without deepclone

YannanGao-gs: Here we are comparing two Property. they have same values but different reference.

YannanGao-gs: hashcode is generated upon  👀

gayathrir11: Got it. I just feel hashcode is something we use to do change detection. We generally don't use it do these kinds of equality checks across the codebase If deepClone is failing this check to preserve the reference of property we can try using `buildPropertyExpressionChain` function to build the new expression. This was you hold no references to the property expression present in project. @MauricioUyaguari  what do you think?

PR #2704, opened 2023-10-18
Fix changing dataspace for an existing query saving it in inconsistent state

gayathrir11:
## Summary

When user tries to change the dataspace for an existing query
1. If user is the owner, we ask user to abort or update the current query and proceed. Updating the current query would save the state of current query and redirect them to create a query from the new dataspace they have selected.
2. If user is not owner, it would just show them a popup saying doing this they would lose changes and if they choose to proceed they will be redirected to create a new query from the dataspace they have selected

- [ ] Test(s) added

https://github.com/finos/legend-studio/assets/87973126/3c7fabde-623c-4eff-aec9-b7b1c0ee24c3

### DIFF
--- a/.changeset/seven-phones-search.md
+++ b/.changeset/seven-phones-search.md
@@ -1,0 +1,5 @@
+---
+'@finos/legend-extension-dsl-data-space': patch
+---
+
+Fix changing dataspace for an existing query saving it in inconsistent state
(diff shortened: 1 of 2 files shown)


YannanGao-gs:
With this, it seems that users (owners) can't easily switch to another dataspace without saving the changes.
how should we handle the case where users want to switch dataspace but don't want to save the changes as well?  👀

MauricioUyaguari:
@gayathrir11 will approve as its better than what we currently have but I think we should most likely do the following. 
We probably need a flowchart but basically,
we should just have a warning since users will lose unsafe changes if they change dataspaces, but that shouldnt move to a new view. we should be able to change and when we save/save as we should handle the saving of dataspaces. can we look into seeing if we can add tests in that realm so we can make those changes.

MauricioUyaguari (approved):

PR #2804, opened 2023-11-27
Minor function activator improvements

YannanGao-gs:
## Summary

Minor function activator improvements

- [ ] Test(s) added

before:

Screenshot 2023-11-27 at 9 45 17 AM

after: 

Screenshot 2023-11-29 at 3 12 54 PM

### DIFF
--- a/packages/legend-art/src/input/Input.tsx
+++ b/packages/legend-art/src/input/Input.tsx
@@ -41,7 +41,7 @@ export const InputWithInlineValidation: React.FC<
         })}
         {...inputProps}
       />
-      {showEditableIcon && (
+      {showEditableIcon && !error && (
         <div className="input--with-validation__editable" title={error}>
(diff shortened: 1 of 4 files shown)


gayathrir11 (approved):

packages/legend-art/src/input/Input.tsx
```diff
@@ -41,7 +41,7 @@ export const InputWithInlineValidation: React.FC<
         })}
         {...inputProps}
       />
-      {showEditableIcon && (
+      {showEditableIcon && !error && (
```
MauricioUyaguari, line 44: confused about this change in general and what we are doing here. 
first we ensure !error but use error in the component. 
Second do we want this, i think we should make the contract of the component clear to the users. 
If the input has an error shouldnt the edit icon still show to allow users to be able to edit it and fix the input. Regardless seems like we need to make this component clearer

PR #2808, opened 2023-11-30
Support getAllVersions() and getAllVersionsInRange() for milestoned root class

gayathrir11:

- [x] Test(s) added

***AllVersions***

https://github.com/finos/legend-studio/assets/87973126/7e175046-beeb-4879-97d0-a33990e18625

***AllVersionsInRange***

https://github.com/finos/legend-studio/assets/87973126/53ead727-d644-444b-b4f8-dd05c6717f65


packages/legend-query-builder/src/stores/QueryBuilderValueSpecificationBuilder.ts
```diff
@@ -158,6 +182,39 @@ export const buildLambdaFunction = (
       Multiplicity.ONE,
     );
     lambdaFunction.expressionSequence[0] = getAllVersionsFunction;
+  } else if (
```
YannanGao-gs, line 185: This looks good! But i feel like maybe combine the first two conditions could be a good option (person preference)
 if (options?.useAllVersionsForMilestoning ||  queryBuilderState.getAllFunction ===
    QUERY_BUILDER_SUPPORTED_GET_ALL_FUNCTIONS.GET_ALL_VERSIONS) {
      if (milestoningStereotype) {
      const getAllVersionsFunction = buildGetAllVersionsFunction(
        _class,
        Multiplicity.ONE,
      );
      lambdaFunction.expressionSequence[0] = getAllVersionsFunction;
    } else {
      throw new UnsupportedOperationError(
        `Unable to build query lamdba: getAllVersions() expects source class to be milestoned`,
      );
    }
  }

gayathrir11: I think i would just use switch case for all the `else if` conditions. Would be more cleaner

gayathrir11:
> Hello @gayathrir11, I might do something wrong but it seems like i triggered an issue for resetting parameters
> 
>  Screen.Recording.2023-12-01.at.12.45.01.PM.mov

fixed it!!

packages/legend-query-builder/src/stores/milestoning/QueryBuilderMilestoningState.ts
```diff
@@ -86,8 +111,185 @@ export class QueryBuilderMilestoningState implements Hashable {
...
+      this.initialiseAllVersionsInRangeParameters();
+    } else {
+      this.queryBuilderState.setGetAllFunction(
+        QUERY_BUILDER_SUPPORTED_GET_ALL_FUNCTIONS.GET_ALL_VERSIONS,
+      );
+      this.clearAllVersionsInRangeParameters();
+    }
+  }
+
+  toggleAllVersions(val: boolean | undefined): void {
```
MauricioUyaguari, line 267: these methods arent really toggling, they are more setting the value right? or handling the seting of the turning on/off the feature. can we rename

packages/legend-query-builder/src/components/explorer/QueryBuilderMilestoningEditor.tsx
```diff
@@ -156,7 +137,12 @@ const BiTemporalMilestoningEditor = observer(
           </div>
           <MilestoningParameterEditor
             queryBuilderState={queryBuilderState}
```
MauricioUyaguari, line 139: bit concernered here with the usage of `guaranteeNonNullable` within the component. 
Have we caught where the nullable values are possible. Doing this in the component causes a total wipeout of the UI which is very bad UX. 

An exampe is `AllVersionsInRangelMilestoningParametersEditor`. you notify if its invaid (without start/end date ) but still render the components with the guaranteeNonNullable.

packages/legend-query-builder/src/components/__tests__/QueryBuilderMilestoning.test.tsx
```diff
@@ -458,3 +462,127 @@ test(
     expect(getByText(executeDialog, 'Set Parameter Values'));
   },
 );
+
```
MauricioUyaguari, line 465: Can we add versionsInRange tests. Maybe creating it and resetting it.

MauricioUyaguari (changes requested):
looks good. Thanks

packages/legend-query-builder/src/components/explorer/QueryBuilderMilestoningEditor.tsx
gayathrir11: My bad i forgot to remove the invalid part from there. It's not revelant anymore. Once we render these editors all these milestoning dates are non-null

PR #2839, opened 2023-12-14
Persist filter condition value for is_in and is_not_in operators

gayathrir11:

- [ ] Test(s) added

https://github.com/finos/legend-studio/assets/87973126/a32a10ce-22db-40e1-88c6-3f814dbefbda

### DIFF
--- a/packages/legend-query-builder/src/stores/filter/QueryBuilderFilterState.ts
+++ b/packages/legend-query-builder/src/stores/filter/QueryBuilderFilterState.ts
@@ -42,6 +42,8 @@ import {
   type ValueSpecification,
   type VariableExpression,
   observe_ValueSpecification,
+  CollectionInstanceValue,
+  InstanceValue,
 } from '@finos/legend-graph';
 import { DEFAULT_LAMBDA_VARIABLE_NAME } from '../QueryBuilderConfig.js';
 import type { QueryBuilderProjectionColumnDragSource } from '../fetch-structure/tds/projection/QueryBuilderProjectionColumnState.js';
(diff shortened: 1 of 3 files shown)


packages/legend-query-builder/src/stores/filter/QueryBuilderFilterState.ts
```diff
@@ -195,7 +198,24 @@ export class FilterConditionState implements Hashable {
   changeOperator(val: QueryBuilderFilterOperator): void {
     this.setOperator(val);
     if (!this.operator.isCompatibleWithFilterConditionValue(this)) {
-      this.setValue(this.operator.getDefaultFilterConditionValue(this));
+      let defaultValue = this.operator.getDefaultFilterConditionValue(this);
```
MauricioUyaguari, line 201: can we add a test

packages/legend-query-builder/src/stores/filter/QueryBuilderFilterState.ts
```diff
@@ -195,7 +198,24 @@ export class FilterConditionState implements Hashable {
...
+      if (
+        defaultValue instanceof CollectionInstanceValue &&
```
MauricioUyaguari, line 203: i think we missing the case where the current value is collection and the default value is collection.
so if they switch from `is in` to `is not in`, i think we should maintain the list.

gayathrir11: that works already

MauricioUyaguari (approved):

PR #2892, opened 2024-01-25
go to snapshot version of a workspace if there is one when creating an Ad-hoc query

YannanGao-gs:
## Summary

Go to snapshot version of a workspace if there is one when creating an Ad-hoc query

- [ ] Test(s) added


packages/legend-application-studio/src/components/editor/editor-group/service-editor/ServiceExecutionQueryEditor.tsx
```diff
@@ -249,14 +252,33 @@ export const ServiceExecutionQueryEditor = observer(
           };
         }
       } else {
+        const currentWorkSpaceId =
```
MauricioUyaguari, line 255: is there any other places where we call legend query that we can leverage this or no?

If so maybe we should calculate whether a workspace contains snapshot in the initial fetching of versions,etc.

MauricioUyaguari (approved):

MauricioUyaguari (changes requested):
looks good. code check fails however, please fix.

MauricioUyaguari (approved):

PR #2919, opened 2024-02-01
update query version I -  provide query information to users and make query version changeable

YannanGao-gs:
## Summary

provide query information to users and make query version changeable

- [ ] Test(s) added

https://github.com/finos/legend-studio/assets/73408381/8c597605-dd07-4231-8a4c-4e90aa69fb98

https://github.com/finos/legend-studio/assets/73408381/40f11580-aad6-42a0-a918-1c0fb3606847

### DIFF
--- a/packages/legend-application-query/src/components/QueryEditor.tsx
+++ b/packages/legend-application-query/src/components/QueryEditor.tsx
@@ -44,6 +44,9 @@ import {
   clsx,
   ModalHeaderActions,
   TimesIcon,
+  Panel,
+  PanelFullContent,
+  CustomSelectorInput,
 } from '@finos/legend-art';
 import { observer } from 'mobx-react-lite';
 import { Fragment, useEffect, useMemo, useRef, useState } from 'react';
@@ -54,6 +57,7 @@ import {
   LEGEND_QUERY_QUERY_PARAM_TOKEN,
   LEGEND_QUERY_ROUTE_PATTERN_TOKEN,
   generateQuerySetupRoute,
+  generateExistingQueryEditorRoute,
 } from '../__lib__/LegendQueryNavigation.js';
 import {
   createViewProjectHandler,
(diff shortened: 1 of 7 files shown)


packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -360,6 +365,111 @@ const QueryEditorExistingQueryHeader = observer(
...
+              <div className="query-preview__field">
+                <div className="query-preview__field__label">Runtime</div>
+                <div className="query-preview__field__value">
+                  {query.runtime.value.name}
+                </div>
+              </div>
+              <div className="query-preview__field">
+                <div className="query-preview__field__label">Version</div>
+                <div className="query-preview__field__input">
+                  <InputWithInlineValidation
```
MauricioUyaguari, line 436: why is this not a dropdown? 
we should be able to get all the versions of the project.

MauricioUyaguari:
when user saves initial query, the default is `latest` right, so thats what they will see in the first video?

MauricioUyaguari:
i think we should think a bit about the user workflow. If a user updates the query version from a saved query, we should probably reload. after updating as they would be facing a different model. we should also probably add a warning if they change the version that this will happen.

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -360,6 +365,111 @@ const QueryEditorExistingQueryHeader = observer(
...
+      },
+    );
+
+    const closeModal = (): void => updateState.setShowQueryInfo(false);
+
+    return (
+      <Dialog
+        open={updateState.showQueryInfo}
+        onClose={closeModal}
+        classes={{ container: 'search-modal__container' }}
```
travisstebbins, line 399: Just for my own understanding, how do we decide the names/syntax used in our class names?

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -360,6 +365,111 @@ const QueryEditorExistingQueryHeader = observer(
...
+                    className="btn btn--dark query-preview__field__input__action"
+                    onClick={updateQueryVersionId}
+                  >
+                    save
+                  </button>
+                </div>
+              </div>
+              {query.owner && (
+                <div className="query-preview__field">
+                  <div className="query-preview__field__label">owner</div>
```
travisstebbins, line 454: Should this be capitalized like the other labels?
```suggestion
                  <div className="query-preview__field__label">Owner</div>
```

YannanGao-gs:
> when user saves initial query, the default is `latest` right, so thats what they will see in the first video?

I think it depends on how you create a query.
Currently, i just chose to `create a new query on a mapping`

Screenshot 2024-02-13 at 1 53 08 PM

YannanGao-gs:
> when user saves initial query, the default is `latest` right, so thats what they will see in the first video?

The first video is about displaying query's information e.g. project id, its current version id, owner..

packages/legend-application-query/src/components/QueryEditor.tsx
YannanGao-gs: I just reuse some shared modals in the codebase

YannanGao-gs:
> i think we should think a bit about the user workflow. If a user updates the query version from a saved query, we should probably reload. after updating as they would be facing a different model. we should also probably add a warning if they change the version that this will happen.


query now will be reloaded once users click the 'save' button to update version


To enable revert to a working version, I think I need to update the engine/depot to record the original version ID that this query is created at, which will be implemented in another PR

MauricioUyaguari (approved):

PR #2947, opened 2024-02-14
provide a way to update version if query is incompatible with current version

YannanGao-gs:
## Summary

provide a way to update version if query is incompatible with current version

- [ ] Test(s) added


packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -819,6 +819,114 @@ const QueryEditorHeaderContent = observer(
   },
 );
 
+const QueryEditorExistingQueryVersionRevertModal = observer(
```
travisstebbins, line 822: Just from a style/code organization perspective, could we move this modal component into its own file? I've noticed a lot of our files contain multiple components within them and can get really long. I think it would be cleaner to keep each file as small as possible and make sure each component (at least components that are relatively large) get their own file. Thoughts?

packages/legend-application-query/src/components/QueryEditor.tsx
```diff
@@ -819,6 +819,114 @@ const QueryEditorHeaderContent = observer(
...
+    ): Promise<void> => {
+      if (option?.value && option.value !== queryVersion) {
+        setQueryVersion(option.value);
+      }
+    };
+    const updateQueryVersionId = applicationStore.guardUnhandledError(
+      async (): Promise<void> => {
+        if (queryVersion) {
+          flowResult(
+            existingEditorStore.updateState.updateQueryVersionId(
```
travisstebbins, line 851: How come here we use `existingEditorStore.updateState` but on line 857 we just use `updateState`?

packages/legend-graph/src/graph-manager/AbstractPureGraphManager.ts
```diff
@@ -582,6 +587,10 @@ export abstract class AbstractPureGraphManager {
   abstract getQueryInfo(queryId: string): Promise<QueryInfo>;
   abstract createQuery(query: Query, graph: PureModel): Promise<Query>;
   abstract updateQuery(query: Query, graph: PureModel): Promise<Query>;
+  abstract updateQuerySelectedFields(
```
travisstebbins, line 590: Should we update this to be `patchQuery` to match the `legend-engine` change?

packages/legend-graph/src/graph-manager/action/query/Query.ts
```diff
@@ -71,13 +73,41 @@ export class LightQuery {
   isCurrentUserQuery = false;
 }
 
+// use PartialQuery to update an existing query in DB with given attributes that are not null/undefined
+export class PartialQuery {
```
travisstebbins, line 77: Instead of creating a new class `PartialQuery`, could we just use the built-in TypeScript `Partial` utility type? i.e., `Partial<Query>`

packages/legend-graph/src/graph-manager/protocol/pure/v1/V1_PureGraphManager.ts
```diff
@@ -2971,6 +2973,17 @@ export class V1_PureGraphManager extends AbstractPureGraphManager {
     );
   }
 
+  async updateQuerySelectedFields(
```
travisstebbins, line 2976: Should we also rename this to `patchQuery`?

packages/legend-graph/src/graph-manager/AbstractPureGraphManager.ts
YannanGao-gs: thanks missed this one

packages/legend-application-query/src/components/QueryEditor.tsx
YannanGao-gs: loll you are right thanks resolved

packages/legend-application-query/src/components/QueryEditor.tsx
YannanGao-gs: good point. 
Moved it to its own file

packages/legend-graph/src/graph-manager/action/query/Query.ts
YannanGao-gs: I did try to use PartialQuery, but I faced some issues like 
Screenshot 2024-03-08 at 6 32 36 PM
😅
Haven't figured out a way to fix this

travisstebbins: Does that error still happen if you do this? Do we need to do `Query | Partial<Query>`?
```
export const V1_transformQuery = (
  metamodel: Partial<Query>,
): Partial<V1_Query> => {
```

YannanGao-gs: I added some if conditions to make it work 🥹😅

MauricioUyaguari (approved):

PR #2958, opened 2024-02-20
support undo/redo action in query builder

YannanGao-gs:
## Summary

support undo/redo action in query builder

## How did you test this change?

- [ ] Test(s) added
- [ ] Manual testing (please provide screenshots/recordings)
- [ ] No testing (please provide an explanation)

legend query:

https://github.com/finos/legend-studio/assets/73408381/88cb09f3-38cc-481e-943d-2a764c7c5ea6

legend studio: 

https://github.com/finos/legend-studio/assets/73408381/12066010-4fe2-4992-8bbb-40db22db00b2

PR #2965, opened 2024-02-26
[Refactored] support undo/redo action in query builder

YannanGao-gs:
## Summary

support undo/redo action in query builder

improve https://github.com/finos/legend-studio/pull/2958 by encapsulating undo/redo button 

- [x] support undo/redo action in query builder
- [x] bind ctrl-z and ctrl-y
- [x] make undo/redo contextual

- [ ] Test(s) added

legend query:

https://github.com/finos/legend-studio/assets/73408381/b11e1d60-a673-4535-9310-61a3cd0240ec

legend studio: 

https://github.com/finos/legend-studio/assets/73408381/996829cb-3834-4bab-a91a-f7de3af31125

bind ctrl + z to undo , ctrl + y to redo

https://github.com/finos/legend-studio/assets/73408381/c2330929-54e9-46e1-a94c-74682bb9269a

make undo/redo contextual  -> ctrl + z won't close a new open modal

https://github.com/finos/legend-studio/assets/73408381/f3525a56-65e4-41a5-bdb6-80830e4db3c1

### DIFF
--- a/packages/legend-lego/src/application/RedoButton.tsx
+++ b/packages/legend-lego/src/application/RedoButton.tsx
@@ -18,15 +18,22 @@ import { RedoIcon } from '@finos/legend-art';
 import { useEffect } from 'react';
 
 export const RedoButton: React.FC<{
-  isRedoUnderContext: boolean;
+  parent: React.RefObject<HTMLDivElement>;
   canRedo: boolean;
   redo: () => void;
 }> = (props) => {
-  const { isRedoUnderContext, canRedo, redo } = props;
+  const { parent, canRedo, redo } = props;
 
   useEffect(() => {
     const onCtrlY = (event: KeyboardEvent): void => {
-      if (event.ctrlKey && event.key === 'y' && isRedoUnderContext) {
+      const target = event.target as Node | null;
+      if (
+        event.ctrlKey &&
+        event.key === 'y' &&
+        target !== null &&
+        parent.current !== null &&
+        (parent.current.contains(target) || target.contains(parent.current))
+      ) {
         event.preventDefault();
         redo();
       }
@@ -35,7 +42,7 @@ export const RedoButton: React.FC<{
     return () => {
       document.removeEventListener('keydown', onCtrlY);
     };
-  }, [isRedoUnderContext, redo]);
+  }, [parent, redo]);
 
   return (
     <div className="undo-redo">
--- a/packages/legend-lego/src/application/RedoButton.tsx
+++ b/packages/legend-lego/src/application/RedoButton.tsx
@@ -1,0 +1,54 @@
+/**
+ * Copyright (c) 2020-present, Goldman Sachs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { RedoIcon } from '@finos/legend-art';
+import { useEffect } from 'react';
+
+export const RedoButton: React.FC<{
+  isRedoUnderContext: boolean;
+  canRedo: boolean;
+  redo: () => void;
+}> = (props) => {
+  const { isRedoUnderContext, canRedo, redo } = props;
+
+  useEffect(() => {
+    const onCtrlY = (event: KeyboardEvent): void => {
+      if (event.ctrlKey && event.key === 'y' && isRedoUnderContext) {
+        event.preventDefault();
+        redo();
+      }
+    };
+    document.addEventListener('keydown', onCtrlY);
+    return () => {
+      document.removeEventListener('keydown', onCtrlY);
+    };
+  }, [isRedoUnderContext, redo]);
+
+  return (
+    <div className="undo-redo">
+      <button
+        className="undo-redo__button"
+        onClick={redo}
+        tabIndex={-1}
+        title={'Redo(ctrl + y)'}
+        disabled={!canRedo}
+      >
+        <RedoIcon />
+        <div className="undo-redo__button__label">Redo</div>
+      </button>
+    </div>
+  );
+};
--- a/packages/legend-lego/src/application/RedoButton.tsx
+++ b/packages/legend-lego/src/application/RedoButton.tsx
@@ -32,6 +32,10 @@ export const RedoButton: React.FC<{
         event.key === 'y' &&
         target !== null &&
         parent.current !== null &&
+        // This is for making sure undo/redo is contextual so that it won't modify the underlying query when a new modal is open
+        // The reason why not to do this via registerCommands() is that it's hard to compose contextual condtion without introducing an explicit
+        // dependency on parameters/constants/watermark/other modals. If we are going to introduce more modals,
+        // we have to update the condition correspondingly but developers might forget to update the condition.
         (parent.current.contains(target) || target.contains(parent.current))
       ) {
         event.preventDefault();
--- a/packages/legend-query-builder/src/components/QueryBuilder.tsx
+++ b/packages/legend-query-builder/src/components/QueryBuilder.tsx
@@ -383,10 +383,7 @@ export const QueryBuilder = observer(
     useEffect(() => {
       // this condition is for passing all exisitng tests because when we initialize a queryBuilderState for a test,
       // we use an empty RawLambda with an empty class and this useEffect is called earlier than initializeWithQuery()
-      if (
-        (queryBuilderState.isQuerySupported && queryBuilderState.class) ||
-        !queryBuilderState.isQuerySupported
-      ) {
+      if (queryBuilderState.isQuerySupported && queryBuilderState.class) {
         queryBuilderState.changeHistoryState.cacheNewQuery(
           queryBuilderState.buildQuery(),
         );
@@ -437,12 +434,18 @@ export const QueryBuilder = observer(
                 <div className="query-builder__header__actions__undo-redo">
                   <UndoButton
                     parent={queryBuilderRef}
-                    canUndo={queryBuilderState.changeHistoryState.canUndo}
+                    canUndo={
+                      queryBuilderState.changeHistoryState.canUndo &&
+                      queryBuilderState.isQuerySupported
+                    }
                     undo={undo}
                   />
                   <RedoButton
                     parent={queryBuilderRef}
-                    canRedo={queryBuilderState.changeHistoryState.canRedo}
+                    canRedo={
+                      queryBuilderState.changeHistoryState.canRedo &&
+                      queryBuilderState.isQuerySupported
+                    }
                     redo={redo}
                   />
                 </div>
--- a/packages/legend-query-builder/src/components/QueryBuilder.tsx
+++ b/packages/legend-query-builder/src/components/QueryBuilder.tsx
@@ -43,8 +43,6 @@ import {
   CalendarClockIcon,
   ChatIcon,
   PanelLoadingIndicator,
-  UndoIcon,
-  RedoIcon,
 } from '@finos/legend-art';
 import { QueryBuilderFilterPanel } from './filter/QueryBuilderFilterPanel.js';
 import { QueryBuilderExplorerPanel } from './explorer/QueryBuilderExplorerPanel.js';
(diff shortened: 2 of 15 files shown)


YannanGao-gs:
noted: I just found out that Legend Query is not behaving consistently with Legend Studio as Legend Query's document.focus is not reset to be query builder after you close one modal

packages/legend-lego/src/application/RedoButton.tsx
```diff
@@ -0,0 +1,61 @@
...
+  redo: () => void;
+}> = (props) => {
+  const { parent, canRedo, redo } = props;
+
+  useEffect(() => {
+    const onCtrlY = (event: KeyboardEvent): void => {
+      const target = event.target as Node | null;
+      if (
+        event.ctrlKey &&
+        event.key === 'y' &&
```
gayathrir11, line 32: Have you tried registering the commands in queryBuilderState instead of binding them like this?

packages/legend-query-builder/src/components/QueryBuilder.tsx
```diff
@@ -367,10 +370,35 @@ export const QueryBuilder = observer(
...
+        queryBuilderState.changeHistoryState.redo();
+      }
+    };
+
+    useEffect(() => {
+      // this condition is for passing all exisitng tests because when we initialize a queryBuilderState for a test,
+      // we use an empty RawLambda with an empty class and this useEffect is called earlier than initializeWithQuery()
+      if (
+        (queryBuilderState.isQuerySupported && queryBuilderState.class) ||
+        !queryBuilderState.isQuerySupported
```
gayathrir11, line 389: Is it this condition you are adding to prevent test failures `!queryBuilderState.isQuerySupported` ? What error do the tests throw?

packages/legend-query-builder/src/stores/QueryBuilderChangeHistoryState.ts
```diff
@@ -0,0 +1,111 @@
...
+    this.initState.inProgress();
+    this.currentQuery = initialQuery;
+    this.querySnapshotBuffer.push(initialQuery);
+    this.pointer = this.pointer + 1;
+    this.initState.complete();
+  }
+
+  cacheNewQuery(query: RawLambda): void {
+    if (query.hashCode !== this.currentQuery?.hashCode) {
+      if (
```
gayathrir11, line 94: Do we want to handle the case where query could be unsupported? I think we shouldn't allow these undo and redo operations if query is unsupported. It's just the lambda we would have not any UI state so might just make sense to disable these operations

packages/legend-query-builder/src/stores/__tests__/QueryBuilderChangeDetection.test.ts
```diff
@@ -89,5 +89,11 @@ test(integrationTest('Test change detection'), () => {
     expect(queryBuilderState.hashCode).toBe(
       queryBuilderState.changeDetectionState.hashCodeSnapshot,
     );
+
+    // Test Redo/Undo action in Query Builder
+    fireEvent.click(renderResult.getByText('Undo'));
```
gayathrir11, line 94: can we please add a new test in a new file instead of modifying existing one?

gayathrir11 (dismissed):

packages/legend-lego/src/application/RedoButton.tsx
```diff
@@ -0,0 +1,61 @@
...
+  }, [parent, redo]);
+
+  return (
+    <div className="undo-redo">
+      <button
+        className="undo-redo__button"
+        onClick={redo}
+        tabIndex={-1}
+        title={'Redo(ctrl + y)'}
+        disabled={!canRedo}
```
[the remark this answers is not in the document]
YannanGao-gs: I think maybe we could keep it for unsupported queries since we keep query snapshots no matter it's unsupported or supported in form mode

It 's currently working upon unsupported queries.

packages/legend-query-builder/src/stores/__tests__/QueryBuilderChangeDetection.test.ts
YannanGao-gs: sure! I will create a new one. I thought we could  avoid some duplicated code If we could use the existing one for query setup

packages/legend-query-builder/src/stores/QueryBuilderChangeHistoryState.ts
YannanGao-gs: I think it's better to support undo/redo on unsupported queries as it might take a while to open it in the lambda editor to modify the query back to what it was before. With undo/redo, you could just hit keyboard shortcuts or click the undo button. 👀👀
However, I'm okay with making it not work with unsupported queries if others also agree on it

packages/legend-lego/src/application/RedoButton.tsx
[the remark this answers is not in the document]
YannanGao-gs: I did! However, I found out that registerCommand() will only be initially called once. But If we want to make undo/redo contextual, we should bind undo/redo correspondingly when the condition is true or false. 
😅😅👀👀👀

packages/legend-lego/src/application/RedoButton.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  useEffect(() => {
+    const onCtrlY = (event: KeyboardEvent): void => {
+      const target = event.target as Node | null;
+      if (
+        event.ctrlKey &&
+        event.key === 'y' &&
+        target !== null &&
+        parent.current !== null &&
+        (parent.current.contains(target) || target.contains(parent.current))
```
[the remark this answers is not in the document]
YannanGao-gs: I did think about checking/tracking if parameters/constants/watermark/other modals are not opened, but eventually i don't think it's a good idea as I really don't want to make binding keyboard shortcuts explicitly depend on **parameters/constants/watermark/other modals** if we could just check if event.target is the root query builder.

If we are going to introduce more modals, we have to update the condition correspondingly but developers might forget to update the condition.

packages/legend-query-builder/src/components/QueryBuilder.tsx
YannanGao-gs: all existing tests will fail to render result for query builder

```
  const renderResult = render(
    <ApplicationStoreProvider store={MOCK__applicationStore}>
      <TEST__BrowserEnvironmentProvider>
        <ApplicationFrameworkProvider>
          <QueryBuilder queryBuilderState={queryBuilderState} />
        </ApplicationFrameworkProvider>
      </TEST__BrowserEnvironmentProvider>
    </ApplicationStoreProvider>,
  );
```

 as this useEffect is called earlier and it will trigger queryBuilderState.buildQuery()
and will eventually fail at ..../src/stores/QueryBuilderValueSpecificationBuilder.ts line 158
```
  const _class = guaranteeNonNullable(
    queryBuilderState.class,
    'Class is required to build query',
  );
```

🫣🫣🫣🫣

packages/legend-lego/src/application/RedoButton.tsx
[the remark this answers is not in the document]
YannanGao-gs: Yes, I did! But last time, when I tried to test this function to see if the key could be registered conditionally by turning a dummy boolean to true and false, the undo/redo behaved very weirdly. 😅 Based on John's feedback on encapsulating the undo/redo (make binding part of it), I didn't try more.
I'm open to giving it another try. Waiting for the final decision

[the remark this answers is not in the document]
YannanGao-gs: I did. 👀👀😅 The current issue is it's a little hard to pass the contextual condition to queryBuilderState if we don't want to check if parameters/constants/watermark/other modals are not opened. 🥹

packages/legend-query-builder/src/stores/QueryBuilderChangeHistoryState.ts
gayathrir11: @MauricioUyaguari  what do you think on this?

MauricioUyaguari: i dont think it makes sense to support this in unsupported mode. the user doesn't see the query lambda and is not editable so if they accidently hit these shortcuts, they wont know what they are switching to/from

YannanGao-gs: I disabled undo/redo in unsupported mode

packages/legend-query-builder/src/stores/__tests__/QueryBuilderChangeHistory.test.ts
```diff
@@ -0,0 +1,95 @@
...
+    guaranteeNonNullable(
+      guaranteeType(
+        queryBuilderState.fetchStructureState.implementation,
+        QueryBuilderTDSState,
+      ).projectionColumns[0],
+    ).setColumnName('Legal Name');
+    await waitFor(() => getByText(projectionCols, 'Legal Name'));
+
+    // Test Redo/Undo action in Query Builder
+    fireEvent.click(renderResult.getByText('Undo'));
```
MauricioUyaguari, line 90: This test does not really test undo/redo. 
After first undo, you should ensure you no longer see Legal Name in the projection panel. 

Additionally can you add more complicated operations. 
Maybe DND in filter. 

If the expectation is that undo/redo do not work in modals, can you test this here as well ?

packages/legend-query-builder/src/components/QueryBuilder.tsx
```diff
@@ -368,10 +371,30 @@ export const QueryBuilder = observer(
...
+    };
+
+    const redo = (): void => {
+      queryBuilderState.changeHistoryState.redo();
+    };
+
+    useEffect(() => {
+      // this condition is for passing all exisitng tests because when we initialize a queryBuilderState for a test,
+      // we use an empty RawLambda with an empty class and this useEffect is called earlier than initializeWithQuery()
+      if (queryBuilderState.isQuerySupported && queryBuilderState.class) {
```
MauricioUyaguari, line 386: `cacheNewQuery` runs `hashCode` each time. 

Have you tested the performance against larger queries (lots of columns) to ensure the UI isnt affected by this computation between each action?

packages/legend-query-builder/src/stores/QueryBuilderChangeHistoryState.ts
```diff
@@ -0,0 +1,115 @@
...
+      this.initState.inProgress();
+      this.currentQuery = initialQuery;
+      this.querySnapshotBuffer.push(initialQuery);
+      this.pointer = this.pointer + 1;
+      this.initState.complete();
+    }
+  }
+
+  cacheNewQuery(query: RawLambda): void {
+    if (this.queryBuilderState.isQuerySupported) {
```
MauricioUyaguari, line 95: could we ensure the methods here are safe to call, if this is run on every action, we should probably wrap it in a try catch and log if we fail to cached the query.

MauricioUyaguari (changes requested):

packages/legend-query-builder/src/stores/__tests__/QueryBuilderChangeHistory.test.ts
YannanGao-gs: `await waitFor(() => getByText(projectionCols, 'Name'));` is checking the projection panel contains 'Name' as getByText looks for an exact match of the provided string within the text content of the elements.
 However, I could add a check to expect Legal Name to be null

packages/legend-lego/src/application/RedoButton.tsx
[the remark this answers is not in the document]
YannanGao-gs: We are still going to bind the command keyboards for undo/redo. The discussion above is about how to bind it

packages/legend-query-builder/src/stores/__tests__/QueryBuilderChangeHistory.test.ts
MauricioUyaguari: yes but the way the current test is written, undo/redo could both be broken and the test would still pass as you are loading it with both 'Name' and 'LegalName` and then doing checks on both those columns without checking if the one who shouldnt show is actually gone.

packages/legend-lego/src/application/RedoButton.tsx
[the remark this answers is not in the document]
MauricioUyaguari: yeah exactly we should a comment that we are not binding them via the registry command because of the detection of whether of modall is open

packages/legend-query-builder/src/stores/QueryBuilderChangeHistoryState.ts
```diff
@@ -0,0 +1,126 @@
...
+            );
+          }
+          this.querySnapshotBuffer.push(query);
+          this.pointer = this.querySnapshotBuffer.length - 1;
+          this.setCurrentQuery(query);
+        }
+      }
+    } catch (error) {
+      assertErrorThrown(error);
+      this.queryBuilderState.applicationStore.notificationService.notifyError(
```
MauricioUyaguari, line 121: we should log the error not notify i think. create a new legend query. CHANGE_HISTORY_ERROR or something like that

MauricioUyaguari:
Awesome!

MauricioUyaguari (approved):

PR #2969, opened 2024-02-27
Fetch structure panel is cluttered

travisstebbins:
## Summary

Improve UX by cleaning up the fetch structure panel:
1. Ensure that aggregate operator and remove buttons don't get hidden when the fetch structure panel width is reduced
2. For non-derived columns, show the column name in the property expression badge instead of showing the property name in the property expression badge with a separate input element for the column name.
3. For non-derived columns, clicking on the property expression badge will switch the element to an input to allow editing the column name.
4. Add the property name to the property info tooltip, so that if the column name is edited, the user can still see the original property name.
5. Add a button next to the "path" row in the property info tooltip to highlight the property in the explorer panel.

- [x] Test(s) added

Clicking on property expression badge allows editing a column name:
EditColumnName

Edited column names and derived column names work correctly when query is executed:
QueryWithNamedColumns

Aggregation query works correctly when query is executed.
AggregationQuery

Aggregate operator and remove buttons don't get hidden when the fetch structure panel is reduced:
ResizingFetchStructurePanel

Property info tooltip has property name and button next to "path" row that highlights property in explorer panel:
PathNodeHighlight

Property info tooltips in the fetch structure panel and filter panel have the "show in tree" button. Property info tooltips in the explorer panel and search panel don't have the "show in tree" button since it is not needed, as those tooltips already exist within the explorer tree.
TooltipsWithAndWithoutButton

### DIFF
--- a/packages/legend-query-builder/src/stores/explorer/QueryBuilderExplorerState.ts
+++ b/packages/legend-query-builder/src/stores/explorer/QueryBuilderExplorerState.ts
@@ -745,7 +745,11 @@ export class QueryBuilderExplorerState {
       }
       this.refreshTree();
       nodeToHighlight?.setIsHighlighting(true);
-      nodeToHighlight?.ref?.current?.scrollIntoView();
+      // scrollIntoView must be called in a setTimeout because it must happen after
+      // the tree nodes are recursively opened and the tree is refreshed.
+      setTimeout(() => {
+        nodeToHighlight?.ref?.current?.scrollIntoView();
+      }, 0);
     }
   }
 
--- a/packages/legend-query-builder/src/stores/explorer/QueryBuilderExplorerState.ts
+++ b/packages/legend-query-builder/src/stores/explorer/QueryBuilderExplorerState.ts
@@ -737,11 +737,10 @@ export class QueryBuilderExplorerState {
         if (!nodeToOpen.isOpen) {
           nodeToOpen.isOpen = true;
         }
-        if (nodeToOpen instanceof QueryBuilderExplorerTreePropertyNodeData) {
-          nodeToOpen = this.treeData?.nodes?.get(nodeToOpen.parentId) ?? null;
-        } else {
-          nodeToOpen = null;
-        }
+        nodeToOpen =
+          nodeToOpen instanceof QueryBuilderExplorerTreePropertyNodeData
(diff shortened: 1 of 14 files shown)


MauricioUyaguari:
@travisstebbins  looks amazing. Thanks

MauricioUyaguari:
Looks good code wise. Could we add one video with an aggregation column here to see the behavior. Thanks

packages/legend-query-builder/src/stores/explorer/QueryBuilderExplorerState.ts
```diff
@@ -114,7 +117,9 @@ export abstract class QueryBuilderExplorerTreeNodeData implements TreeNodeData {
     mappingData: QueryBuilderExplorerTreeNodeMappingData,
   ) {
     makeObservable(this, {
```
travisstebbins, line 119: @MauricioUyaguari or @YannanGao-gs , do you know why we don't include `isOpen`/`setIsOpoen` as properties in this `makeObservable` call? I tested out adding those and it made it so I wouldn't have to call `this.refreshTree()` on line 745 in the `highlightTreeNode` function (https://github.com/finos/legend-studio/pull/2969/files#diff-f03db7e81110555697c5eeec710ace3f7123cc5f1eed7004008bff10c70b7843R745). However, I didn't want to make this change just in case we intentionally left out `isOpen`/`setIsOpen` here.

travisstebbins:
> @travisstebbins looks amazing. Thanks

Thanks! I also left one comment myself on a line in my changes, just for my own curiosity, if you could review.

travisstebbins:
> Looks good code wise. Could we add one video with an aggregation column here to see the behavior. Thanks

Edited the description to add a screen recording of a query with an aggregation column.

MauricioUyaguari (approved):

packages/legend-query-builder/src/stores/explorer/QueryBuilderExplorerState.ts
MauricioUyaguari: we may have overlooked this. however it may also be the case, that when we go through the processes of finding a node, we want to open the nodes all at once instead of opening them while recursively looking for them. we can look to enable this but we should look at places where we open an unopened node deep into the tree and make sure that works as expected.

PR #2986, opened 2024-03-05
Refactor query options modal to have "Apply" and "Cancel" buttons

travisstebbins:
## Summary

Currently, updating values in the "Query Options" modal automatically updates the values in the store. The modal only has a close button to dismiss.

Instead, the Query Options modal should have an "Apply" button and a "Cancel" button. The user can open the modal to see the current state of query options (query builder result modifiers) and make changes as he/she likes. If the user clicks the Apply button, then the changes will be applied to the state; however, if the user clicks the Cancel button, then the user's changes should be discarded and the state should be left unchanged.

This PR also makes the below changes to the fields within the query options modal:

- Only allow numbers into the limit result input (currently, `+`,`.`, and `-` characters can be entered into the input)
- Only allow numbers into the slice inputs (currently, `+`,`.`, and `-` characters can be entered into the input)
- Make slice input always visible instead of enabled with a button.
  - If the user doesn't enter values into the slice inputs, then the slice state is set to undefined.
- Add validation to the slice input (both boxes must contain values or be empty, and the slice start value cannot be >= the slice end value).

- [x] Test(s) added

Editing column sort property:
ColumnSort

Editing eliminate duplicate rows property:
Distinct

Editing limit results property:
LimitResults

Editing slice property:
Slice

### DIFF
--- a/.changeset/brown-clocks-listen.md
+++ b/.changeset/brown-clocks-listen.md
@@ -1,6 +1,5 @@
 ---
 '@finos/legend-query-builder': patch
-'@finos/legend-art': patch
 ---
 
 Refactor query options modal to have Apply and Cancel buttons
--- a/.changeset/brown-clocks-listen.md
+++ b/.changeset/brown-clocks-listen.md
@@ -1,0 +1,6 @@
+---
+'@finos/legend-query-builder': patch
+'@finos/legend-art': patch
+---
+
+Refactor query options modal to have Apply and Cancel buttons
--- a/.changeset/chilled-parents-wait.md
+++ b/.changeset/chilled-parents-wait.md
@@ -1,0 +1,5 @@
+---
+'@finos/legend-art': patch
+---
+
+Add secondary property to ModalFooterButton
--- a/packages/legend-art/src/dialog/Modal.tsx
+++ b/packages/legend-art/src/dialog/Modal.tsx
@@ -105,7 +105,7 @@ export const ModalFooterButton: React.FC<{
   } = props;
   const isDarkMode = darkMode ?? true;
   const isFormatText = formatText ?? true;
-  const isSecondary = type === 'secondary' ?? false;
+  const isSecondary = type === 'secondary';
 
   return (
     <button
--- a/packages/legend-art/src/dialog/Modal.tsx
+++ b/packages/legend-art/src/dialog/Modal.tsx
@@ -89,6 +89,7 @@ export const ModalFooterButton: React.FC<{
   className?: string;
   onClick: (event: React.MouseEvent<HTMLButtonElement>) => void;
   formatText?: boolean | undefined;
+  type?: 'primary' | 'secondary';
 }> = (props) => {
   const {
     onClick,
@@ -100,9 +101,11 @@ export const ModalFooterButton: React.FC<{
     className,
     darkMode,
     formatText,
+    type,
   } = props;
   const isDarkMode = darkMode ?? true;
   const isFormatText = formatText ?? true;
+  const isSecondary = type === 'secondary' ?? false;
 
   return (
     <button
@@ -112,6 +115,9 @@ export const ModalFooterButton: React.FC<{
           'btn--dark': isDarkMode,
         },
         { 'btn--light': !isDarkMode },
+        {
+          'modal__footer__btn--secondary': isSecondary,
+        },
         className,
       )}
       title={title}
--- a/packages/legend-query-builder/src/components/__tests__/QueryBuilderResultModifierPanel.test.tsx
+++ b/packages/legend-query-builder/src/components/__tests__/QueryBuilderResultModifierPanel.test.tsx
@@ -35,7 +35,10 @@ import { guaranteeNonNullable, guaranteeType } from '@finos/legend-shared';
 import { integrationTest } from '@finos/legend-shared/test';
 import { create_RawLambda, stub_RawLambda } from '@finos/legend-graph';
 import { QUERY_BUILDER_TEST_ID } from '../../__lib__/QueryBuilderTesting.js';
-import { TEST__setUpQueryBuilder } from '../__test-utils__/QueryBuilderComponentTestUtils.js';
+import {
+  TEST__setUpQueryBuilder,
+  selectFromCustomSelectorInput,
+} from '../__test-utils__/QueryBuilderComponentTestUtils.js';
 import { COLUMN_SORT_TYPE } from '../../graph/QueryBuilderMetaModelConst.js';
 import type { QueryBuilderState } from '../../stores/QueryBuilderState.js';
 import { QueryBuilderTDSState } from '../../stores/fetch-structure/tds/QueryBuilderTDSState.js';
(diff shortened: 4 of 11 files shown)


packages/legend-query-builder/src/components/__tests__/QueryBuilderResultModifierPanel.test.tsx
```diff
@@ -0,0 +1,467 @@
...
+      expect(queryByText(resultModifierPanel, 'Edited First Name')).toBeNull();
+      expect(await findByText(resultModifierPanel, 'asc')).not.toBeNull();
+
+      const applyButton = await findByRole(resultModifierPanel, 'button', {
+        name: 'Apply',
+      });
+      fireEvent.click(applyButton);
+
+      // Check state
+      expect(
```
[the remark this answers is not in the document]
travisstebbins: Yes good idea! I'll make that change.

packages/legend-query-builder/src/components/__tests__/QueryBuilderResultModifierPanel.test.tsx
```diff
@@ -0,0 +1,467 @@
...
+      expect(await findByText(resultModifierPanel, 'asc')).not.toBeNull();
+
+      // Don't apply the changes
+      const cancelButton = await findByRole(resultModifierPanel, 'button', {
+        name: 'Cancel',
+      });
+      fireEvent.click(cancelButton);
+
+      // Check new state
+      expect(
```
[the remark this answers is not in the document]
travisstebbins: Will do!

packages/legend-art/src/dialog/Modal.tsx
```diff
@@ -89,6 +89,7 @@ export const ModalFooterButton: React.FC<{
   className?: string;
   onClick: (event: React.MouseEvent<HTMLButtonElement>) => void;
   formatText?: boolean | undefined;
+  type?: 'primary' | 'secondary';
```
YannanGao-gs, line 92: maybe could we create a enum for this ?

packages/legend-query-builder/src/components/__tests__/QueryBuilderResultModifierPanel.test.tsx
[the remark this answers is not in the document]
MauricioUyaguari: i see the test below updated but not this one. maybe thats missed?

packages/legend-art/src/dialog/Modal.tsx
travisstebbins: I was doing some research and it seems like there's arguments for enums and for string unions (https://www.typescriptcourse.com/string-literal-unions-over-enums). In this case, I think a string union would be better, because there's only 2 options, so the string union feels simpler and cleaner to me. Also, VS Code will still provide you with autocomplete options and highlight invalid values when you use string unions.

packages/legend-query-builder/src/components/__tests__/QueryBuilderResultModifierPanel.test.tsx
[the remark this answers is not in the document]
travisstebbins: I think this one was also updated. I think the lines shown in this comment are just outdated for some reason, but if you look at the file changes now, all the tests should be updated.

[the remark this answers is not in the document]
MauricioUyaguari: thanks see it now

.changeset/brown-clocks-listen.md
```diff
@@ -0,0 +1,6 @@
+---
+'@finos/legend-query-builder': patch
+'@finos/legend-art': patch
+---
+
+Refactor query options modal to have Apply and Cancel buttons
```
YannanGao-gs, line 6: could we separate changesets?

packages/legend-query-builder/src/stores/fetch-structure/tds/QueryResultSetModifierState.ts
```diff
@@ -96,12 +91,12 @@ export class QueryResultSetModifierState implements Hashable {
...
+    this.distinct = val;
   }
 
-  deleteSortColumn(val: SortColumnState): void {
-    deleteEntry(this.sortColumns, val);
+  setSortColumns(val: SortColumnState[]): void {
+    this.sortColumns = val;
   }
 
   addSortColumn(val: SortColumnState): void {
```
YannanGao-gs, line 102: could this be deleted?

packages/legend-query-builder/src/components/fetch-structure/QueryBuilderResultModifierPanel.tsx
```diff
@@ -58,20 +70,26 @@ const ColumnSortEditor = observer(
...
         sortState.setColumnState(val.value);
       }
     };
-    const sortType = sortState.sortType;
 
-    const deleteColumnSort = (): void =>
-      tdsState.resultSetModifierState.deleteSortColumn(sortState);
+    const deleteColumnSort = (): void => {
+      const newSortColumns = [...sortColumns];
+      deleteEntry(newSortColumns, sortState);
```
YannanGao-gs, line 85: may I know the reason why we want to do it in this way instead of the original approach?

.changeset/brown-clocks-listen.md
travisstebbins: Done

packages/legend-query-builder/src/components/fetch-structure/QueryBuilderResultModifierPanel.tsx
travisstebbins: The original approach would apply your changes to the state as soon as you made a change in the modal, so there was no good way to cancel your changes. This way, all the changes are stored locally within the modal first and then only applied to the state if you click Apply.

Since the modal has a local state storing the sortColumns array, we can't just mutate that array. We need to replace the whole array in the modal's state so that the component correctly re-renders when we add/remove sort columns.

packages/legend-query-builder/src/stores/fetch-structure/tds/QueryResultSetModifierState.ts
travisstebbins: It looks like it's still being used here: https://github.com/finos/legend-studio/blob/master/packages/legend-query-builder/src/stores/fetch-structure/tds/projection/QueryBuilderProjectionStateBuilder.ts#L500

packages/legend-query-builder/src/components/fetch-structure/QueryBuilderResultModifierPanel.tsx
```diff
@@ -58,20 +70,26 @@ const ColumnSortEditor = observer(
       label: sortState.columnState.columnName,
       value: sortState,
     };
+    const sortType = sortState.sortType;
```
YannanGao-gs, line 73: just personal preference, since we onlu use it once, maybe we could remove this variable

.changeset/chilled-parents-wait.md
```diff
@@ -0,0 +1,5 @@
+---
+'@finos/legend-art': patch
+---
+
```
YannanGao-gs, line 4: i think you could leave the description empty 🥹😅😅👀

MauricioUyaguari (approved):

PR #3005, opened 2024-03-19
uplift query search

MauricioUyaguari:
## Summary
- [x] depends on https://github.com/finos/legend-engine/pull/2706

- [ ] Test(s) added

### DIFF
--- a/packages/legend-application-query/src/stores/QueryEditorStore.ts
+++ b/packages/legend-application-query/src/stores/QueryEditorStore.ts
@@ -506,12 +506,10 @@ export abstract class QueryEditorStore {
     const isValidSearchString =
       searchText.length >= DEFAULT_TYPEAHEAD_SEARCH_MINIMUM_SEARCH_LENGTH;
     try {
-      const searchSpecification = new QuerySearchSpecification();
+      const searchSpecification = QuerySearchSpecification.createDefault(
+        isValidSearchString ? searchText : undefined,
+      );
       searchSpecification.showCurrentUserQueriesOnly = true;
-      searchSpecification.exactMatchName = true;
-      searchSpecification.searchTerm = isValidSearchString
-        ? searchText
-        : undefined;
       searchSpecification.limit = 1;
       const queries = (yield this.graphManagerState.graphManager.searchQueries(
         searchSpecification,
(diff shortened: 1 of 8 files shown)


packages/legend-query-builder/src/stores/QueryLoaderState.ts
```diff
@@ -212,10 +212,9 @@ export class QueryLoaderState {
     this.showingDefaultQueries = false;
     this.searchQueriesState.inProgress();
     try {
-      let searchSpecification = new QuerySearchSpecification();
-      searchSpecification.searchTerm = searchText;
+      let searchSpecification =
+        QuerySearchSpecification.createDefault(searchText);
```
YannanGao-gs, line 216: before searchSpecification.exactMatchName = true;

now with QuerySearchSpecification.createDefault(searchText);
    term.exactMatchName = false;
is this expected?

packages/legend-application-query/src/stores/QueryEditorStore.ts
```diff
@@ -506,12 +506,10 @@ export abstract class QueryEditorStore {
     const isValidSearchString =
       searchText.length >= DEFAULT_TYPEAHEAD_SEARCH_MINIMUM_SEARCH_LENGTH;
     try {
-      const searchSpecification = new QuerySearchSpecification();
+      const searchSpecification = QuerySearchSpecification.createDefault(
+        isValidSearchString ? searchText : undefined,
```
YannanGao-gs, line 510: before searchSpecification.exactMatchName = true;

now with QuerySearchSpecification.createDefault(searchText);
term.exactMatchName = false;
is this expected?

packages/legend-query-builder/src/stores/QueryLoaderState.ts
MauricioUyaguari: yeah it is expected. 
we werent actuallyl sending `exactMatchName = true` to the server because the serialization for the V1 model was missing the property and hence it was always defaulting to false.

packages/legend-application-query/src/stores/QueryEditorStore.ts
MauricioUyaguari: same comment as above

YannanGao-gs (approved):

PR #3011, opened 2024-03-21
Support template query in dataspace IV - support share dataspace template query in Legend Query

YannanGao-gs:
## Summary

support share dataspace template query in Legend Query

pattern:
/dataspace/:${projection_gav}/:${dataspace_path}/template/:${template_query_title}

- [ ] Test(s) added

https://github.com/finos/legend-studio/assets/73408381/e4b43f05-7849-4afe-bad1-7af55f40574f

### DIFF
--- a/packages/legend-extension-dsl-data-space/src/__lib__/query/DSL_DataSpace_LegendQueryNavigation.ts
+++ b/packages/legend-extension-dsl-data-space/src/__lib__/query/DSL_DataSpace_LegendQueryNavigation.ts
@@ -29,10 +29,17 @@ export enum DATA_SPACE_QUERY_CREATOR_ROUTE_PATTERN_TOKEN {
   DATA_SPACE_PATH = 'dataSpacePath',
   EXECUTION_CONTEXT = 'executionContext',
   RUNTIME_PATH = 'runtimePath',
+  TEMPLATE = 'template',
 }
 
 export enum DATA_SPACE_QUERY_CREATOR_QUERY_PARAM_TOKEN {
   CLASS_PATH = 'class',
+}
+
+export enum DATA_SPACE_TEMPLATE_QUERY_CREATOR_ROUTE_PATTERN_TOKEN {
+  GAV = 'gav',
+  DATA_SPACE_PATH = 'dataSpacePath',
+  TEMPLATE_QUERY_TITLE = 'templateQueryTitle',
 }
 
 export type DataSpaceQueryCreatorPathParams = {
@@ -42,6 +49,12 @@ export type DataSpaceQueryCreatorPathParams = {
   [DATA_SPACE_QUERY_CREATOR_ROUTE_PATTERN_TOKEN.RUNTIME_PATH]?: string;
 };
 
+export type DataSpaceTemplateQueryCreatorPathParams = {
+  [DATA_SPACE_TEMPLATE_QUERY_CREATOR_ROUTE_PATTERN_TOKEN.GAV]: string;
+  [DATA_SPACE_TEMPLATE_QUERY_CREATOR_ROUTE_PATTERN_TOKEN.DATA_SPACE_PATH]: string;
+  [DATA_SPACE_TEMPLATE_QUERY_CREATOR_ROUTE_PATTERN_TOKEN.TEMPLATE_QUERY_TITLE]: string;
+};
+
 export type DataSpaceQueryEditorQueryParams = {
   [DATA_SPACE_QUERY_CREATOR_QUERY_PARAM_TOKEN.CLASS_PATH]?: string | undefined;
 };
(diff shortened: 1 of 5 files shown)


packages/legend-extension-dsl-data-space/src/stores/query/DataSpaceTemplateQueryViewerStore.ts
```diff
@@ -0,0 +1,215 @@
...
+          }]`,
+      decorator: (query: Query): void => {
+        query.id = uuid();
+        query.groupId = this.groupId;
+        query.artifactId = this.artifactId;
+        query.versionId = this.versionId;
+        query.taggedValues = [
+          createQueryDataSpaceTaggedValue(this.dataSpacePath),
+          createQueryClassTaggedValue(
+            guaranteeNonNullable(this.queryBuilderState?.class?.path),
```
MauricioUyaguari, line 209: lets add a null check here for guaranteeNonNullable. 
and add the class tagged value only if the class is defined (supported query). 
Could we check this in other places where we do guaranteNonNullable on the class path when decorating the query

packages/legend-extension-dsl-data-space/src/stores/query/DataSpaceTemplateQueryViewerStore.ts
```diff
@@ -0,0 +1,215 @@
...
+import type { QueryBuilderState } from '@finos/legend-query-builder';
+import type { ProjectGAVCoordinates } from '@finos/legend-storage';
+import { DSL_DataSpace_getGraphManagerExtension } from '../../graph-manager/protocol/pure/DSL_DataSpace_PureGraphManagerExtension.js';
+import { retrieveAnalyticsResultCache } from '../../graph-manager/action/analytics/DataSpaceAnalysisHelper.js';
+import {
+  createQueryDataSpaceTaggedValue,
+  createQueryClassTaggedValue,
+} from './DataSpaceQueryCreatorStore.js';
+
+export class DataSpaceTemplateQueryViewerStore extends QueryEditorStore {
```
MauricioUyaguari, line 56: should this be a creator store. does it make sense to abstract this and `DataSpaceQueryCreatorStore` or no?

packages/legend-extension-dsl-data-space/src/__lib__/query/DSL_DataSpace_LegendQueryNavigation.ts
```diff
@@ -35,20 +35,33 @@ export enum DATA_SPACE_QUERY_CREATOR_QUERY_PARAM_TOKEN {
...
+};
+
 export type DataSpaceQueryEditorQueryParams = {
   [DATA_SPACE_QUERY_CREATOR_QUERY_PARAM_TOKEN.CLASS_PATH]?: string | undefined;
 };
 
 export const DATA_SPACE_QUERY_ROUTE_PATTERN = Object.freeze({
   SETUP: `/dataspace`,
   CREATE: `/dataspace/:${DATA_SPACE_QUERY_CREATOR_ROUTE_PATTERN_TOKEN.GAV}/:${DATA_SPACE_QUERY_CREATOR_ROUTE_PATTERN_TOKEN.DATA_SPACE_PATH}/:${DATA_SPACE_QUERY_CREATOR_ROUTE_PATTERN_TOKEN.EXECUTION_CONTEXT}/:${DATA_SPACE_QUERY_CREATOR_ROUTE_PATTERN_TOKEN.RUNTIME_PATH}?`,
+  TEMPLATE_QUERY: `/template/dataspace/:${DATA_SPACE_TEMPLATE_QUERY_VIEWER_ROUTE_PATTERN_TOKEN.GAV}/:${DATA_SPACE_TEMPLATE_QUERY_VIEWER_ROUTE_PATTERN_TOKEN.DATA_SPACE_PATH}/:${DATA_SPACE_TEMPLATE_QUERY_VIEWER_ROUTE_PATTERN_TOKEN.TEMPLATE_QUERY_TITLE}`,
```
MauricioUyaguari, line 64: not sure this URL makes sense. wouldnt
`dataspace/gavDataspaceCreatore/datspacePath/template/{templateTitle}`
make more sense.

MauricioUyaguari (changes requested):

MauricioUyaguari:
some comments, mostly i think the url should be like the other creator dataspace one except with /template/templateTitle
 instead of the executionKey

packages/legend-extension-dsl-data-space/src/stores/query/DataSpaceTemplateQueryViewerStore.ts
YannanGao-gs: For template query, it's not `create` a template query but more about opening and editing it, therefore I named it a `viewer`. However, I renamed it to `DataSpaceTemplateQueryCreatorStore`.


I don't think we need to abstract it and `DataSpaceQueryCreatorStore ` again since QueryEditorStore is already the base class, and all methods in `DataSpaceTemplateQueryViewerStore ` and `DataSpaceQueryCreatorStore` are inherited from `QueryEditorStore`.

Besides, there are very few overlaps between `DataSpaceTemplateQueryViewerStore ` and `DataSpaceQueryCreatorStore`.  No common concrete function could be abstracted into a new abstract class. 👀😅

MauricioUyaguari (approved):

PR #3017, opened 2024-03-22
Support server side push down operations for REPL using Ag-grid

gayathrir11:

- [ ] Test(s) added


packages/legend-application-repl/src/components/grid/ServerSideDataSource.ts
```diff
@@ -0,0 +1,184 @@
...
+    this.editorStore = editorStore;
+  }
+
+  *fetchRows(
+    params: IServerSideGetRowsParams<unknown, unknown>,
+  ): GeneratorFn<void> {
+    try {
+      if (this.executions > 0) {
+        if (this.editorStore) {
+          const request = this.extractRequest(params);
```
MauricioUyaguari, line 74: i see no tests in this PR :). probably a good place to start is the extracting of the request from the ag grid params.

packages/legend-application-repl/src/components/grid/TDSRequest.ts
```diff
@@ -0,0 +1,153 @@
...
+    groupKeys: string[],
+    aggregations: TDSAggregation[],
+  ) {
+    this.columns = columns;
+    this.groupKeys = groupKeys;
+    this.aggregations = aggregations;
+  }
+}
+
+export class TDSRequest {
```
MauricioUyaguari, line 130: is this class specific to the repl server? 
what happens in the repl, takes the current query and just appends this?

MauricioUyaguari (approved):
small comments but looks good to merge. Main thing I would say is you should set up some testing for creating the tds request based on ag grid params inputs.
Additionally, we should clean this up so we can plug in any other type of tds request. ideally a lambda so we can use the engine execute endpoint.

packages/legend-application-repl/src/components/grid/TDSRequest.ts
gayathrir11: yeah it just replaces existing lambda with the lambda i send and execute it

packages/legend-application-repl/src/components/grid/ServerSideDataSource.ts
gayathrir11: Will address that in a different commit

packages/legend-application-repl/src/stores/REPLGridState.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  setCurrentResult(val: TDSExecutionResult | undefined): void {
+    this.currentResult = val;
+  }
+
+  setColumns(val: string[]): void {
+    this.columns = val;
+  }
+
+  setLicenseKey(val: string | undefined): void {
```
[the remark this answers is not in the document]
gayathrir11: The reason we were able to do that in other parts of codebase is we have a different deployment webpack config for studio where we add this license key. But in this case i would just be using the default deployemnt config we have externally into repl client code. Hence we thought it is better to just configure it in repl client in backend

PR #3030, opened 2024-03-29
Enhance update parameter modal

travisstebbins:
## Summary

Make the update parameter modal have an Apply and Cancel button so that changes made in the modal only take effect if Apply is clicked. If Cancel is clicked, then the changes made in the modal shouldn't be saved to the state.

- [x] Test(s) added

Validation of parameter name when creating a new parameter (against existing parameter and constant names):
CreateParameter_Validation

Validation of parameter name when updating a parameter (against existing parameter and constant names):
UpdateParameter_Validation

Showing that changes don't take effect when Cancel button is clicked, but changes do take effect when Apply button is clicked
UpdateParameter_ApplyCancel

Showing that changes made in the create parameter modal take effect when you create a new parameter
CreateParameter_ChangeValues

### DIFF
--- a/packages/legend-query-builder/src/components/QueryBuilderParametersPanel.tsx
+++ b/packages/legend-query-builder/src/components/QueryBuilderParametersPanel.tsx
@@ -133,15 +133,14 @@ const VariableExpressionEditor = observer(
     };
 
     const handleApply = (): void => {
+      lambdaParameterState.changeVariableType(selectedType.value);
+      lambdaParameterState.changeMultiplicity(
+        varState,
+        selectedMultiplicity.value,
+      );
+      variableExpression_setName(varState, selectedName ?? '');
       if (isCreating) {
         queryParametersState.addParameter(lambdaParameterState);
-      } else {
-        lambdaParameterState.changeVariableType(selectedType.value);
-        lambdaParameterState.changeMultiplicity(
-          varState,
-          selectedMultiplicity.value,
-        );
-        variableExpression_setName(varState, selectedName ?? '');
       }
       handleCancel();
     };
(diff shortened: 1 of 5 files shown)


packages/legend-query-builder/src/components/QueryBuilderParametersPanel.tsx
```diff
@@ -108,46 +109,47 @@ const VariableExpressionEditor = observer(
...
-          (isCreating ? 0 : 1)
-        ? 'Parameter name already exists'
-        : (isCreating &&
-            queryParametersState.parameterStates.find(
-              (p) => p.parameter.name === input,
-            )) ||
-          (!isCreating &&
-            queryParametersState.parameterStates.filter(
-              (p) => p.parameter.name === input,
-            ).length > 1)
```
travisstebbins, line 131: `allVariableNames` already contains all the names of the variables from `queryParametersState.parameterStates`, so these lines seemed redundant; hence, I removed them. Let me know if there was a reason for having this check here, too.

packages/legend-query-builder/src/components/__tests__/QueryBuilderParametersPanel.test.tsx
```diff
@@ -107,7 +172,7 @@ test(
 
 test(
   integrationTest(
-    'Query builder shows validation error for constant name if existing duplicate parameter name',
```
travisstebbins, line 110: There were a bunch of tests related to the constants panel in this file, so I moved them from here to `QueryBuilderConstantsPanel.test.tsx`

packages/legend-query-builder/src/components/QueryBuilderParametersPanel.tsx
YannanGao-gs: I think it's safe to remove it as the var name got changed immediately before so the check number for create/update is different.

PR #3031, opened 2024-04-01
Enhance constants modal

travisstebbins:
## Summary

Make the constants modal have an Apply and Cancel button so that changes made in the modal only take effect if Apply is clicked. If Cancel is clicked, then the changes made in the modal shouldn't be saved to the state.

- [x] Test(s) added

Setting various constant types and values:
SetConstantValues

Setting an invalid expression for numeric type constant:
NumericConstant_InvalidExpression

Creating a constant with a duplicate name:
CreateConstant_DuplicateConstantName

Updating a constant with a duplicate name:
UpdateConstant_DuplicateConstantName

Creating a constant and updating the values in the modal before clicking "Create":
CreateConstant_ModalChangesTakeEffect

Cancel button doesn't update the state:
UpdateConstant_CancelButton

Update button does update the state:
UpdateConstant_UpdateButton

Updating derived constant name with duplicate name:
DerivedConstant_DuplicateConstantName

Cancel button doesn't update the state name or lambda for derived constant:
DerivedConstant_CancelButton

Update button does update the state name and lambda for derived constant:
DerivedConstant_UpdateButton

### DIFF
--- a/packages/legend-query-builder/src/components/QueryBuilderConstantExpressionPanel.tsx
+++ b/packages/legend-query-builder/src/components/QueryBuilderConstantExpressionPanel.tsx
@@ -140,10 +140,10 @@ const QueryBuilderSimpleConstantExpressionEditor = observer(
     const resetConstantValue = (): void => {
       const valSpec = buildDefaultInstanceValue(
         queryBuilderState.graphManagerState.graph,
-        stateType,
+        selectedType.value,
         queryBuilderState.observerContext,
       );
-      constantState.setValueSpec(valSpec);
+      setSelectedValue(valSpec);
     };
 
     return (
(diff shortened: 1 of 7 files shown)


YannanGao-gs (dismissed):

packages/legend-query-builder/src/components/QueryBuilderConstantExpressionPanel.tsx
```diff
@@ -43,45 +42,80 @@ import {
...
+import React, { useEffect, useMemo, useState } from 'react';
 import { variableExpression_setName } from '../stores/shared/ValueSpecificationModifierHelper.js';
 import { LambdaEditor } from './shared/LambdaEditor.js';
 import { VariableViewer } from './shared/QueryBuilderVariableSelector.js';
 import { flowResult } from 'mobx';
 
+const getConstantNameValidationMessage = (
+  constantInput: string,
+  allVariableNames: string[],
+  stateName: string,
```
MauricioUyaguari, line 71: what do we mean by stateName here? maybe a better name would explain the funtion a bit better.

MauricioUyaguari:
should be also be consistent. i think we use apply and update for updating properties. 
Shoud we change it to one or the other. Apply may be best.

packages/legend-query-builder/src/components/QueryBuilderConstantExpressionPanel.tsx
travisstebbins: It's the current name of the constant in the state. I'll rename it to be clearer.

MauricioUyaguari (approved):

MauricioUyaguari (approved):

PR #3032, opened 2024-04-03
Enhance watermark modal

travisstebbins:
## Summary

Make the watermark modal have an Apply and Cancel button so that changes made in the modal only take effect if Apply is clicked. If Cancel is clicked, then the changes made in the modal shouldn't be saved to the state.

- [x] Test(s) added

Setting values in the watermark modal:
WatermarkModal

Setting values with parameters/constants in the watermark modal:
WatermarkModal_WithParamsAndConsts

### DIFF
--- a/packages/legend-query-builder/src/components/watermark/QueryBuilderWatermark.tsx
+++ b/packages/legend-query-builder/src/components/watermark/QueryBuilderWatermark.tsx
@@ -33,10 +33,11 @@ import {
   PrimitiveType,
   type VariableExpression,
   type ValueSpecification,
-  type GraphManagerState,
+  PrimitiveInstanceValue,
+  GenericType,
+  GenericTypeExplicitReference,
 } from '@finos/legend-graph';
 import { observer } from 'mobx-react-lite';
-import { createViewModel } from 'mobx-utils';
 import { useCallback, useState } from 'react';
 import { useDrop } from 'react-dnd';
 import type { QueryBuilderState } from '../../stores/QueryBuilderState.js';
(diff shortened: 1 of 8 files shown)


packages/legend-query-builder/src/components/watermark/QueryBuilderWatermark.tsx
```diff
@@ -101,20 +109,18 @@ const WatermarkValueEditor = observer(
             dropTargetConnector={dropTargetConnector}
           >
             <BasicValueSpecificationEditor
-              valueSpecification={watermarkValue}
+              valueSpecification={selectedValue}
               setValueSpecification={(val: ValueSpecification): void => {
-                watermarkState.setValue(val);
+                setSelectedValue(deepClone(val));
```
MauricioUyaguari, line 114: is this deepClone necessary, since we already deep clone the orginal value. woudnt this update just the cloned value spec?

travisstebbins: The `deepClone` seems to be necessary because the `val` parameter in the `setValueSpecification` callback seems to be the same reference as is passed in in the `valueSpecification`. Without the `deepClone` (if we just call `setValueSpecification(val)`) the component doesn't re-render because the prop is technically unchanged (its pointing to the same object). I think if it were a primitive value we wouldn't need the `deepClone`, but since it's an object, the `deepClone` creates a branch new object, which causes the component to re-render since the prop changes.

MauricioUyaguari: mhm if thats the case, do we need a deepClone for this. just a shallow clone should be enough?

travisstebbins: Good point. I switched it to shallow clone and it worked fine. So now we only `deepClone` in the `useState` hook to copy the object from the state and shallow clone it when updating it from the editor.

MauricioUyaguari (approved):

PR #3041, opened 2024-04-08
Enhance window function modal

travisstebbins:
## Summary

Make the window function modal have an Apply and Cancel button so that changes made in the modal only take effect if Apply is clicked. If Cancel is clicked, then the changes made in the modal shouldn't be saved to the state.

Additionally, fix some validations that were missing:
- Don't allow duplicate window columns
- Disable the "Run Query" button if a window column has the same name as a projection column or another window column

- [x] Test(s) added

Duplicate column name validation:
DuplicateNames

Preventing duplicate window columns by removing used window columns from the window column drop downs:
DisallowDuplicateWindowColumns

Cancel button doesn't save changes to state:
CancelButton

Apply button does save changes to state:
ApplyButton


packages/legend-query-builder/src/components/fetch-structure/QueryBuilderTDSWindowPanel.tsx
```diff
@@ -89,14 +95,20 @@ const createWindowColumnState = (
       operator,
       columnState,
     );
+    const columnName = operator.isColumnAggregator()
```
YannanGao-gs, line 98: maybe we could aggregate line 98 and 109 together to create columnName above if, but current code looks good too 
just a personal preference

YannanGao-gs (approved):

YannanGao-gs (approved):

PR #3094, opened 2024-04-29
No default filter/post-filter, constant, and derived property values

travisstebbins:
## Summary

To prevent users from accidentally using incorrect values in their filters/post-filters, constants, and derived properties, we will not set default values when a user adds these to their query. These values (filter/post-filter, constant, and derived property) will show a red input box and the panels will have error messages for null values, so the user will be forced to select a value. We will also disable any actions being done besides editing the query, since the user should not be able to perform major query actions on an invalid query.

- [x] Test(s) added

Adding properties to the filter panel:
FilterValues

Using filter operation that takes a list:
FilterList

Adding properties to the post-filter panel:
PostFilterValues

Adding a derived property with a required parameter:
RequiredDerivedProperty

Adding a derived property with an optional parameter:
OptionalDerivedProperty

Creating a new constant:
Constants

### DIFF
--- a/packages/legend-application-query/src/components/__tests__/QueryEditorQueryActions.test.tsx
+++ b/packages/legend-application-query/src/components/__tests__/QueryEditorQueryActions.test.tsx
@@ -27,6 +27,7 @@ import {
   act,
   fireEvent,
   getByDisplayValue,
+  getByRole,
   getByText,
   getByTitle,
   waitFor,
@@ -36,7 +37,10 @@ import {
   TEST__provideMockedQueryEditorStore,
   TEST__setUpQueryEditor,
 } from '../__test-utils__/QueryEditorComponentTestUtils.js';
-import { QUERY_BUILDER_TEST_ID } from '@finos/legend-query-builder';
+import {
+  QUERY_BUILDER_TEST_ID,
+  dragAndDrop,
+} from '@finos/legend-query-builder';
 import {
   TEST_DATA__ResultState_entities,
   TEST_DATA__result,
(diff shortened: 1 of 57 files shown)


packages/legend-query-builder/src/stores/shared/ValueSpecificationEditorHelper.ts
```diff
@@ -134,19 +136,27 @@ export const buildDefaultInstanceValue = (
...
 ): ValueSpecification => {
   const path = type.path;
   switch (path) {
     case PRIMITIVE_TYPE.STRING:
-    case PRIMITIVE_TYPE.BOOLEAN:
     case PRIMITIVE_TYPE.STRICTDATE:
     case PRIMITIVE_TYPE.DATETIME:
     case PRIMITIVE_TYPE.NUMBER:
     case PRIMITIVE_TYPE.DECIMAL:
     case PRIMITIVE_TYPE.FLOAT:
```
YannanGao-gs, line 148: where are `case PRIMITIVE_TYPE.BINARY` and `case PRIMITIVE_TYPE.BYTE`?

packages/legend-application-query/src/components/__tests__/QueryEditorQueryActions.test.tsx
```diff
@@ -225,3 +229,183 @@ test(
...
+      getByRole(queryActionsPanel, 'button', { name: 'Save' }).hasAttribute(
+        'disabled',
+      ),
+    ).toBe(true);
+    expect(
+      getByRole(queryActionsPanel, 'button', {
+        name: 'Save As...',
+      }).hasAttribute('disabled'),
+    ).toBe(true);
+  },
```
YannanGao-gs, line 317: would you mind also testing the issue-count-badge you added?

packages/legend-query-builder/src/stores/shared/ValueSpecificationEditorHelper.ts
travisstebbins: I talked with @MauricioUyaguari about it and we decided to remove it, because it's not actually supported in constants and filters.
- If you try to DND a byte property into the filter panel, you get an error because no operations are registered for the byte type.
- If you create a binary or byte constant and then go to "Edit Query in Pure," you also get an error because it's unable to build the query.
Since it's not actually supported, we decided to remove it, and also removed binary and byte from the constant type selection dropdown, since no one should be using it.

YannanGao-gs: https://github.com/finos/legend-studio/pull/2333
Could you please confirm this PR won't be broken?

travisstebbins: It looks like the same actions still work:
ByteEditor

I think the only places where I would have concerns are the `EqualToAsssertionEditor` component in `packages/legend-application-studio/src/components/editor/editor-group/testable/TestableSharedComponents.tsx` and the `createFunctionTest` function in `packages/legend-application-studio/src/stores/editor/editor-state/element-editor-state/function-activator/testable/FunctionTestableState.ts`. Do you know if that component/function would ever get used with a byte or binary type?

packages/legend-application-query/src/components/__tests__/QueryEditorQueryActions.test.tsx
travisstebbins: Done

packages/legend-query-builder/src/stores/shared/ValueSpecificationEditorHelper.ts
YannanGao-gs: May I know if there is a lot of extra work you need to do for adding these two back?
Besides this one, your PR LGTM.

travisstebbins: I think it shouldn't be an issue to add these two back, especially since we removed them from the constants type dropdown. I'll work on adding them back.

travisstebbins: I have re-added BYTE and BINARY to the function switch statement.

YannanGao-gs (dismissed):

packages/legend-query-builder/src/components/QueryBuilderConstantExpressionPanel.tsx
```diff
@@ -121,28 +131,33 @@ const QueryBuilderSimpleConstantExpressionEditor = observer(
           queryBuilderState.graphManagerState.graph,
           val.value,
           queryBuilderState.observerContext,
+          queryBuilderState.INTERNAL__enableInitializingDefaultSimpleExpressionValue,
         );
         setSelectedValue(newValSpec);
       }
+      setShouldFocusOnValue(true);
     };
+    const supportedPrimitiveTypes: PrimitiveType[] = [
```
MauricioUyaguari, line 140: lets add a comment here on the ones we are disabling

MauricioUyaguari (changes requested):

packages/legend-query-builder/src/components/QueryBuilderConstantExpressionPanel.tsx
travisstebbins: Done

packages/legend-query-builder/src/stores/QueryBuilderPropertyEditorState.ts
```diff
@@ -142,44 +138,23 @@ export const getPropertyPath = (
...
-            type === PrimitiveType.DATE ? PrimitiveType.STRICTDATE : type,
-          ),
-        ),
-      );
-      if (type !== PrimitiveType.LATESTDATE) {
-        instanceValue_setValues(
-          primitiveInstanceValue,
-          [generateDefaultValueForPrimitiveType(type.name as PRIMITIVE_TYPE)],
-          observerContext,
+    if (type instanceof PrimitiveType || type instanceof Enumeration) {
```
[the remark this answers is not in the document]
travisstebbins: Added a test here with a derived property with enum parameter: https://github.com/finos/legend-studio/pull/3094/files#diff-1fb89b80940400504f134ed1a8252e470d4e8bd6587e1a87a462461e4daed8d7R96

packages/legend-query-builder/src/stores/shared/ValueSpecificationEditorHelper.ts
```diff
@@ -80,7 +80,9 @@ export const buildPrimitiveInstanceValue = (
       new GenericType(graph.getPrimitiveType(type)),
     ),
   );
-  instanceValue_setValues(instance, [value], observerContext);
+  if (value !== undefined) {
```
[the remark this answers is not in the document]
travisstebbins: Yeah `null` is the case the user hasn't set the value, so it should be `if (value !== null)`, not `if (value !== undefined)`. However, the current behavior sets it to an array like `[undefined]` if the value is `undefined`, not an empty array, so I think we should leave that behavior as is instead of setting it to an empty array if value is `undefined`.

packages/legend-query-builder/src/stores/filter/operators/QueryBuilderFilterOperator_LessThanEqual.ts
```diff
@@ -94,22 +93,15 @@ export class QueryBuilderFilterOperator_LessThanEqual
       case PRIMITIVE_TYPE.FLOAT:
       case PRIMITIVE_TYPE.INTEGER:
       case PRIMITIVE_TYPE.STRICTDATE:
-      case PRIMITIVE_TYPE.DATETIME: {
```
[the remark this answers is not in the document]
travisstebbins: Added a test that uses datetime type property: https://github.com/finos/legend-studio/pull/3094/files#diff-8d686aaab2ab24e211af8615cc4382680d9bde556f8921a1f9da82d69170fae2R2327

packages/legend-query-builder/src/stores/shared/ValueSpecificationEditorHelper.ts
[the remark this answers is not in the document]
travisstebbins: Removed the conditional statement so we aren't changing the behavior here. If `value` is null, we will create an array like `[null]` and if it's undefined, it will be `[undefined]`.

MauricioUyaguari (approved):

PR #3108, opened 2024-05-03
Clean Legend Query Header I  - add query application logo header

YannanGao-gs:
## Summary

Clean Legend Query Header I  - add query application logo header

## How did you test this change?

- [ ] Test(s) added
- [x] Manual testing (please provide screenshots/recordings)
- [ ] No testing (please provide an explanation)

Screenshot 2024-05-03 at 3 04 32 PM

Screenshot 2024-05-03 at 3 03 54 PM

PR #3112, opened 2024-05-04
Clean Legend Query Header II - aggregate Legend Query and Query Builder action header

YannanGao-gs:
## Summary

aggregate Legend Query and Query Builder action header

depends on https://github.com/finos/legend-studio/pull/3108

- [ ] Test(s) added

Screenshot 2024-05-03 at 9 45 14 PM

Screenshot 2024-05-03 at 9 45 23 PM

Screenshot 2024-05-03 at 9 45 30 PM


packages/legend-query-builder/src/components/QueryBuilder.tsx
```diff
@@ -405,244 +281,46 @@ export const QueryBuilder = observer(
...
-                  />
-                  <RedoButton
-                    parent={queryBuilderRef}
-                    canRedo={
-                      queryBuilderState.changeHistoryState.canRedo &&
-                      queryBuilderState.isQuerySupported
-                    }
-                    redo={redo}
-                  />
+            {!hideQueryBuilderHeader && (
```
MauricioUyaguari, line 284: seems like we are over doing it here with extensions. can we please keep the core actions directly on this editor.
Then lets think about how to separate the different buttons coming in. 

Maybe we can group the saved query ones into an extension which would include, productize query, dev service, rename query, new query, and a loader ?
that way if something wants to provide this stored query extension, they would  need to provide those items.

packages/legend-query-builder/src/stores/QueryBuilderState.ts
```diff
@@ -146,6 +148,7 @@ export abstract class QueryBuilderState implements CommandRegistrar {
   isCalendarEnabled = false;
   isQueryChatOpened = false;
   isLocalModeEnabled = false;
+  queryBuilderRef!: React.RefObject<HTMLDivElement>;
```
MauricioUyaguari, line 151: can you explain the user of this and why it is needed

packages/legend-query-builder/src/components/QueryBuilder.tsx
```diff
@@ -229,127 +215,13 @@ export const QueryBuilder = observer(
   (props: { queryBuilderState: QueryBuilderState }) => {
     const { queryBuilderState } = props;
     const applicationStore = queryBuilderState.applicationStore;
+    const hideQueryBuilderHeader = applicationStore.pluginManager
```
MauricioUyaguari, line 218: as mentioned we should not be making any changes in legend-application. 

Two modules i see changes are in query-builder and application-query. 

Also are we really hiding the header in application query. the header is just a simple panel with Legend Query right?

packages/legend-application-query/src/application/LegendQueryApplicationConfig.ts
```diff
@@ -70,6 +70,8 @@ class LegendQueryApplicationCoreOptions {
    */
   queryBuilderConfig: QueryBuilderConfig | undefined;
 
+  applicationName = 'Legend Query';
```
MauricioUyaguari, line 73: is this really needed or can't we just load the query-builder header with the panel having this?

MauricioUyaguari (commented):
I think we need to think more about how the extension lay out is thought here. Also lets combine the other Clean Legend QUery Header I
into this PR

PR #3179, opened 2024-05-22
enable new query button in query create flow

YannanGao-gs:
## Summary

enable new query button in query create flow

- [ ] Test(s) added

### DIFF
--- a/packages/legend-application-query/src/components/Core_LegendQueryApplicationPlugin.tsx
+++ b/packages/legend-application-query/src/components/Core_LegendQueryApplicationPlugin.tsx
@@ -50,6 +50,8 @@ import {
   LEGEND_QUERY_ROUTE_PATTERN,
 } from '../__lib__/LegendQueryNavigation.js';
 import {
+  ActionAlertActionType,
+  ActionAlertType,
   type ApplicationPageEntry,
   type LegendApplicationSetup,
 } from '@finos/legend-application';
@@ -65,9 +67,9 @@ import {
   generateDataSpaceQueryCreatorRoute,
   generateDataSpaceQuerySetupRoute,
 } from '../__lib__/DSL_DataSpace_LegendQueryNavigation.js';
-import type {
-  QueryBuilderHeaderActionConfiguration,
-  QueryBuilderMenuActionConfiguration,
+import {
+  type QueryBuilderHeaderActionConfiguration,
+  type QueryBuilderMenuActionConfiguration,
 } from '@finos/legend-query-builder';
 import {
   ExistingQueryEditorStore,
@@ -425,8 +427,6 @@ export class Core_LegendQueryApplicationPlugin extends LegendQueryApplicationPlu
           ) {
             const editorStore =
               queryBuilderState.workflowState.actionConfig.editorStore;
-            const isExistingQuery =
-              editorStore instanceof ExistingQueryEditorStore;
             const handleNewQuery = (): void => {
               if (editorStore instanceof ExistingQueryEditorStore) {
                 const query = editorStore.query;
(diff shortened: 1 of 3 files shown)


packages/legend-application-query/src/components/Core_LegendQueryApplicationPlugin.tsx
```diff
@@ -463,24 +465,85 @@ export class Core_LegendQueryApplicationPlugin extends LegendQueryApplicationPlu
                     }
                   }
                 }
+              } else if (editorStore instanceof MappingQueryCreatorStore) {
```
MauricioUyaguari, line 468: Do we really need to navigate users ? 
I would expect us to just clear the current query if they are not in  `ExistingQueryEditorStore`.
We should also add a blocking modal if there are changes in their query.

MauricioUyaguari (changes requested):

packages/legend-application-query/src/components/Core_LegendQueryApplicationPlugin.tsx
```diff
@@ -463,24 +463,47 @@ export class Core_LegendQueryApplicationPlugin extends LegendQueryApplicationPlu
                     }
                   }
                 }
+              } else {
+                if (queryBuilderState.changeDetectionState.hasChanged) {
```
MauricioUyaguari, line 467: i think here we can just check if the rawLambda is not stub right ?
then they will have changes

packages/legend-application-query/src/stores/QueryEditorStore.ts
```diff
@@ -525,6 +525,9 @@ export abstract class QueryEditorStore {
       this.queryBuilderState = (yield this.initializeQueryBuilderState(
         stopWatch,
       )) as QueryBuilderState;
+      this.queryBuilderState.changeDetectionState.initialize(
```
MauricioUyaguari, line 528: why do we need this ?

MauricioUyaguari (changes requested):

packages/legend-application-query/src/components/Core_LegendQueryApplicationPlugin.tsx
YannanGao-gs: export const isStubbed_RawLambda = (rawLambda: RawLambda): boolean =>
  !rawLambda.body && !rawLambda.parameters;

May I know if you are referring to this function above?

If so, I don't think it works because the rawLambda body is not undefined as it has getAll()

packages/legend-application-query/src/stores/QueryEditorStore.ts
YannanGao-gs: This is for making sure `queryBuilderState.changeDetectionState.hasChanged` is functioning 
without this, change detection doesn't detect there are changes

MauricioUyaguari: and at what point are we comparing the query? 
does have `Class.all()` count as a change ?

YannanGao-gs: I don't think .all() should be counted as a change e.g. for the`creating a query from mapping` entry, the class has been automatically chosen.

MauricioUyaguari: yeah i dont either, which is why i asked the question, where do we start the change detection

YannanGao-gs: since the query doesn't just contain .all()

 `check if the rawLambda is not stub` will mark .all() as a change

MauricioUyaguari: My question was more on using change detection method. 
Would we consider .all() as a change using change detection.

YannanGao-gs: No, it only will mark there is a change if the query doesn't just contain .all()

MauricioUyaguari (approved):

PR #3185, opened 2024-05-23
Allow DND into entire fetch structure panel

travisstebbins:
## Summary

Allow the user to DND properties into the entire fetch structure panel instead of just below the existing columns. This makes it easier for users to add columns to the fetch structure panel, especially when there are a lot of existing columns in the panel already. This removes the ability for the user to replace an existing property in the fetch structure and filter panels by DND a property over an existing property name.

- [ ] Test(s) added


MauricioUyaguari:
@travisstebbins just confirming we can still rearrange columns through drag and drop?

travisstebbins:
> @travisstebbins just confirming we can still rearrange columns through drag and drop?

Good call out. Yes we can still rearrange columns through drag and drop. I added a screen recording of this.

PR #3194, opened 2024-05-28
Default to last dataspace

codrex:
## Summary
This PR adds a feature that allows legend query to default to last dataspace/class used when opening up new query from dataspace

- [X] Test(s) added

### DIFF
--- a/.changeset/thirty-deers-swim.md
+++ b/.changeset/thirty-deers-swim.md
@@ -1,0 +1,3 @@
+---
+'@finos/legend-application': patch
+---
--- a/.changeset/thirty-deers-swim.md
+++ b/.changeset/thirty-deers-swim.md
@@ -1,3 +1,5 @@
 ---
 '@finos/legend-application': patch
 ---
+
+Add save last queried dataspace functionality to the LegendQueryUserDataHelper class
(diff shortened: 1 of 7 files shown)


travisstebbins:
We'll need to add a few additional error handling/edge case handling to this code:
1. Handle if the most recent data space/class no longer exist.
2. Handle if an error is thrown while loading the most recent data space.
In both of these cases we should make sure to remove the dataspace/class from the most recent list and redirect to the next most recent valid data space/class combo, or to the default data space selector page if no more valid entries exist in the recent list.

We should look at the `initializeQueryBuilderState` function in `packages/legend-application-query/src/stores/data-space/DataSpaceQuerySetupState.ts` or this block
```
useEffect(() => {
      flowResult(queryBuilderState.loadDataSpaces()).catch(
        applicationStore.alertUnhandledError,
      );
    }, [queryBuilderState, applicationStore]);
```
in `packages/legend-application-query/src/components/data-space/DataSpaceQuerySetup.tsx` for possible areas to add this logic.

MauricioUyaguari:
@travisstebbins @codrex 
Merging the teest part of this.
As discussed with @travisstebbins we need to put this in the proper place as well as handle the error messaging.
Additionally, we should not be saving urls but dataspace name etc

MauricioUyaguari:
closing in favor of this https://github.com/finos/legend-studio/pull/3200 see @travisstebbins  comments as to improvements needed

PR #3196, opened 2024-05-28
bump dependencies and improvement for REPL

akphi:


akphi:
Consider this order of operation

```
TDS (accessor)
->extend() // leaf level
->rename() // since rename() only supports one column
->filter() // DataCube either puts filter before the leaf level extend() or disallows filtering on extended columns
->groupBy() // v-pivot (form the tree map on the left)
->select() // column omitted by select() will not show up in pivot()
->pivot() // h-pivot
->extend() // group level
->sort() // excludes
->limit()
(->slice()) // an optimization we can do to lessen memory usage
->from() // DB
```

MauricioUyaguari (approved):

PR #3198, opened 2024-05-29
Optimize "is in" filter

travisstebbins:

- [ ] Test(s) added

### DIFF
--- a/packages/legend-query-builder/src/components/shared/BasicValueSpecificationEditor.tsx
+++ b/packages/legend-query-builder/src/components/shared/BasicValueSpecificationEditor.tsx
@@ -657,6 +657,7 @@ const PrimitiveCollectionInstanceValueEditor = observer(
     const applicationStore = useApplicationStore();
     const inputRef = useRef(null);
     const [inputValue, setInputValue] = useState('');
+    const [inputValueIsError, setInputValueIsError] = useState(false);
     const [showAdvancedEditorPopover, setShowAdvancedEditorPopover] =
       useState(false);
     const [selectedOptions, setSelectedOptions] = useState<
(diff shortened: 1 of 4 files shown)


packages/legend-query-builder/src/components/__tests__/QueryBuilderFetchStructure.test.tsx
```diff
@@ -1309,74 +1307,6 @@ test(
   },
 );
 
-test(
-  integrationTest(
-    'Query builder filter supports actions/flows for is in clause',
```
travisstebbins, line 1314: I removed this test as it's already covered by tests in `packages/legend-query-builder/src/components/__tests__/QueryBuilderFilterPanel.test.tsx`.

packages/legend-query-builder/src/components/shared/BasicValueSpecificationEditor.tsx
```diff
@@ -628,153 +623,271 @@ const stringifyValue = (values: ValueSpecification[]): string => {
...
+          }))
+        : undefined;
+    const noMatchMessage =
+      isTypeaheadSearchEnabled && isLoading ? 'Loading...' : undefined;
+
+    // helper functions
+    const buildOptionForValueSpec = (
+      value: ValueSpecification,
+    ): { label: string; value: string } => ({
+      label: guaranteeNonNullable(getValueSpecificationStringValue(value)),
```
YannanGao-gs, line 709: maybe we could create a local variable with value `getValueSpecificationStringValue(value)` 
so that this calculation will only be done once

YannanGao-gs (dismissed):

packages/legend-query-builder/src/components/shared/BasicValueSpecificationEditor.tsx
travisstebbins: Good point! Made the change

MauricioUyaguari (approved):

PR #3212, opened 2024-05-31
Display 10 default queries when there is no recently reviewed query

YannanGao-gs:
## Summary

Display 10 default queries when there is no recently reviewed query

- [ ] Test(s) added

### DIFF
--- a/packages/legend-application-query/src/stores/data-space/DataSpaceQueryCreatorStore.ts
+++ b/packages/legend-application-query/src/stores/data-space/DataSpaceQueryCreatorStore.ts
@@ -20,6 +20,8 @@ import {
   RuntimePointer,
   PackageableElementExplicitReference,
   type RawLambda,
+  QueryProjectCoordinates,
+  QuerySearchSpecification,
 } from '@finos/legend-graph';
 import {
   type DepotServerClient,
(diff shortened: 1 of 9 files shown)


packages/legend-application-query/src/stores/data-space/DataSpaceQueryCreatorStore.ts
```diff
@@ -342,4 +348,29 @@ export class DataSpaceQueryCreatorStore extends QueryEditorStore {
       ),
     );
   }
+
```
MauricioUyaguari, line 351: what happens when user enters the default route of query (no project info). 
Do we have a search specification for this case ?

YannanGao-gs: by default,
the function is empty 

https://github.com/finos/legend-studio/pull/3212/files#diff-5dae4710ada9d402191df7256d5cc7d4502e3b40a8f1b9acf960143f7ee6adfeR437

MauricioUyaguari: yeah but what does that return in terms of queries ?

MauricioUyaguari:
Does it make sensse to also run a query on queries owned by current user?

MauricioUyaguari (approved):

PR #3226, opened 2024-06-06
Enable UI support for percentile aggregation

YannanGao-gs:
## Summary

Enable UI support for percentile aggregation

- [x] Test(s) added

https://github.com/finos/legend-studio/assets/73408381/65e3a735-9ffc-40c6-82f4-3969184957cd

Screenshot 2024-06-07 at 11 19 04 AM

https://github.com/finos/legend-studio/assets/73408381/31cb605a-09e9-4a12-8bf5-1ea48047b5c2


MauricioUyaguari:
Nice!
Couple of feedback here:
1. can we add a roundtrip tests using percentile 
2. can we improve and follow the default value pattern here. When percentile is selected as an aggregation, there should be a validation error on top indication that the query is invalid as no percentile number was written. As well as some kind of UI indicator that an input is needed.

YannanGao-gs:
> Nice! Couple of feedback here:
> 
> 1. can we add a roundtrip tests using percentile
> 2. can we improve and follow the default value pattern here. When percentile is selected as an aggregation, there should be a validation error on top indication that the query is invalid as no percentile number was written. As well as some kind of UI indicator that an input is needed.

Thanks

- [ ] One roundtrip test has been added
- [ ] a validation error has been added when the percentile value input is not set

packages/legend-manual-tests/src/__tests__/roundtrip-grammar/cases/DSL_Service-with-percentile-aggregation-func.pure
```diff
@@ -0,0 +1,208 @@
+###Service
```
MauricioUyaguari, line 1: nice but more thinking about lambda and querybuilderstate roundtrip tests

MauricioUyaguari (approved):

PR #3237, opened 2024-06-11
Enable default value flow for milestoning parameters

YannanGao-gs:
## Summary

Enable default value flow for milestoning parameters

This PR also contains changes in https://github.com/finos/legend-studio/pull/3230 

- [ ] Test(s) added


packages/legend-query-builder/src/stores/milestoning/QueryBuilderMilestoningHelper.ts
```diff
@@ -316,10 +317,54 @@ export const validateMilestoningPropertyExpressionChain = (
           `Property of milestoning sterotype '${MILESTONING_STEREOTYPE.BITEMPORAL}' should not have more than two parameters`,
         );
       }
+      queryBuilderState.milestoningState.setProcessingDate(
+        propertyExpression.parametersValues[1],
```
gayathrir11, line 321: Why did we need to add this logic to validation stage? Don't we want to throw errors in UI if people do something unexpected in text mode?

packages/legend-query-builder/src/components/fetch-structure/QueryBuilderResultModifierPanel.tsx
```diff
@@ -380,6 +385,283 @@ export const QueryResultModifierModal = observer(
...
+        }
+      }
+      setIsAllVersionsEnabled(Boolean(value));
+    };
+
+    const buildAllVersionsInRangeParameters = (): void => {
+      let startDateParameterState;
+      let endDateParameterState;
+      if (!(startDate && startDate instanceof VariableExpression)) {
+        const startDateVar = new VariableExpression(
```
gayathrir11, line 525: Does it mean we don't let users to have hardcoded startDate and endDate?

packages/legend-query-builder/src/components/fetch-structure/QueryBuilderResultModifierPanel.tsx
```diff
@@ -380,6 +385,283 @@ export const QueryResultModifierModal = observer(
...
+                (milestoningState.businessDate &&
+                milestoningState.businessDate instanceof VariableExpression
+                  ? milestoningState.businessDate.name
+                  : BUSINESS_DATE_MILESTONING_PROPERTY_NAME),
+            )?.parameter,
+          );
+        }
+        if (!processingDate) {
+          setProcessingDate(
+            allParamStates.find(
```
gayathrir11, line 507: Can you check if it works for the case when the root class is not milestoned but one of the property in projection is milestoned. I see that you are updating newparamstates based on root class milestoning type, then it wouldn't have any paramter named processingDate or businessdate to set these values if one of the property in projection is milestoned.

gayathrir11 (approved):

packages/legend-query-builder/src/components/fetch-structure/QueryBuilderResultModifierPanel.tsx
YannanGao-gs: yeah this case also works as I keep any used milestoning parameters 

https://github.com/finos/legend-studio/assets/73408381/49a96122-b0af-4899-8c0e-b4a81d42de9b

packages/legend-query-builder/src/stores/milestoning/QueryBuilderMilestoningHelper.ts
YannanGao-gs: The changes here are more about initializing `milestongState`. 
If the root class is not milestoning, I found that the `milestongState`  is not initialized at all.

packages/legend-query-builder/src/components/fetch-structure/QueryBuilderResultModifierPanel.tsx
YannanGao-gs: https://github.com/finos/legend-studio/assets/73408381/a8d0b9eb-d6e7-442d-8994-656aca39ff20

This is the demo video for the case where the root class could be set to use allVersionsInRange 


https://github.com/finos/legend-studio/assets/73408381/c4f482e8-cee8-445f-9c86-959e46e3fa9f

MauricioUyaguari (approved):

PR #3285, opened 2024-06-24
Enhance query filter editors

travisstebbins:

- [x] Test(s) added

Enhance the query filter editor to behave similarly to fetch structure column name editor. Filter condition values should be displayed as blue editable text, and clicking the value shows the BasicValueSpecificationEditor to allow editing the value.

For date values, I have updated the `DateInstanceValueEditor` to take a `displayAsEditableValue` prop, so that instead of showing the default trigger button, it displays the blue, underlined editable text, and clicking on the text brings up the date picker popover.

"Is in"/"Is not in" filter value editors are not changed.

String filter:
StringFilter

Numeric filter:
NumericFilter

Date filter:
DateFilter

Boolean filter:
BooleanFilter

Enum filter:
EnumFilter

Filter with an exploded property:
ExplodedProperty

"is in" filter (unchanged):
ListEditor

DND constant to filter value:
NumericFilter_Constant

DND invalid constant to filter value:
InvalidConstantFilter

Complex filter example:
ComplexFilter

Filter with a property with a long name:
FilterWithLongName


packages/legend-query-builder/src/components/__tests__/QueryBuilderFetchStructure.test.tsx
```diff
@@ -1329,112 +1320,6 @@ test(
   },
 );
 
-test(
```
travisstebbins, line 1332: Moved this test and the below one to `QueryBuilderFilterPanel.test.tsx` since they are tests with the filter panel and not with the fetch structure.

packages/legend-query-builder/src/components/__tests__/QueryBuilderFilterPanel.test.tsx
```diff
@@ -1535,103 +1547,7 @@ test(
...
-      getByText(guaranteeNonNullable(contentNodes[4]), 'Last Name'),
-    );
-    await waitFor(() => getByText(guaranteeNonNullable(contentNodes[4]), 'is'));
-    await waitFor(() =>
-      getByDisplayValue(guaranteeNonNullable(contentNodes[4]), ''),
-    );
-  },
-);
-
-test(
```
travisstebbins, line 1557: This test is a duplicate of the one above so I removed it.

packages/legend-query-builder/src/stores/shared/ValueSpecificationEditorHelper.ts
```diff
@@ -293,21 +301,51 @@ export const generateVariableExpressionMockValue = (
 
 export const getValueSpecificationStringValue = (
   valueSpecification: ValueSpecification,
+  applicationStore: ApplicationStore<
+    LegendApplicationConfig,
+    LegendApplicationPluginManager<LegendApplicationPlugin>
+  >,
+  omitEnumOwnerName?: boolean,
```
YannanGao-gs, line 308: do you think if it's better to change this input to be something like 
```
option: {
omitEnumOwnerName?: boolean,
} 
```

packages/legend-query-builder/src/components/filter/QueryBuilderFilterPanel.tsx
```diff
@@ -872,6 +885,7 @@ const QueryBuilderFilterConditionEditor = observer(
                     isConstant={queryBuilderState.constantState.isValueSpecConstant(
                       node.condition.value,
                     )}
+                    initializeAsEditable={node.isNewlyAdded}
```
YannanGao-gs, line 888: do you think if it's better to rename `isNewlyAdded` to match `initializeAsEditable`?

packages/legend-query-builder/src/components/shared/BasicValueSpecificationEditor.tsx
```diff
@@ -35,6 +35,7 @@ import {
   DragPreviewLayer,
   CalculateIcon,
   InputWithInlineValidation,
+  type SelectComponent,
```
YannanGao-gs, line 38: if you don't mind, could we please clean imports in this file to move `type` above?

YannanGao-gs (approved):

packages/legend-query-builder/src/stores/shared/ValueSpecificationEditorHelper.ts
travisstebbins: Good idea! I've made that change.

packages/legend-query-builder/src/components/filter/QueryBuilderFilterPanel.tsx
travisstebbins: Hmm, I think `isNewlyAdded` might make more sense, because the property tracks if the node was just added to the filter panel and hasn't been rendered yet. We could use `isNewlyAdded` to control other actions related to a newly added node, not just if it should be initialized as editable, so keeping it more generic with the `isNewlyAdded` name makes more sense to me than making it specific to initializing as editable.

packages/legend-query-builder/src/components/shared/BasicValueSpecificationEditor.tsx
travisstebbins: Done!

YannanGao-gs (approved):

PR #3309, opened 2024-07-01
Result panel "filter by" produces incorrect list

travisstebbins:
## Summary

Ag-grid row virtualization caused the "filter by" feature in the result panel to function incorrectly. If the result panel had a large number of rows, and the user selected all the rows in one column and selected "filter by", we wouldn't add all the values to the filter list. This PR fixes that bug.

- [ ] Test(s) added

### DIFF
--- a/packages/legend-query-builder/src/components/result/tds/QueryBuilderTDSGridResult.tsx
+++ b/packages/legend-query-builder/src/components/result/tds/QueryBuilderTDSGridResult.tsx
@@ -278,15 +278,9 @@ export const QueryBuilderTDSGridResult = observer(
       api: DataGridApi<QueryBuilderTDSRowDataType>,
     ): QueryBuilderTDSResultCellData[] => {
       const selectedRanges: DataGridCellRange[] | null = api.getCellRanges();
-      console.log('selectedRanges:', selectedRanges);
-      const nodes = api.getRenderedNodes();
-      console.log('nodes:', nodes);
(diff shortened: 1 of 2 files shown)


packages/legend-query-builder/src/components/result/tds/QueryBuilderTDSGridResult.tsx
```diff
@@ -278,7 +278,8 @@ export const QueryBuilderTDSGridResult = observer(
       api: DataGridApi<QueryBuilderTDSRowDataType>,
     ): QueryBuilderTDSResultCellData[] => {
       const selectedRanges: DataGridCellRange[] | null = api.getCellRanges();
-      const nodes = api.getRenderedNodes();
```
travisstebbins, line 281: `getRenderedNodes()` only returns the nodes that ag-grid is currently rendering. Due to row virtualization for performance optimization, this function doesn't actually return all the nodes in the grid if there are a large number of nodes. Instead, we need to get all the nodes in the table (not just the rendered ones).

YannanGao-gs (approved):

packages/legend-query-builder/src/components/result/tds/QueryBuilderTDSGridResult.tsx
YannanGao-gs: Thank you

PR #3340, opened 2024-07-11
Allow DND column and explorer property to filter condition

travisstebbins:
## Summary

This change allows users to drag a property from the explorer panel or a column from the fetch structure panel and drop it on a filter condition value. This enables users to create a filter where they can directly compare one model property to another model property, which is not currently possible without manually editing the filter in text mode.

- [x] Test(s) added

DND valid explorer property to filter panel:
DNDValidExplorerNode

DND invalid explorer property to filter panel:
DNVInvalidExplorerNode

DND valid fetch structure column to filter panel:
DNDValidFetchStructureColumn

DND invalid fetch structure column to filter panel:
DNDInvalidFetchStructureColumn

DND derived property from explorer panel to filter panel:
DNDDerivedPropFromExplorer

DND derived property from fetch structure panel to filter panel:
DNDDerivedPropFromFetchStructure

End-to-end example of using derived properties:
DerivedProperty_E2E_Example

Collection (exploded) property:
ExplodedProperty

Editing pure code:
EditingPure

Constants still work:
DNDConstant

Parameters still work:
DNDParameter

Post-filter panel still works:
PostFilter


packages/legend-query-builder/src/stores/filter/QueryBuilderFilterState.ts
```diff
@@ -164,90 +320,84 @@ export class FilterConditionState implements Hashable {
     }
   }
 
-  changeProperty(propertyExpression: AbstractPropertyExpression): void {
```
travisstebbins, line 167: This function is no longer used since we no longer support DND a property on top of the existing filter property to replace it.

packages/legend-query-builder/src/stores/filter/QueryBuilderFilterState.ts
```diff
@@ -965,20 +1128,32 @@ export class QueryBuilderFilterState
...
     );
   }
 
+  isInvalidFilterValue(node: QueryBuilderFilterTreeNodeData): boolean {
+    return (
+      node instanceof QueryBuilderFilterTreeConditionNodeData &&
+      node.condition.rightConditionValue instanceof
+        FilterValueSpecConditionValueState &&
+      node.condition.rightConditionValue.value instanceof InstanceValue &&
+      !isValidInstanceValue(node.condition.rightConditionValue.value)
```
YannanGao-gs, line 1148: should we also check         

 node.condition.rightConditionValue instanceof
            FilterPropertyExpressionStateConditionValueState &&
          !node.condition.rightConditionValue.propertyExpressionState.isValid

?

packages/legend-query-builder/src/stores/filter/QueryBuilderFilterState.ts
```diff
@@ -164,90 +320,86 @@ export class FilterConditionState implements Hashable {
...
-      : undefined;
+  setRightConditionValue(val: FilterConditionValueState | undefined): void {
+    this.rightConditionValue = val;
   }
 
   addExistsLambdaParamNames(val: string): void {
     this.existsLambdaParamNames.push(val);
   }
 
+  buildFromValueSpec(val: ValueSpecification | undefined): void {
```
YannanGao-gs, line 366: do you think if it's better if we could rename this to `buildRightConditionValueFromValueSpec` ?

packages/legend-query-builder/src/stores/filter/QueryBuilderFilterState.ts
```diff
@@ -164,90 +320,86 @@ export class FilterConditionState implements Hashable {
...
+    ) {
+      this.rightConditionValue.setValue(val);
+    } else {
+      this.setRightConditionValue(
+        new FilterValueSpecConditionValueState(this, val),
+      );
+    }
+  }
+
+  buildFromPropertyExpressionState(
```
YannanGao-gs, line 378: do you think if it's better if we could rename this to `buildRightConditionValueFromPropertyExpressionState` ?

packages/legend-query-builder/src/stores/filter/QueryBuilderFilterState.ts
```diff
@@ -988,25 +1163,35 @@ export class QueryBuilderFilterState
             `Derived property parameter value for ${node.condition.propertyExpressionState.title} is missing or invalid`,
           );
         }
+        if (
+          node.condition.rightConditionValue instanceof
+            FilterPropertyExpressionStateConditionValueState &&
+          !node.condition.rightConditionValue.propertyExpressionState.isValid
```
YannanGao-gs, line 1169: could you please correct me if my understanding about `FilterPropertyExpressionStateConditionValueState` is wrong?
e.g. if we drag and drop a node from exploreTree to the right hand side of filter node, it will use `FilterPropertyExpressionStateConditionValueState` ?
if so, may i know why this error message is specific to `Derived property parameter`? 
Should we also add this check to `isInvalidFilterValue()` ?

packages/legend-query-builder/src/stores/filter/QueryBuilderFilterState.ts
travisstebbins: Done!

packages/legend-query-builder/src/stores/filter/QueryBuilderFilterState.ts
travisstebbins: Done!

packages/legend-query-builder/src/stores/filter/QueryBuilderFilterState.ts
travisstebbins: `FilterPropertyExpressionStateConditionValueState` will be used if we DND a node from explorer tree to the right hand side of the filter node OR if we DND a column from the fetch structure panel to the right hand side of the filter node. If the property expression on the right hand side of the filter node is a derived property and it's missing a parameter value, then we need to show an error for that.

I've renamed some of the functions used here to try to make them clearer.
- Renamed `isInvalidFilterValue()` to `isInvalidValueSpecFilterValue()` since it checks if the right hand side is a ValueSpecification and if it's an invalid ValueSpecification
- Added a new `isInvalidFilterPropertyExpressionState()` function to check if the left hand side is an invalid property expression state.
- Added a new `isInvalidPropertyExpressionStateFilterValue()` function to check if the right hand side is a PropertyExpressionState and if it's an invalid PropertyExpressionState.

packages/legend-query-builder/src/stores/filter/QueryBuilderFilterState.ts
```diff
@@ -965,48 +1137,86 @@ export class QueryBuilderFilterState
...
             `Derived property parameter value for ${node.condition.propertyExpressionState.title} is missing or invalid`,
           );
         }
+        if (
+          node.condition.rightConditionValue instanceof
+            FilterPropertyExpressionStateConditionValueState &&
+          this.isInvalidPropertyExpressionStateFilterValue(node)
+        ) {
+          validationIssues.push(
+            `Derived property parameter value for ${node.condition.rightConditionValue.propertyExpressionState.title} is missing or invalid`,
```
YannanGao-gs, line 1201: would you mind explaining again why if the right hand of filter node is a property expression, then it is a `Derived property parameter value` ?

packages/legend-query-builder/src/stores/filter/QueryBuilderFilterState.ts
```diff
@@ -965,48 +1137,86 @@ export class QueryBuilderFilterState
...
           validationIssues.push(
             `Derived property parameter value for ${node.condition.propertyExpressionState.title} is missing or invalid`,
           );
         }
+        if (
+          node.condition.rightConditionValue instanceof
+            FilterPropertyExpressionStateConditionValueState &&
+          this.isInvalidPropertyExpressionStateFilterValue(node)
+        ) {
+          validationIssues.push(
```
YannanGao-gs, line 1200: would you mind explaining again why if the right hand of filter node is a property expression, then it is a Derived property parameter value ?

travisstebbins: It's not always a derived property parameter value, but if the right side is a property expression that has a required parameter that is missing, then `propertyExpressionState.isValid` will be false. You can see the `DND derived property from explorer panel to filter panel` and `DND derived property from fetch structure panel to filter panel` screen recordings for examples of where the right hand side is a property expression that has a missing required parameter (we get the red (...) icon indicating the missing parameter).

YannanGao-gs (approved):

PR #3347, opened 2024-07-15
support default query url params

MauricioUyaguari:
## Summary
- support opening saved queries with defaulted param values such as `/edit/{queryID}?p:{savedQueryParam}={savedQueryParamValue}`

- [ ] Test(s) added

### DIFF
--- a/.changeset/bright-horses-wonder.md
+++ b/.changeset/bright-horses-wonder.md
@@ -1,0 +1,5 @@
+---
+'@finos/legend-application-query': patch
+---
+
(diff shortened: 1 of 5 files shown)


akphi (changes requested):
The main logic looks good to me, just 2 things:
- Can we try to leverage the methods in `NetworkUtils` for extracting/injecting query params
- I'm not sure if `URLSearchParams` does the encoding/decoding automatically for you, so you should double check that

akphi (approved):

PR #3350, opened 2024-07-16
datacube: application infrastructure improvements

akphi:


MauricioUyaguari:
@akphi  as mentioned, lets split bump dependencies into its own PR so the team is aware of what has changed. makes it hard if it is buried under 146 changes of repl. those changes affect the rest of the application and not just repl.

PR #3365, opened 2024-07-22
extensibility updates to legend-vscode-extension-dependencies

sameersaini:

- [ ] Test(s) added

### DIFF
--- a/.changeset/six-rocks-wonder.md
+++ b/.changeset/six-rocks-wonder.md
@@ -1,0 +1,5 @@
+---
+'@finos/legend-vscode-extension-dependencies': minor
+---
+
+updates for @finos/legend-vscode-extension-dependencies
--- a/packages/legend-vscode-extension-dependencies/package.json
+++ b/packages/legend-vscode-extension-dependencies/package.json
@@ -17,6 +17,9 @@
     ".": {
       "import": "./lib/index.js",
       "require": "./lib/bundles/bundle.cjs.js"
+    },
+    "./style/index.css": {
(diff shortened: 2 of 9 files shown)


packages/legend-graph/package.json
```diff
@@ -22,12 +22,16 @@
   "sideEffects": false,
   "type": "module",
   "exports": {
-    ".": "./lib/index.js",
+    ".": {
```
akphi, line 25: I think we should follow the same strategy we have discussed before. Why can't we import anything you need from this module and expose that in `legend-vscode-extension-dependencies`? Why do you need to directly import `@finos/legend-graph`, also if you changed it this way, have you checked if this has any internal impact when we chain the projects together?

sameersaini: This is little bit complicated. Better to explain over screen share. 
The graph compilation code exists  legend studio which takes array of presets. From legend-area, I want to create a cjs bundle of legend-area specific presets. (that bundle contains legend-graph).

The legend-studio cis bundle also contains legend-graph. Both cjs bundles contain legend-graph.  Legend graph has function guranteeType() which checks `instanceof` . Since legend-graph exists in both cis bundles, guranteeType fails. 

So, I have to keep a separate cis bundle of legend-graph which will be common/external to both cjs bundles of legend-area & finos/legend-studio. 

I have tested this approach and it works. I beleive this extra bundling is only required for legend-graph package for compiling the graph.

akphi (approved):

PR #3377, opened 2024-07-25
Move query execution error message into query builder result panel, highlight entitlement errors, show no data warning, and show data rounding warning

YannanGao-gs:
## Summary

- Move query execution error message into query builder result panel
- highlight entitlement errors, show no data warning 
- show data incompleteness warning 
- show data rounding warning

- [ ] Test(s) added

- [x]  move execution error inside data grid

https://github.com/user-attachments/assets/7b6a5479-b3e9-4a8c-be51-3f3a16759b21

Screenshot 2024-07-25 at 1 17 50 PM

Screenshot 2024-07-25 at 3 16 12 PM

- [x]  check entitlement highlight

Screenshot 2024-07-25 at 3 13 25 PM

- [x]  no data warning

Screenshot 2024-07-30 at 4 04 03 PM

- [x]  data incompleteness warning

Screenshot 2024-07-30 at 5 36 15 PM

- [x] data truncate warning 

Screenshot 2024-07-31 at 3 41 42 PM

### DIFF
--- a/packages/legend-query-builder/src/components/result/QueryBuilderResultPanel.tsx
+++ b/packages/legend-query-builder/src/components/result/QueryBuilderResultPanel.tsx
@@ -44,7 +44,6 @@ import {
   ReportIcon,
   CubesLoadingIndicatorIcon,
   CubesLoadingIndicator,
-  InfoCircleIcon,
 } from '@finos/legend-art';
 import { observer } from 'mobx-react-lite';
 import { flowResult } from 'mobx';
@@ -61,7 +60,7 @@ import {
   DEFAULT_TAB_SIZE,
   useApplicationStore,
 } from '@finos/legend-application';
-import { deepClone, prettyDuration } from '@finos/legend-shared';
+import { prettyDuration } from '@finos/legend-shared';
 import { useRef, useState } from 'react';
 import { QueryBuilderTDSState } from '../../stores/fetch-structure/tds/QueryBuilderTDSState.js';
 import { PARAMETER_SUBMIT_ACTION } from '../../stores/shared/LambdaParameterState.js';
@@ -158,7 +157,7 @@ export const QueryBuilderEmptyExecutionResultPanel = observer(
           Query returned no data
         </div>
         <div className="query-builder__result__empty-result-warning__body">
-          If you believe the query should return data, please
+          if you believe the query should return data, please
           <button
             className="query-builder__result__permission-error__button"
             disabled={
@@ -380,32 +379,6 @@ export const QueryBuilderResultPanel = observer(
       resultState.isGeneratingPlan ||
       resultState.pressedRunQuery.isInProgress;
 
-    const resultLimit = Math.min(
-      queryBuilderState.fetchStructureState.implementation instanceof
-        QueryBuilderTDSState &&
-        queryBuilderState.fetchStructureState.implementation
-          .resultSetModifierState.limit
-        ? queryBuilderState.fetchStructureState.implementation
-            .resultSetModifierState.limit
-        : Number.MAX_SAFE_INTEGER,
-      resultState.previewLimit,
-    );
-
-    const truncateExecutionResultIfExceed = (
-      execResult: ExecutionResult,
-    ): ExecutionResult => {
-      if (
-        execResult instanceof TDSExecutionResult &&
-        execResult.result.rows.length > resultLimit
-      ) {
-        const truncatedExecutionResult = deepClone(execResult);
-        truncatedExecutionResult.result.rows =
-          truncatedExecutionResult.result.rows.slice(0, resultLimit);
-        return truncatedExecutionResult;
-      }
-      return execResult;
-    };
-
     const getResultSetDescription = (
       _executionResult: ExecutionResult,
     ): string | undefined => {
@@ -416,9 +389,7 @@ export const QueryBuilderResultPanel = observer(
         : undefined;
       if (_executionResult instanceof TDSExecutionResult) {
         const rowLength = _executionResult.result.rows.length;
-        const truncatedRowLength =
-          rowLength > resultLimit ? resultLimit : rowLength;
-        return `${truncatedRowLength} row(s)${
+        return `${rowLength} row(s)${
           queryDuration ? ` in ${queryDuration}` : ''
         }`;
       }
@@ -575,25 +546,6 @@ export const QueryBuilderResultPanel = observer(
                 </div>
               </div>
             )}
-            {executionResult &&
-              executionResult instanceof TDSExecutionResult &&
-              executionResult.result.rows.length > resultLimit && (
-                <div className="query-builder__result__stale-status">
-                  <div className="query-builder__result__stale-status__icon">
-                    <ExclamationTriangleIcon />
-                  </div>
-                  <div className="query-builder__result__stale-status__label">
-                    Data below is not complete - query produces more rows than
-                    the set grid preview limit
-                  </div>
-                  <div
-                    className="query-builder__result__stale-status__icon"
-                    title={`The preview limit is set to ${resultLimit}. The results in the grid below are being limited by this limit and running query with a higher limit would produce more rows. Export will not apply this limit.`}
-                  >
-                    <InfoCircleIcon />
-                  </div>
-                </div>
-              )}
           </div>
           <div className="panel__header__actions query-builder__result__header__actions">
             {resultState.exportState.isInProgress && (
@@ -807,9 +759,7 @@ export const QueryBuilderResultPanel = observer(
           {executionResult && !isLoading && !resultState.executionError && (
             <div className="query-builder__result__values">
               <QueryBuilderResultValues
-                executionResult={truncateExecutionResultIfExceed(
-                  executionResult,
-                )}
+                executionResult={executionResult}
                 queryBuilderState={queryBuilderState}
               />
             </div>
--- a/packages/legend-query-builder/src/components/result/QueryBuilderResultPanel.tsx
+++ b/packages/legend-query-builder/src/components/result/QueryBuilderResultPanel.tsx
@@ -77,6 +77,60 @@ import { QueryBuilderTDSSimpleGridResult } from './tds/QueryBuilderTDSSimpleGrid
 import { getExecutedSqlFromExecutionResult } from './tds/QueryBuilderTDSResultShared.js';
 import { QueryBuilderTDSGridResult } from './tds/QueryBuilderTDSGridResult.js';
 import type { QueryBuilder_LegendApplicationPlugin_Extension } from '../../stores/QueryBuilder_LegendApplicationPlugin_Extension.js';
+import type { QueryBuilderResultState } from '../../stores/QueryBuilderResultState.js';
+
+const PERMISSION_ERRORS = ['permission denied', 'invalid user id or password'];
+
+export const QueryBuilderExecutionErrorPanel = observer(
+  (props: { resultState: QueryBuilderResultState }) => {
+    const { resultState } = props;
+    const error = resultState.executionError;
+    const errorMessage = error
+      ? resultState.queryBuilderState.applicationStore.notificationService.getErrorMessage(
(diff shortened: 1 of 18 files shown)


packages/legend-query-builder/src/stores/QueryBuilderResultState.ts
```diff
@@ -468,6 +475,7 @@ export class QueryBuilderResultState {
           this.queryBuilderState.applicationStore.telemetryService,
           reportWithState,
         );
+        this.setExecutionError(undefined);
```
MauricioUyaguari, line 478: is this correct ? why are we setting undefined here if on line 460 we set the results. 
Shouldnt the reset of the execution error come at the start of when the "execute" button is clicked ?

packages/legend-query-builder/src/components/result/QueryBuilderResultPanel.tsx
```diff
@@ -638,12 +692,15 @@ export const QueryBuilderResultPanel = observer(
...
             <CubesLoadingIndicatorIcon />
           </CubesLoadingIndicator>
-          {!executionResult && !isLoading && (
+          {!executionResult && !isLoading && !resultState.executionError && (
             <BlankPanelContent>
               Build or load a valid query first
             </BlankPanelContent>
           )}
-          {executionResult && !isLoading && (
+          {!isLoading && (
```
MauricioUyaguari, line 700: does this need a !executionResult as well in the boolean condition?

packages/legend-query-builder/src/components/result/QueryBuilderResultPanel.tsx
```diff
@@ -77,6 +77,60 @@ import { QueryBuilderTDSSimpleGridResult } from './tds/QueryBuilderTDSSimpleGrid
...
+          {isPermissionDeniedError() && (
+            <div className="query-builder__result__permission-error">
+              <div className="query-builder__result__permission-error__header">
+                Entitlement / Authorization error - Please
+              </div>
+              <button
+                className="query-builder__result__permission-error__button"
+                onClick={openCheckEntitlmentsEditor}
+              >
+                Click Here to Check Entitlements
```
MauricioUyaguari, line 115: should we only show this if entitlements view is available. is there anyway we can know if we have ability to check proper entitlements. 
Would be bad user experience if this shows up but doesn't work.

MauricioUyaguari (changes requested):

packages/legend-query-builder/src/stores/QueryBuilderResultState.ts
YannanGao-gs: This is correct because if the query can be executed successfully then there is no error so we should set the ExecutionError to be undefined. 
 `reset of the execution error come at the start of when the "execute" button is clicked` should also work

packages/legend-query-builder/src/components/result/QueryBuilderResultPanel.tsx
YannanGao-gs: executionResult is not reset to be undefined if the second execution failed but the first execution succeeded.
Therefore adding `!executionResult` to the condition will cause error messages not to be displayed.
The new three conditions for showing `blank`, `error message`, `grid data` should have worked well as expected

packages/legend-query-builder/src/components/result/QueryBuilderResultPanel.tsx
YannanGao-gs: I added the same disable condition for `Check Entitlements` here 
However, Studio will never know what engine will return without actually running APIs therefore the result panel can't predict if proper entitlements will be returned.

MauricioUyaguari:
lets update PR with description of everything you have done.

YannanGao-gs:
> lets update PR with description of everything that has been updated with regards to the result grid.

Thanks. the changeset has been updated

packages/legend-query-builder/src/stores/fetch-structure/tds/projection/QueryBuilderProjectionValueSpecificationBuilder.ts
```diff
@@ -128,8 +128,12 @@ const appendResultSetModifier = (
           );
           limit.values = [
             Math.min(
-              resultModifierState.limit ?? Number.MAX_SAFE_INTEGER,
-              options?.overridingLimit ?? Number.MAX_SAFE_INTEGER,
+              resultModifierState.limit
```
MauricioUyaguari, line 131: im concern on how we are doing the overload data here. Could cause us some bugs further down the line. 
We are mixing logic in what a "builder" is suppose to do.
I think we should be editing the input of runQuery function rather than changing this builder function and having a bunch of downstream effects. 
We are causing us to have to change a whole bunch of roundtrip tests due to this with a bunch of comments even in the preview data. 
The builder should not have to know this in my opinion. 
can we move this part to a separate PR.

packages/legend-query-builder/src/components/result/QueryBuilderResultPanel.tsx
```diff
@@ -269,6 +380,32 @@ export const QueryBuilderResultPanel = observer(
...
+    );
+
+    const truncateExecutionResultIfExceed = (
+      execResult: ExecutionResult,
+    ): ExecutionResult => {
+      if (
+        execResult instanceof TDSExecutionResult &&
+        execResult.result.rows.length > resultLimit
+      ) {
+        const truncatedExecutionResult = deepClone(execResult);
```
MauricioUyaguari, line 401: is deepClone needed here ? on every render ?
why can't we save it in the state. Maybe have an ExecutionResultState with executionResult and other flags like has overflow?

packages/legend-query-builder/src/components/result/tds/QueryBuilderTDSSimpleGridResult.tsx
```diff
@@ -47,6 +48,47 @@ import type {
 } from '../../../stores/QueryBuilderResultState.js';
 import { QUERY_BUILDER_TEST_ID } from '../../../__lib__/QueryBuilderTesting.js';
 
+export const getFloatGridColumnCustomHeader = (
+  columnName: string,
+): string => ` <div class="query-builder__result__values__table__custom-header">
+              <div>${columnName}</div>
```
MauricioUyaguari, line 54: i dont see where we actually do the rounding ? 
Shouldnt the CellFormatter need editing here?

MauricioUyaguari (changes requested):
Probably could have been split into 2/3 PRs a bit hard to review with all the different things being added. 

Could we please ensure we add tests where possible. 
I would expect the rounding, overflow of data, and even the error one to be able to tested via mocking of execution result ? please let me know if this is not possible.

packages/legend-query-builder/src/components/result/tds/QueryBuilderTDSSimpleGridResult.tsx
YannanGao-gs: Currently, in grid, we set ` maximumFractionDigits: 4` see in `QueryResultCellRenderer`

```
    const formattedCellValue = (): QueryBuilderTDSResultCellDataType => {
      if (isNumber(cellValue)) {
        return Intl.NumberFormat(DEFAULT_LOCALE, {
          maximumFractionDigits: 4,
        }).format(Number(cellValue));
      } else if (isBoolean(cellValue)) {
        return String(cellValue);
      }
      return cellValue;
    };
```

This is for displaying a warning icon to remind users of abandoning some fraction digits there are more than 4.

MauricioUyaguari (changes requested):
minor comments, 
lets review some of the state when adding executionError. make sure we set/unset and render the error panel only when its set ?

packages/legend-query-builder/src/stores/QueryBuilderResultState.ts
```diff
@@ -479,9 +486,7 @@ export class QueryBuilderResultState {
           LogEvent.create(GRAPH_MANAGER_EVENT.EXECUTION_FAILURE),
           error,
         );
-        this.queryBuilderState.applicationStore.notificationService.notifyError(
-          error,
-        );
+        this.setExecutionError(error);
```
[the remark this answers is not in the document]
MauricioUyaguari: I do see this in the component layer, does it make sense to move to the state ?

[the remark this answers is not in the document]
YannanGao-gs: the execution error is reset to undefined at the beginning of `runQuery` at the moment 
```
    const runQuery = (): void => {
      resultState.setSelectedCells([]);
      resultState.setExecutionError(undefined);
      resultState.pressedRunQuery.inProgress();
      if (
        queryParametersState.parameterStates.length &&
        queryParametersState.parameterStates.find(
          (param) =>
            !queryBuilderState.milestoningState.isMilestoningParameter(
              param.parameter,
            ),
        )
      ) {
```
`this.setExecutionError(error);` is setting the ExecutionError when there is one

packages/legend-query-builder/src/components/result/QueryBuilderResultPanel.tsx
```diff
@@ -638,12 +748,15 @@ export const QueryBuilderResultPanel = observer(
...
           </CubesLoadingIndicator>
-          {!executionResult && !isLoading && (
+          {!executionResult && !isLoading && !resultState.executionError && (
             <BlankPanelContent>
               Build or load a valid query first
             </BlankPanelContent>
           )}
-          {executionResult && !isLoading && (
+          {!isLoading && (
+            <QueryBuilderExecutionErrorPanel resultState={resultState} />
```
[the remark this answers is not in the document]
YannanGao-gs: The condition is added inside the < QueryBuilderExecutionErrorPanel/>

```
    return (
      error && (
        <>
          {isPermissionDeniedError() && (
            <div className="query-builder__result__permission-error">
              <div className="query-builder__result__permission-error__header">
                Entitlement / Authorization error - Please
              </div>
              <button
                className="query-builder__result__permission-error__button"
                disabled={

```

packages/legend-query-builder/src/components/result/QueryBuilderResultPanel.tsx
```diff
@@ -77,6 +77,107 @@ import { QueryBuilderTDSSimpleGridResult } from './tds/QueryBuilderTDSSimpleGrid
...
+import type { QueryBuilderResultState } from '../../stores/QueryBuilderResultState.js';
+
+const PERMISSION_ERRORS = ['permission denied', 'invalid user id or password'];
+
+export const QueryBuilderExecutionErrorPanel = observer(
+  (props: { resultState: QueryBuilderResultState }) => {
+    const { resultState } = props;
+    const queryBuilderState = resultState.queryBuilderState;
+    const error = resultState.executionError;
+    const errorMessage = error
```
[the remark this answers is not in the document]
YannanGao-gs: the component is empty if error is undefined 
```
    return (
      error && (
        <>
          {isPermissionDeniedError() && (
            <div className="query-builder__result__permission-error">
              <div className="query-builder__result__permission-error__header">
                Entitlement / Authorization error - Please
              </div>
              <button
                className="query-builder__result__permission-error__button"
                disabled={
```

packages/legend-query-builder/src/stores/QueryBuilderResultState.ts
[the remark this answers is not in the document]
MauricioUyaguari: does it make more sense to move it to the state ?

packages/legend-query-builder/src/components/result/QueryBuilderResultPanel.tsx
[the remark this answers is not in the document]
MauricioUyaguari: yeah but the error here gets it from the global state as well if not in executionerror which i dont think makes sense. can we remove that. and just add it as a condition to even show `QueryBuilderExecutionErrorPanel`

MauricioUyaguari (approved):

PR #3393, opened 2024-07-30
Show zipkin trace id for query execution

YannanGao-gs:
## Summary

Show zipkin trace id for query execution

`x-b3-traceid` has been added in the CORS exposed header params (CrossOriginFilter.EXPOSED_HEADERS_PARAM), which means traceId is provided in the API response header. 

In studio, we need to get the value from the response header then store in the ExecutionResult and display it in the Result panel.

Screenshot 2024-08-01 at 10 33 00 AM

- [ ] Test(s) added

### DIFF
--- a/packages/legend-application-studio/src/stores/editor/editor-state/element-editor-state/FunctionEditorState.ts
+++ b/packages/legend-application-studio/src/stores/editor/editor-state/element-editor-state/FunctionEditorState.ts
@@ -38,6 +38,9 @@ import { ElementEditorState } from './ElementEditorState.js';
 import {
   type CompilationError,
   type PackageableElement,
+  type ExecutionResult,
+  type RawExecutionPlan,
+  type ExecutionResultWithMetadata,
   GRAPH_MANAGER_EVENT,
   LAMBDA_PIPE,
   ParserError,
@@ -45,8 +48,6 @@ import {
   RawLambda,
   buildSourceInformationSourceId,
   isStubbed_PackageableElement,
-  type ExecutionResult,
-  type RawExecutionPlan,
   reportGraphAnalytics,
   buildLambdaVariableExpressions,
   VariableExpression,
@@ -255,7 +256,7 @@ export class FunctionEditorState extends ElementEditorState {
   executionResult?: ExecutionResult | undefined; // NOTE: stored as lossless JSON string
   executionPlanState: ExecutionPlanState;
   parametersState: FunctionParametersState;
-  funcRunPromise: Promise<ExecutionResult> | undefined = undefined;
+  funcRunPromise: Promise<ExecutionResultWithMetadata> | undefined = undefined;
 
   constructor(editorStore: EditorStore, element: PackageableElement) {
     super(editorStore, element);
@@ -388,7 +389,9 @@ export class FunctionEditorState extends ElementEditorState {
     this.executionResult = executionResult;
   };
 
-  setFuncRunPromise = (promise: Promise<ExecutionResult> | undefined): void => {
+  setFuncRunPromise = (
+    promise: Promise<ExecutionResultWithMetadata> | undefined,
+  ): void => {
     this.funcRunPromise = promise;
   };
 
@@ -536,9 +539,9 @@ export class FunctionEditorState extends ElementEditorState {
         report,
       );
       this.setFuncRunPromise(promise);
-      const result = (yield promise) as ExecutionResult;
+      const result = (yield promise) as ExecutionResultWithMetadata;
       if (this.funcRunPromise === promise) {
-        this.setExecutionResult(result);
+        this.setExecutionResult(result.executionResult);
         this.parametersState.setParameters([]);
         // report
         report.timings =
--- a/packages/legend-graph/src/graph-manager/AbstractPureGraphManager.ts
+++ b/packages/legend-graph/src/graph-manager/AbstractPureGraphManager.ts
@@ -17,6 +17,7 @@
 import type {
   ExecutionResult,
   EXECUTION_SERIALIZATION_FORMAT,
+  ExecutionResultWithMetadata,
 } from './action/execution/ExecutionResult.js';
 import type {
   ServiceRegistrationResult,
@@ -156,6 +157,7 @@ export interface ExecutionOptions {
   serializationFormat?: EXECUTION_SERIALIZATION_FORMAT | undefined;
   parameterValues?: ParameterValue[];
   abortController?: AbortController | undefined;
+  preservedResponseHeadersList?: string[];
 }
 
 export interface ServiceRegistrationOptions {
@@ -503,7 +505,7 @@ export abstract class AbstractPureGraphManager {
     graph: PureModel,
     options?: ExecutionOptions,
     report?: GraphManagerOperationReport,
-  ): Promise<ExecutionResult>;
+  ): Promise<ExecutionResultWithMetadata>;
 
   abstract exportData(
     lambda: RawLambda,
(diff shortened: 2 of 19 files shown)


travisstebbins:
Could you add a recording/gif of the feature? It's hard to tell from the screenshot where that tooltip is getting triggered from.

packages/legend-application-studio/src/stores/editor/editor-state/element-editor-state/FunctionEditorState.ts
```diff
@@ -255,7 +255,9 @@ export class FunctionEditorState extends ElementEditorState {
   executionResult?: ExecutionResult | undefined; // NOTE: stored as lossless JSON string
   executionPlanState: ExecutionPlanState;
   parametersState: FunctionParametersState;
-  funcRunPromise: Promise<ExecutionResult> | undefined = undefined;
+  funcRunPromise:
+    | Promise<{ executionResult: ExecutionResult; executionTraceID?: string }>
```
travisstebbins, line 259: Nit: it looks like in most places we format it as `Id`, so I think this would be better as `executionTraceId`

packages/legend-application-studio/src/stores/editor/editor-state/element-editor-state/FunctionEditorState.ts
```diff
@@ -388,7 +390,11 @@ export class FunctionEditorState extends ElementEditorState {
     this.executionResult = executionResult;
   };
 
-  setFuncRunPromise = (promise: Promise<ExecutionResult> | undefined): void => {
+  setFuncRunPromise = (
+    promise:
```
travisstebbins, line 394: Since we're using this type union so much, would it make sense to define this as it's own type?

packages/legend-graph/src/graph-manager/AbstractPureGraphManager.ts
```diff
@@ -156,6 +156,7 @@ export interface ExecutionOptions {
   serializationFormat?: EXECUTION_SERIALIZATION_FORMAT | undefined;
   parameterValues?: ParameterValue[];
   abortController?: AbortController | undefined;
+  preservedResponseHeaders?: string[];
```
travisstebbins, line 159: Nit: I think this should be
```suggestion
  preserveResponseHeaders?: string[];
```

packages/legend-graph/src/graph-manager/protocol/pure/v1/engine/V1_Engine.ts
```diff
@@ -698,17 +702,27 @@ export class V1_Engine {
...
         returnUndefOnError(() =>
           this.parseExecutionResults(executionResultInText, options),
         ) ?? executionResultInText;
-      return V1_serializeExecutionResult(rawExecutionResult);
+      const executionResult = V1_serializeExecutionResult(rawExecutionResult);
+      const traceID = executionResultMap.get(V1_ZIPKIN_TRACE_HEADER);
+      if (traceID) {
+        return { executionResult: executionResult, executionTraceID: traceID };
+      }
+      return { executionResult: executionResult };
```
travisstebbins, line 725: Nit: don't need to include value separately if it's the same as the property name:
```suggestion
        return { executionResult, executionTraceID: traceID };
      }
      return { executionResult };
```

packages/legend-graph/src/graph-manager/protocol/pure/v1/engine/V1_Engine.ts
```diff
@@ -752,17 +766,25 @@ export class V1_Engine {
   async runQueryAndReturnString(
```
travisstebbins, line 766: Maybe since this is no longer returning a string, we should rename it?
```suggestion
  async runQueryAndReturnMap(
```

packages/legend-graph/src/graph-manager/protocol/pure/v1/engine/execution/V1_ExecutionResult.ts
```diff
@@ -83,6 +87,7 @@ export class V1_JsonExecutionResult extends V1_ExecutionResult {
       _type: primitive(),
       builder: usingModelSchema(V1_ResultBuilder.builderSerialization.schema),
       values: raw(),
+      traceId: optional(primitive()),
```
travisstebbins, line 90: this is `traceId` but above on line 70 it's `traceID`. I think we should standardize it to all be `traceId`.

packages/legend-query-builder/src/components/__tests__/QueryBuilderRunQuery.test.tsx
```diff
@@ -71,8 +72,10 @@ test(
     const param = renderResult.getByRole('dialog');
     const graphManager = queryBuilderState.graphManagerState.graphManager;
     const pureManager = guaranteeType(graphManager, V1_PureGraphManager);
+    const executionResultMap = new Map<string, string>();
+    executionResultMap.set(V1_EXECUTION_RESULT, mocked);
     createSpy(pureManager.engine, 'runQueryAndReturnString').mockResolvedValue(
-      mocked,
+      executionResultMap,
```
travisstebbins, line 78: Why not just do the same thing here as in your other tests?
```suggestion
      { executionResult: mocked, }
```

packages/legend-query-builder/src/components/result/QueryBuilderResultPanel.tsx
```diff
@@ -419,13 +421,54 @@ export const QueryBuilderResultPanel = observer(
...
+                <QueryBuilderBaseInfoTooltip
+                  title="Execution Result Analytics"
+                  data={[
+                    {
+                      label: 'Trace',
+                      value: 'available here',
+                      actionButton: (
+                        <div className="query-builder__tooltip__item__action">
+                          <button
+                            onClick={() => {
```
travisstebbins, line 440: Should we disable this button and add a title if there's no base url configured?
```suggestion
                            disabled={!queryBuilderState.config?.zipkinTraceBaseURL}
                            title={queryBuilderState.config?.zipkinTraceBaseURL ? '' : 'No zipkin trace URL configured'}
                            onClick={() => {
```

packages/legend-query-builder/style/_query-builder.scss
```diff
@@ -527,9 +527,13 @@
 
   &__item__action {
     margin-left: 0.5rem;
+    display: flex;
```
travisstebbins, line 530: Can you make sure this doesn't break the styling for the "Show in tree" button on the fetch structure column info tooltip?

YannanGao-gs: Screenshot 2024-08-02 at 4 22 34 PM

It will help align this icon too

packages/legend-query-builder/src/components/__tests__/QueryBuilderRunQuery.test.tsx
YannanGao-gs: This function returns a Map not a composite object, therefore I created a Map first

packages/legend-graph/src/graph-manager/protocol/pure/v1/engine/execution/V1_ExecutionResult.ts
YannanGao-gs: Sry! forgot to clean these code as well. This is not needed as I didn't store traceId inside ExecutionResult anymore.

packages/legend-graph/src/graph-manager/action/execution/ExecutionResult.ts
```diff
@@ -74,6 +74,11 @@ export abstract class ExecutionResult {
   activities: ExecutionActivities[] | undefined;
 }
 
+export type ExecutionResultWithPreservedResponseHeadersList = {
```
akphi, line 77:
```suggestion
export type ExecutionResultWithMetadata = {
```

akphi (dismissed):

MauricioUyaguari (approved):

PR #3403, opened 2024-08-06
Add data incomplete warning

YannanGao-gs:
## Summary

Add data incomplete warning

- [x] Test(s) added

Screenshot 2024-08-06 at 9 39 54 AM

### DIFF
--- a/packages/legend-query-builder/src/components/result/QueryBuilderResultPanel.tsx
+++ b/packages/legend-query-builder/src/components/result/QueryBuilderResultPanel.tsx
@@ -552,7 +552,7 @@ export const QueryBuilderResultPanel = observer(
             )}
             {executionResult &&
               executionResult instanceof TDSExecutionResult &&
-              executionResult.result.rows.length > resultLimit && (
+              resultState.isExecutionResultOverflowing && (
                 <div className="query-builder__result__stale-status">
                   <div className="query-builder__result__stale-status__icon">
                     <ExclamationTriangleIcon />
@@ -785,9 +785,7 @@ export const QueryBuilderResultPanel = observer(
           {executionResult && !isLoading && !resultState.executionError && (
             <div className="query-builder__result__values">
               <QueryBuilderResultValues
-                executionResult={resultState.getTruncatedExecutionResultIfExceed(
-                  executionResult,
-                )}
+                executionResult={executionResult}
                 queryBuilderState={queryBuilderState}
               />
             </div>
--- a/packages/legend-query-builder/src/components/result/QueryBuilderResultPanel.tsx
+++ b/packages/legend-query-builder/src/components/result/QueryBuilderResultPanel.tsx
@@ -390,10 +390,7 @@ export const QueryBuilderResultPanel = observer(
           })
         : undefined;
       if (_executionResult instanceof TDSExecutionResult) {
-        const rowLength = _executionResult.result.rows.length;
-        const truncatedRowLength =
-          rowLength > resultLimit ? resultLimit : rowLength;
-        return `${truncatedRowLength} row(s)${
+        return `${_executionResult.result.rows.length} row(s)${
           queryDuration ? ` in ${queryDuration}` : ''
         }`;
       }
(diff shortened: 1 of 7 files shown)


packages/legend-query-builder/src/stores/QueryBuilderResultState.ts
```diff
@@ -238,6 +241,34 @@ export class QueryBuilderResultState {
...
+
+  getTruncatedExecutionResultIfExceed = (
+    executionResult: ExecutionResult,
+  ): ExecutionResult => {
+    const resultLimit = this.getExecutionResultLimit();
+    if (
+      executionResult instanceof TDSExecutionResult &&
+      executionResult.result.rows.length > resultLimit
+    ) {
+      const truncatedExecutionResult = deepClone(executionResult);
```
MauricioUyaguari, line 264: lets avoid the deepClone here.

packages/legend-query-builder/src/stores/QueryBuilderResultState.ts
```diff
@@ -238,6 +241,34 @@ export class QueryBuilderResultState {
...
+        QueryBuilderTDSState &&
+        this.queryBuilderState.fetchStructureState.implementation
+          .resultSetModifierState.limit
+        ? this.queryBuilderState.fetchStructureState.implementation
+            .resultSetModifierState.limit
+        : Number.MAX_SAFE_INTEGER,
+      this.previewLimit,
+    );
+
+  getTruncatedExecutionResultIfExceed = (
```
MauricioUyaguari, line 256: Don't really want to have to call this upon each render and having to do deep clone. 

Can't we all handle this in the `runQuery` function ?
maybe replace executionResult with 
`executionResultState?: ExecutionResultState`

```
ExecutionResultState
{
  result: ExecutionResult:
  isOverflowing: boolean (set to false by default)
}

```
Then when you get the result back, if it overflows you set that to true and at the time of setting the executionResult you remove the result from the orginal object (no deep clone). as you should never need that value again for anything. 
My fear is we use may orginal object for other things causing other issues and there is no reason to deep clone as that extra value is only for that warning purpose. 
Might as well handle this in run query.

MauricioUyaguari (changes requested):
some comments.

packages/legend-query-builder/src/components/result/QueryBuilderResultPanel.tsx
```diff
@@ -389,7 +391,9 @@ export const QueryBuilderResultPanel = observer(
         : undefined;
       if (_executionResult instanceof TDSExecutionResult) {
         const rowLength = _executionResult.result.rows.length;
-        return `${rowLength} row(s)${
+        const truncatedRowLength =
```
MauricioUyaguari, line 394: this shouldnt be needed anymore right? since we remove the extra value in `runQuery` ?

packages/legend-query-builder/src/stores/QueryBuilderResultState.ts
```diff
@@ -239,12 +243,46 @@ export class QueryBuilderResultState {
...
+        QueryBuilderTDSState &&
+        this.queryBuilderState.fetchStructureState.implementation
+          .resultSetModifierState.limit
+        ? this.queryBuilderState.fetchStructureState.implementation
+            .resultSetModifierState.limit
+        : Number.MAX_SAFE_INTEGER,
+      this.previewLimit,
+    );
+
+  processExecutionResult = (
```
MauricioUyaguari, line 268:
````suggestion
processExecutionResult = (result: ExecutionResult): void => {
    this.setIsExecutionResultOverflowing(false);
    if(result instanceof TDSExecutionResult)
    {
       const resultLimit = this.getExecutionResultLimit();
       if(result.result.rows.length > resultLimit)
       {
         this.setIsExecutionResultOverflowing(true);
         result.result.rows = result.result.rows.slice(0, resultLimit);
       }
    }
  };
````

MauricioUyaguari (changes requested):

MauricioUyaguari (approved):

PR #3404, opened 2024-08-06
Allow DND window function column to post-filter panel value

travisstebbins:
## Summary

This PR allows users to DND a window function column to a post-filter condition value.

- [x] Test(s) added

DND window function to post-filter condition value that supports it:
ValidFilterCondition

Trying to DND window function to post-filter condition value that doesn't support it:
InvalidFilterCondition

### DIFF
--- a/packages/legend-query-builder/src/components/__tests__/QueryBuilderPostFilterPanel.test.tsx
+++ b/packages/legend-query-builder/src/components/__tests__/QueryBuilderPostFilterPanel.test.tsx
@@ -1004,6 +1004,11 @@ test(
     await findByText(windowFunctionPanel, 'Age');
     await findByDisplayValue(windowFunctionPanel, 'sum of Age');
 
+    // Verify that window function column can be removed
+    expect(
+      getByTitle(windowFunctionPanel, 'Remove').hasAttribute('disabled'),
+    ).toBe(false);
+
     // DND sum of Age from window function panel to post-filter condition value
     const sumOfAgeDragSource = await findByTitle(
       windowFunctionPanel,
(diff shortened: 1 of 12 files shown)


packages/legend-query-builder/src/components/__tests__/QueryBuilderPostFilterPanel.test.tsx
```diff
@@ -109,16 +114,12 @@ test(
         ),
       );
     });
-    const queryBuilderFilterPanel = await waitFor(() =>
```
travisstebbins, line 112: I replaced all the instances of `await waitFor` with `await findBy` in this test file because it's cleaner, so that's why there's so many changes here.

packages/legend-query-builder/src/components/__tests__/QueryBuilderPostFilterPanel.test.tsx
```diff
@@ -1249,309 +1416,285 @@ test(
...
+      },
+      testClass: 'model::Firm',
+      property: 'Inc Type',
+      propertyType: Enumeration.prototype,
+      valueToSet: 'Corp',
+    },
+  ];
+
+describe(integrationTest('Post-filter default values are properly set'), () => {
+  test.each(POST_FILTER_PANEL_DEFAULT_VALUE_TEST_CASES)(
```
travisstebbins, line 1512: I moved all the "null as default value" tests into this more generic test, except for boolean and list since those are a bit different.

packages/legend-query-builder/src/components/__tests__/QueryBuilderTDSWindowPanel.test.tsx
```diff
@@ -67,10 +159,8 @@ test(
       tdsState.setShowWindowFuncPanel(true);
     });
 
-    await waitFor(() =>
```
travisstebbins, line 70: I replaced all the instances of `await waitFor` with `await findBy` in this test file because it's cleaner, so that's why there's so many changes here.

YannanGao-gs (approved):

PR #3405, opened 2024-08-07
Allow DND filter node to fetch structure panel

travisstebbins:
## Summary

This PR allows a user to DND a filter condition node from the filter panel to the fetch structure panel to create a new TDS column, similar to the existing behavior of DND an explorer tree node to the fetch structure panel.

- [x] Test(s) added

DND regular nodes from filter panel to fetch structure:
RegularNodes

DND exploded property nodes from filter panel to fetch structure:
ExplodedNodes

DND derived property nodes from filter panel to fetch structure:
DerivedProperty

### DIFF
--- a/packages/legend-query-builder/src/stores/filter/QueryBuilderFilterState.ts
+++ b/packages/legend-query-builder/src/stores/filter/QueryBuilderFilterState.ts
@@ -571,6 +571,19 @@ export class QueryBuilderFilterTreeExistsNodeData
   }
 }
 
+export const isExistsNodeChild = (
+  node: QueryBuilderFilterTreeConditionNodeData,
+): boolean => {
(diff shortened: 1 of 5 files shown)


packages/legend-query-builder/src/stores/filter/QueryBuilderFilterState.ts
```diff
@@ -596,17 +630,6 @@ export class QueryBuilderFilterTreeConditionNodeData
     this.isNewlyAdded = val;
   }
 
-  get isExistsNodeChild(): boolean {
```
travisstebbins, line 599: Moved this to be a function instead of a class method to be more memory efficient, so that every instance of the class doesn't have a copy of this function.

YannanGao-gs (approved):

PR #3414, opened 2024-08-12
Improve handling of changing const/param/derivation types

travisstebbins:
## Summary

Improve how we handle when a user changes the type of a constant/parameter/derivation. This PR implements new behavior such that if a user changes the type of a constant/parameter/derivation, if the value is used in the filter or post-filter panels and is no longer compatible, we show it as an error. This ensures users can't change the types of these values and accidentally create an invalid query.

- [x] Test(s) added

Changing const type:
ChangingConstType

Changing param type:
ChangingParamType

Changing derivation type when used as post-filter condition left value:
DerivationLeftSide

Changing derivation type when used as post-filter condition right value:
DerivationRightSide

### DIFF
--- a/packages/legend-query-builder/src/components/__tests__/QueryBuilderFilterPanel.test.tsx
+++ b/packages/legend-query-builder/src/components/__tests__/QueryBuilderFilterPanel.test.tsx
@@ -3092,7 +3092,7 @@ describe(integrationTest('Filter default values are properly set'), () => {
       const tdsProjectionPanel = await renderResult.findByTestId(
         QUERY_BUILDER_TEST_ID.QUERY_BUILDER_TDS_PROJECTION,
       );
-      const explorerPanel = await renderResult.getByTestId(
+      const explorerPanel = await renderResult.findByTestId(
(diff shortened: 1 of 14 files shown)


packages/legend-query-builder/src/components/__tests__/QueryBuilderFilterPanel.test.tsx
```diff
@@ -2699,199 +2694,575 @@ test(
...
+        { exact: false },
+      ),
+    ).not.toBeNull();
+    expect(
+      renderResult.getByRole('button', { name: 'Run Query' }),
+    ).toHaveProperty('disabled', true);
   },
 );
 
+type FilterPanelDefaultValueTestCase = {
```
travisstebbins, line 2969: Combined all the null default value tests into one `test.each` generic test, similar to what I did for the `QueryBuilderPostFilterPanel` null value tests.

YannanGao-gs (approved):

YannanGao-gs (approved):

PR #3419, opened 2024-08-13
Improve property search panel

travisstebbins:
## Summary

This PR makes the following improvements to the property search panel:
- Search result improvements:
  - Don't show circular dependencies (A->B->A) in the search results
  - Prioritize searching by label (keep path as search key but with lower priority)
- Search display improvements:
  - Don't show full path name for properties nested under class node (just show property name)
  - Show star next to exploded properties
  - Alphabetize properties nested under class node
  - Collapse expanded class nodes upon closing search panel
- Filter improvements
  - Add filter to filter in/out one-many (exploded) properties
  - Add filter to filter in/out enumeration type
  - Add "only" button next to type filters to quickly allow filtering by only 1 type

- [x] Test(s) added

Property search no longer returning circular dependencies:
SearchNoCircularDependencies

One-Many filter:
OneManyFilter

Enumeration and "only" type filters:
EnumerationAndOnlyFilters


packages/legend-query-builder/src/stores/QueryBuilderConfig.ts
```diff
@@ -25,12 +25,13 @@ export const DEFAULT_VARIABLE_NAME = 'var';
 export const DEFAULT_CONSTANT_VARIABLE_NAME = 'c_var';
 export const DEFAULT_POST_FILTER_LAMBDA_VARIABLE_NAME = 'row';
 
-export const QUERY_BUILDER_PROPERTY_SEARCH_MAX_DEPTH = 5;
+export const QUERY_BUILDER_PROPERTY_SEARCH_MAX_DEPTH = 10;
```
travisstebbins, line 28: Since we're eliminating circular dependencies now, it should be safe to increase this depth.

MauricioUyaguari (approved):

PR #3466, opened 2024-08-26
Update the service and Snowflake app model to incorporate stereotypes and tagged values

YannanGao-gs:
## Summary

Update the service and Snowflake app model to incorporate stereotypes and tagged values

- [ ] Test(s) added

https://github.com/user-attachments/assets/6ed7ac1e-ae5a-4f71-b6ed-7a9933cc3d46

### DIFF
--- a/packages/legend-graph/src/graph/metamodel/pure/packageableElements/service/Service.ts
+++ b/packages/legend-graph/src/graph/metamodel/pure/packageableElements/service/Service.ts
@@ -65,6 +65,8 @@ export class Service extends PackageableElement implements Hashable, Testable {
       this.test ?? '',
       hashArray(this.tests),
       hashArray(this.postValidations),
+      hashArray(this.taggedValues),
+      hashArray(this.stereotypes.map((val) => val.pointerHashCode)),
     ]);
   }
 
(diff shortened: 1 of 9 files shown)


packages/legend-graph/src/graph/metamodel/pure/packageableElements/service/Service.ts
```diff
@@ -65,6 +65,8 @@ export class Service extends PackageableElement implements Hashable, Testable {
       this.test ?? '',
       hashArray(this.tests),
       hashArray(this.postValidations),
+      hashArray(this.taggedValues),
+      hashArray(this.stereotypes.map((val) => val.pointerHashCode)),
```
MauricioUyaguari, line 69: why do we use `.pointerHashCode` in some places. what is correct here.

MauricioUyaguari (changes requested):
can we update the roundtrip grammar tests with this please

packages/legend-graph/src/graph/metamodel/pure/packageableElements/service/Service.ts
YannanGao-gs: Screenshot 2024-08-26 at 1 36 31 PM
I think this is how m3-stereotypes is calculating hashcode

MauricioUyaguari (approved):

PR #3470, opened 2024-08-26
added DiagramEditor to vscode-extension-deps

sameersaini:

- [ ] Test(s) added


akphi (dismissed):
As mentioned before, I think eventually we should move all this logic to vscode-client side and only keep this module logic to minimal, i.e. just exports.

PR #3499, opened 2024-09-05
Add QueryBuilder to VS Code extension

travisstebbins:

- [ ] Test(s) added

No testing as this PR largely just forwards exports.


packages/legend-graph/src/graph-manager/protocol/pure/v1/V1_PureGraphManager.ts
```diff
@@ -4015,7 +4015,7 @@ export class V1_PureGraphManager extends AbstractPureGraphManager {
     return graphData;
   }
 
-  private elementToProtocol = <T extends V1_PackageableElement>(
+  elementToProtocol = <T extends V1_PackageableElement>(
```
travisstebbins, line 4018: I had to make this function public because the VS code extension uses the v1 protocol for interacting with the language server, and to save the service to a file, we need to convert the `Service` object to a `V1_Service` object and send it to the language server. Since this function already existed in `V1_PureGraphManager`, it seemed simpler to make it public rather than duplicate the logic somewhere else.

MauricioUyaguari (approved):

PR #3508, opened 2024-09-10
Use the DataSpace analytics result to bypass reading the connection when using PMCD to load the minimal graph

YannanGao-gs:
## Summary

Use the DataSpace analytics result to bypass reading the connection when using PMCD to load the minimal graph

- [ ] Test(s) added


packages/legend-extension-dsl-data-space/src/graph-manager/protocol/pure/v1/engine/analytics/V1_DataSpaceAnalysis.ts
```diff
@@ -107,6 +122,9 @@ const V1_dataSpaceExecutionContextAnalysisResultModelSchema = (
     mapping: primitive(),
     name: primitive(),
     title: optional(primitive()),
+    metadata: usingModelSchema(
```
MauricioUyaguari, line 125: think this needs to be updated to `runtimeMetadata`

MauricioUyaguari (changes requested):

MauricioUyaguari (approved):

PR #3533, opened 2024-09-17
Create pre-filter when right clicking simple projection column in grid output

travisstebbins:
## Summary

This change updates the behavior of the "Filter By"/"Filter Out" buttons that appear when a user right clicks a cell/cell range in the result panel. If the column is a simple projection column, we will create a regular (pre) filter node instead of a post-filter node like we do currently. Derivations, aggregations, and exploded properties still create post-filter nodes.

- [x] Test(s) added

Simple projection column:
SimpleProjection

Derived column:
DerivedColumn

Derivation:
Derivation

Exploded column:
ExplodedColumn

Aggregation column:
AggregationColumn

### DIFF
--- a/packages/legend-query-builder/src/stores/filter/QueryBuilderFilterOperator.ts
+++ b/packages/legend-query-builder/src/stores/filter/QueryBuilderFilterOperator.ts
@@ -27,7 +27,7 @@ import type {
 export abstract class QueryBuilderFilterOperator implements Hashable {
   readonly uuid = uuid();
 
-  abstract getLabel(filterConditionState: FilterConditionState): string;
+  abstract getLabel(): string;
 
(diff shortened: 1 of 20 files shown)


packages/legend-query-builder/src/stores/filter/QueryBuilderFilterOperator.ts
```diff
@@ -27,7 +27,7 @@ import type {
 export abstract class QueryBuilderFilterOperator implements Hashable {
   readonly uuid = uuid();
 
-  abstract getLabel(filterConditionState: FilterConditionState): string;
+  abstract getLabel(): string;
```
travisstebbins, line 30: This param was not used in any of the implementations of this function, so I removed it.

YannanGao-gs (approved):

PR #3553, opened 2024-09-24
Allow plugging in different engine implementations into into V1_PureGraphManager

travisstebbins:

- [ ] Test(s) added


packages/legend-vscode-extension-dependencies/src/index.ts
```diff
@@ -91,8 +160,10 @@ export {
   guaranteeNonNullable,
   guaranteeType,
   isBoolean,
+  NetworkClientError,
```
akphi, line 163: are we good with removing the graph manager logic in this module?

packages/legend-graph/src/graph-manager/protocol/pure/v1/engine/V1_GraphManagerEngine.ts
```diff
@@ -0,0 +1,419 @@
...
+  setup: (config: TEMPORARY__EngineSetupConfig) => Promise<void>;
+
+  // ----------------------------------------- Server Client ----------------------------------------
+  /**
+   * NOTE: ideally, we would not want to leak engine server client like this,
+   * since the communication with engine client should only be done in this class
+   * alone. However, we need to expose the client for plugins, tests, and dev tool
+   * configurations.
+   */
+  getCurrentUserId: () => string | undefined;
```
akphi, line 111: I think we should limit these calls, like said, we're abstracting at `engine` level, we could assume the engine does not come with a `server client`. Except for `serverClientGetBaseUrl`, all others are not needed. In fact, they are convenient utilities, we can move those to the server client within each of engine, let's not expose them at interface level like this.

packages/legend-extension-dsl-data-space/src/graph-manager/protocol/pure/v1/V1_DSL_DataSpace_PureGraphManagerExtension.ts
```diff
@@ -180,19 +180,19 @@ export class V1_DSL_DataSpace_PureGraphManagerExtension extends DSL_DataSpace_Pu
...
       const entities = await entitiesRetriever();
       actionState?.setMessage('Analyzing data space...');
-      analysisResult = await engineClient.postWithTracing<
+      analysisResult = await engine.serverClientPostWithTracing<
         PlainObject<V1_DataSpaceAnalysisResult>
       >(
-        engineClient.getTraceData(ANALYZE_DATA_SPACE_TRACE),
-        `${engineClient._pure()}/analytics/dataSpace/render`,
+        engine.serverClientGetTraceData(ANALYZE_DATA_SPACE_TRACE),
+        `${engine.serverClientGetPureBaseUrl()}/analytics/dataSpace/render`,
```
akphi, line 195: Continue the point mentioned in `V1_GraphManagerEngine`, for example, we can just explicitly call 

```ts
getTraceData(..., `${engine.serverClientGetBaseUrl()`/pure/...`)
```

packages/legend-extension-dsl-data-space/src/graph-manager/protocol/pure/v1/V1_DSL_DataSpace_PureGraphManagerExtension.ts
```diff
@@ -180,19 +180,19 @@ export class V1_DSL_DataSpace_PureGraphManagerExtension extends DSL_DataSpace_Pu
     const cacheResult = cacheRetriever
       ? await this.fetchDataSpaceAnalysisFromCache(cacheRetriever, actionState)
       : undefined;
-    const engineClient = this.graphManager.engine.getEngineServerClient();
+    const engine = this.graphManager.engine;
```
akphi, line 183: This is an example of leaky abstraction. Maybe there's something we need to adjust to expose method from engine to do this operation rather than directly calling a client. For example, let's say the engine can be abstracted further to have capabilities pluggable by extensions (i.e. `PureProtocolProcessorPlugin`), then we can "cast" like below

```ts
const engine = V1_DSL_DataSpace_getEngine(this.graphManager.engine);
```

then after this point, engine is aware of capabilities from `DSL_DataSpace`.

---

Having said that, I think that's overkill for this PR, but something to think about to enhance further later.

akphi (changes requested):

packages/legend-vscode-extension-dependencies/src/index.ts
travisstebbins: Yeah, I just removed the graph logic from it since that now lives in the vscode client code.

packages/legend-extension-dsl-data-space/src/graph-manager/protocol/pure/v1/V1_DSL_DataSpace_PureGraphManagerExtension.ts
travisstebbins: Yeah I agree. There's several instances of plugins and some other classes directly using the engine server client, which I think is not the best way to handle things. Ideally I think things should only be interacting with the engine and not interacting with implementation details. For now, perhaps we can allow the plugins to get the engine server client and check that the type is of `V1_EngineServerClient` so it can access the methods needed without adding things to the abstract engine interface.

akphi: yeah I think for now, it's okay to expose client, but let's keep this to minimum, hence, all the `getBaseUrl...` methods can be scraped

packages/legend-graph/src/graph-manager/protocol/pure/v1/engine/V1_GraphManagerEngine.ts
travisstebbins: I removed these all from the interface, and now wherever they are used I added a `guaranteeType(..., V1_RemoteEngine)` so we can check that only the remote engine is used for things that use the underlying server client implementation.

akphi (approved):

PR #3582, opened 2024-10-08
Create pre-filter when right clicking simple projection column in grid output (bugfix)

travisstebbins:
## Summary

This PR re-implements https://github.com/finos/legend-studio/pull/3533 with a small bug fix.

- [x] Test(s) added

Tested manually with `TEMPORARY__enableGridEnterpriseMode` set to `true` (which is where the bug occurred previously).

Simple projection column:
SimpleProjectionColumn

Derived column:
DerivedColumn

Derivation:
Derivation

Exploded column:
ExplodedColumns

Aggregation column: 
AggregationColumn

### DIFF
--- a/packages/legend-query-builder/src/components/result/tds/QueryBuilderTDSGridResult.tsx
+++ b/packages/legend-query-builder/src/components/result/tds/QueryBuilderTDSGridResult.tsx
@@ -562,7 +562,7 @@ export const QueryBuilderTDSGridResult = observer(
                 );
                 if (
                   resultState.mousedOverCell === null &&
-                  isNonNullable(selectedCells?.[0])
+                  isNonNullable(selectedCells[0])
                 ) {
                   resultState.setMouseOverCell(selectedCells[0]);
                 }
--- a/packages/legend-query-builder/src/components/result/tds/QueryBuilderTDSGridResult.tsx
+++ b/packages/legend-query-builder/src/components/result/tds/QueryBuilderTDSGridResult.tsx
@@ -46,7 +46,6 @@ import { DEFAULT_LOCALE } from '../../../graph-manager/QueryBuilderConst.js';
 import {
   assertErrorThrown,
   isBoolean,
-  isNonNullable,
   isNumber,
   isString,
   isValidURL,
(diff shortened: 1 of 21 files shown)


packages/legend-query-builder/src/components/result/tds/QueryBuilderTDSGridResult.tsx
```diff
@@ -549,12 +552,20 @@ export const QueryBuilderTDSGridResult = observer(
...
                 resultState.setSelectedCells([]);
                 selectedCells.forEach((cell) =>
                   resultState.addSelectedCell(cell),
                 );
+                if (
+                  resultState.mousedOverCell === null &&
+                  isNonNullable(selectedCells[0])
+                ) {
+                  resultState.setMouseOverCell(selectedCells[0]);
+                }
```
travisstebbins, line 568: This is the fix for the bug. The `filterByOrOutValues` function call expects `resultState.mousedOverCell` to be set. `resultState.mousedOverCell` gets set on cell click based on `resultState.selectedCells[0]`. Since `resultState.selectedCells[0]` is null on the very first click of the grid, `resultState.mouseOverCell` was getting set to null. This fixes that bug by setting `resultState.mouseOverCell` if it is null when a cell or range is first selected.

travisstebbins: Went with a different fix so resolving this conversation.

packages/legend-query-builder/src/components/result/tds/QueryBuilderTDSResultShared.tsx
```diff
@@ -394,28 +639,87 @@ const filterByOrOutValue = (
...
+  isFilterBy: boolean,
+  _cellData: QueryBuilderTDSResultCellData,
+  data: QueryBuilderTDSResultCellData | null,
+  tdsState: QueryBuilderTDSState,
+): Promise<void> => {
+  const tdsColState = data?.columnName
+    ? getTDSColumnState(tdsState, _cellData.columnName)
+    : _cellData?.columnName
+      ? getTDSColumnState(tdsState, _cellData.columnName)
+      : undefined;
```
travisstebbins, line 671: This is the fix for the bug that was occurring previously. When the enterprise grid is used, the value passed in for `data` is `null` on the first click, which was causing us to not get the `tdsColState` correctly. To fix this, we can use the value passed in for `_cellData` as a fallback, which does exist on the first grid click.

MauricioUyaguari (approved):

PR #3619, opened 2024-10-21
Handle export data download being handled by external source

travisstebbins:
## Summary

This PR is to support the "Export data" function in the Legend VS Code extension. Currently, the

- [ ] Test(s) added

### DIFF
--- a/packages/legend-query-builder/package.json
+++ b/packages/legend-query-builder/package.json
@@ -70,13 +70,15 @@
   "devDependencies": {
     "@finos/legend-dev-utils": "workspace:*",
     "@jest/globals": "29.7.0",
+    "@types/whatwg-fetch": "^0.0.33",
     "cross-env": "7.0.3",
     "eslint": "9.12.0",
     "jest": "29.7.0",
     "npm-run-all": "4.1.5",
     "rimraf": "6.0.1",
     "sass": "1.80.2",
-    "typescript": "5.6.3"
+    "typescript": "5.6.3",
+    "whatwg-fetch": "^3.6.20"
   },
   "peerDependencies": {
     "react": "^18.0.0"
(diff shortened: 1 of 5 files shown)


packages/legend-query-builder/package.json
```diff
@@ -70,13 +70,15 @@
...
+    "@types/whatwg-fetch": "^0.0.33",
     "cross-env": "7.0.3",
     "eslint": "9.12.0",
     "jest": "29.7.0",
     "npm-run-all": "4.1.5",
     "rimraf": "6.0.1",
     "sass": "1.80.2",
-    "typescript": "5.6.3"
+    "typescript": "5.6.3",
+    "whatwg-fetch": "^3.6.20"
```
travisstebbins, line 81: I had to add `whatwg-fetch` as a `devDependency` because without it, the `Response` object cannot be created in `QueryBuilderResultPanel.test.ts`.

packages/legend-query-builder/src/stores/QueryBuilderResultState.ts
```diff
@@ -417,7 +418,17 @@ export class QueryBuilderResultState {
               this.queryBuilderState.graphManagerState,
             ),
           },
+          undefined,
+          contentType,
         )) as Response;
+      if (result.headers.get(V1_DELEGATED_EXPORT_HEADER) === 'true') {
+        if (result.status === 200) {
+          this.exportState.pass();
```
MauricioUyaguari, line 426: how will the users get notified here. is the delegated client in charge of notifying that the download will run in the background ?

travisstebbins: The user will still get the `Export CSV will run in background` notification within query builder (i.e., the VS Code webview). The delegated client (i.e., vs code) will then handle actually saving the file and displaying a message if the download succeeds/fails.

MauricioUyaguari (approved):

PR #3620, opened 2024-10-21
Add setting to disable Virtual Assistant (for Legend VS Code extension)

travisstebbins:
## Summary

This PR adds a property to `AssistantService` called `isDisabled` to allow us to disable the Virtual Assistant in query builder. This will allow us to disable the Virtual Assistant in the VS Code extension, because it is not currently supported.

It also fixes a bug with the `DataAccessOverview` component where the chart components needed to render were getting removed by tree-shaking optimizations and would no longer be available when the QueryBuilder component is rendered in the Legend VS Code extension.

- [x] Test(s) added

Virtual assistant enabled (default);
VirtualAssistantEnabled

Virtual assistant disabled:
VirtualAssistantDisabled

### DIFF
--- a/packages/legend-query-builder/src/components/data-access/DataAccessOverview.tsx
+++ b/packages/legend-query-builder/src/components/data-access/DataAccessOverview.tsx
@@ -41,9 +41,12 @@ import {
   DatasetEntitlementAccessRequestedReport,
   DatasetEntitlementUnsupportedReport,
 } from '@finos/legend-graph';
+import { Chart as ChartJS, DoughnutController, ArcElement } from 'chart.js';
 import { Doughnut } from 'react-chartjs-2';
 import { getNullableFirstEntry } from '@finos/legend-shared';
 import type { QueryBuilder_LegendApplicationPlugin_Extension } from '../../stores/QueryBuilder_LegendApplicationPlugin_Extension.js';
+
+ChartJS.register(DoughnutController, ArcElement);
 
 const DataAccessOverviewChart = observer(
   (props: { dataAccessState: DataAccessState }) => {
--- a/packages/legend-query-builder/src/graph-manager/QueryBuilderConfig.ts
+++ b/packages/legend-query-builder/src/graph-manager/QueryBuilderConfig.ts
@@ -43,6 +43,12 @@ export class QueryBuilderConfig {
    */
   zipkinTraceBaseURL = '';
 
+  /**
+   * This flag is to disable the "edit pure" and "view pure" buttons
+   * in cases where they are not needed (i.e., VS Code extension)
+   */
+  disableEditViewPure = false;
+
   static readonly serialization = new SerializationFactory(
     createModelSchema(QueryBuilderConfig, {
       TEMPORARY__disableQueryBuilderChat: optional(primitive()),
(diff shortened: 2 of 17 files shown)


MauricioUyaguari:
we shouldnt do this on the config server level. this should be as an input to the application framework when you start up the up. 
For vs code, this would be turned off.

travisstebbins:
> we shouldnt do this on the config server level. this should be as an input to the application framework when you start up the up. For vs code, this would be turned off.

Can you explain more what you mean by "config server level"?

packages/legend-query-builder/src/components/data-access/DataAccessOverview.tsx
```diff
@@ -41,10 +41,13 @@ import {
   DatasetEntitlementAccessRequestedReport,
   DatasetEntitlementUnsupportedReport,
 } from '@finos/legend-graph';
+import { Chart as ChartJS, DoughnutController, ArcElement } from 'chart.js';
 import { Doughnut } from 'react-chartjs-2';
 import { getNullableFirstEntry } from '@finos/legend-shared';
 import type { QueryBuilder_LegendApplicationPlugin_Extension } from '../../stores/QueryBuilder_LegendApplicationPlugin_Extension.js';
 
+ChartJS.register(DoughnutController, ArcElement);
```
YannanGao-gs, line 49: Would you mind explaining to me what it is for ? Thanks!

packages/legend-query-builder/src/graph-manager/QueryBuilderConfig.ts
```diff
@@ -33,6 +33,12 @@ export class QueryBuilderConfig {
    */
   TEMPORARY__enableGridEnterpriseMode = false;
 
+  /**
+   * This flag is to disable the "Show Virtual Assistant" button
+   * in Query Builder
+   */
+  TEMPORARY__disableVirtualAssistant = false;
```
YannanGao-gs, line 40: Based on your PR title, I think maybe we can introduce a boolean flag like TEMPORARY__disableInVSCodeExtension = true; then use this flag on buttons we want to hide

packages/legend-query-builder/src/components/data-access/DataAccessOverview.tsx
travisstebbins: It turns out you need to register the components used in the chart or they might get pruned during treeshaking optimizations: https://stackoverflow.com/questions/70098392/react-chartjs-2-with-chartjs-3-error-arc-is-not-a-registered-element

So this change just registers the 2 components (`DoughnutController` and `ArcElement`) so they are always available. Otherwise, when we render this component in the VS Code extension, it throws this error: `Uncaught Error: "arc" is not a registered element.`

packages/legend-query-builder/src/graph-manager/QueryBuilderConfig.ts
travisstebbins: Just spoke with Mauricio and instead of this approach, I will add a `disabled` flag to the `AssistantService` that controls the virtual assistant, and then add a `disable` function to the service so that we can simply disable the virtual assistant via that function when we create the `ApplicationStore` for the VS Code extension. This way we don't need to keep adding configuration specific to VS Code.

PR #3648, opened 2024-11-05
POC: support multiple tree columns

akphi:
https://github.com/user-attachments/assets/1afe8382-12ad-4c69-bfea-6af700081875


akphi:
```
Demo the DataCube tool because we believe this the tool that can be leveraged to solve your problem

- Virtualize the grid so better performance
- Support standard aggregation and drilling down, each drilling down is an async call so in theory it should not freeze the tool like in Excel

Quickly demo the standard flow:

- Explain it’s definitely more realistic to show some real data, but we’re just using toy DB today for demo purpose, quickly explain the dataset **(what are the dimensions and measures)**
- Show how we group-by one dimension (i.e. Country)
- If they are related, or hierarchical in nature, for example Country > Athlete, we can add another v-pivot
- Sort Descending
- Show how we can filter by a country (i.e. United States)
- Show h-pivot on Year: now another way to work with multi dimensions where dimensions might not have hierarchical relationship is with h-pivot
- Now mention their use case, and said we have experiment with something and would like to show them what we think?
```

PR #3780, opened 2025-01-02
filter only root mapped classes in data product classes

MauricioUyaguari:

- [ ] Test(s) added

### DIFF
--- a/packages/legend-extension-dsl-data-space/src/stores/query-builder/DataSpaceQueryBuilderState.ts
+++ b/packages/legend-extension-dsl-data-space/src/stores/query-builder/DataSpaceQueryBuilderState.ts
@@ -55,6 +55,7 @@ import {
   assertErrorThrown,
   filterByType,
   LogEvent,
+  uniq,
 } from '@finos/legend-shared';
 import { action, flow, makeObservable, observable } from 'mobx';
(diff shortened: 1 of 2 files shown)


packages/legend-extension-dsl-data-space/src/stores/query-builder/DataSpaceQueryBuilderState.ts
```diff
@@ -112,12 +113,11 @@ export const resolveUsableDataSpaceClasses = (
...
-      mappingModelCoverageAnalysisResult.mappedEntities.map(
-        (e) => e.info?.classPath,
-      );
-    const uniqueCompatibleClasses = compatibleClassPaths.filter(
-      (val, index) => compatibleClassPaths.indexOf(val) === index,
+    const uniqueCompatibleClasses = uniq(
+      mappingModelCoverageAnalysisResult.mappedEntities
+        // is root entity filters out mapped classes
+        .filter((e) => e.info?.isRootEntity)
+        .map((e) => e.info?.classPath),
```
YannanGao-gs, line 120: will this handle classes mapped via different mappings? e.g. associationmapping

MauricioUyaguari: yes, as association mappings would dictate if a specific property within a class is mapped. that class would be a root entity. we are looking for overall mapped classes not all mapped properties.

YannanGao-gs (approved):

PR #3818, opened 2025-01-16
Adjust css and vscode extension dependencies to support light mode in VSCode QueryBuilder

travisstebbins:
## Summary

This PR adds some additional exports that the VSCode extension needs to add light mode support.

It also adjusts the light mode css so that it does not wrap query-builder related styling in a `.query-editor` selector, since only the `<QueryEditor>` component wraps its content in the `query-editor` class. The basic `<QueryBuilder>` component (used by the VSCode extension) does not wrap its content in the `query-editor` class, so the light mode styling was not being applied correctly.

- [ ] Test(s) added

QueryBuilder still looks fine in dark mode:
QueryBuilderDarkMode


packages/legend-application-query-bootstrap/style/index.scss
```diff
@@ -213,16 +213,6 @@
     border-color: var(--color-red-180);
   }
 
-  .query-builder__variables__variable__label {
```
travisstebbins, line 216: These were removed as they became duplicates after moving all the below lines out from being nested under `.query-editor`.

packages/legend-application-query-bootstrap/style/index.scss
```diff
@@ -1029,1005 +1019,1000 @@
...
 
-    ::-webkit-scrollbar-thumb {
-      height: 5rem;
-      background: var(--color-legacylight-light-grey-300);
-      outline-offset: -0.2rem;
-      border-radius: 0.4rem;
-      outline-color: var(--color-legacylight-light-grey-100);
-      border-color: var(--color-legacylight-light-grey-100);
-    }
+  ::-webkit-scrollbar-thumb {
```
travisstebbins, line 1028: All the below changes are simply moving these css styles to not be nested under `.query-editor` anymore. Instead, they are now just nested directly under `.theme__legacy-light`.

MauricioUyaguari (approved):

PR #3828, opened 2025-01-22
Support open data cube query with parameters that use function

irisyngao:
## Summary

Support open data cube query with parameters that use function

- [ ] Test(s) added

### DIFF
--- a/packages/legend-application-data-cube/src/stores/LegendDataCubeDataCubeEngine.ts
+++ b/packages/legend-application-data-cube/src/stores/LegendDataCubeDataCubeEngine.ts
@@ -18,11 +18,11 @@ import {
   type V1_Lambda,
   type V1_ValueSpecification,
   type V1_EngineServerClient,
-  V1_PureGraphManager,
   type V1_PureModelContext,
   type ExecutionResult,
   type V1_ExecutionResult,
   RelationalExecutionActivities,
+  V1_PureGraphManager,
   TDSExecutionResult,
   V1_ParameterValue,
   V1_PureModelContextPointer,
@@ -44,12 +44,15 @@ import {
   V1_PackageableType,
   V1_deserializeRawValueSpecificationType,
   V1_Protocol,
+  LET_TOKEN,
+  PURE_DATE_FUNCTIONS_NAMESPACE,
 } from '@finos/legend-graph';
 import {
-  _elementPtr,
-  DataCubeEngine,
   type DataCubeSource,
   type CompletionItem,
+  type DataCubeExecutionOptions,
+  _elementPtr,
+  DataCubeEngine,
   _function,
   DataCubeFunction,
   _deserializeLambda,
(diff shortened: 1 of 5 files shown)


packages/legend-application-data-cube/src/stores/LegendDataCubeDataCubeEngine.ts
```diff
@@ -215,6 +196,50 @@ export class LegendDataCubeDataCubeEngine extends DataCubeEngine {
...
+                          value: _defaultPrimitiveTypeValue(type),
+                        };
+                  return paramValue;
+                }
+              }
+              return undefined;
+            }),
+          )
+        ).filter(isNonNullable);
+        source.lambda.body.push(...letsFuncs);
```
MauricioUyaguari, line 238: why are we adding the let to the end of the expressions. 
Won't this break compilation for lambdas using the parameters within the query ?

packages/legend-application-data-cube/src/stores/LegendDataCubeDataCubeEngine.ts
```diff
@@ -342,6 +367,12 @@ export class LegendDataCubeDataCubeEngine extends DataCubeEngine {
       result = await this._runQuery(query, source.model);
     } else if (source instanceof LegendQueryDataCubeSource) {
       query.parameters = source.lambda.parameters;
+      // since let function was added to the end of the lambda body arrary before
+      // we need to get it by index 1 here
+      const letFunction = source.lambda.body.at(1);
```
MauricioUyaguari, line 372: related to the above comment, isnt this suppose to be at the beginning?

MauricioUyaguari (changes requested):

packages/legend-application-data-cube/src/stores/LegendDataCubeDataCubeEngine.ts
YannanGao-gs: as coded below, 
```
      if (letFunction) {
        query.body = [letFunction, ...query.body];
      }
```

it's added to the beginning.

However, i can switch all to the beginning

MauricioUyaguari: Does this assume there will only be one let statement ? 
what if there is 2. 
let x = now();
let y = now();

would this work and capture both values ?

MauricioUyaguari: mhm okay makes sense, the beginning part

MauricioUyaguari: does it make sense to maybe adding another proeprty on LegendQuerySource -> `letParameterValues` and populate that at the beginining ?

travisstebbins:
Closing in favor of https://github.com/finos/legend-studio/pull/3915

PR #3859, opened 2025-01-30
Bundle legend-graph separately from vscode-extension-dependencies

travisstebbins:
## Summary

This PR bundles `@finos/legend-graph` separately from the rest of `@finos/legend-vscode-extension-dependencies` so that when other VS Code extension packages have multiple dependencies that depend on `@finos/legend-graph`, we can use webpack to alias all of them to the same instance of `@finos/legend-graph`. This is necessary because without this, different dependencies will have their own copies of `@finos/legend-graph` which will cause `instanceof` checks to fail, since they will technically be using different copies of the classes in `@finos/legend-graph`.

- [ ] Test(s) added

No testing as just adding rollup config to @finos/legend-graph.


akphi:
@travisstebbins I understand the rationale of this change, but this is not solving the root problem, we have `instanceof` checks for things coming from `@finos/legend-shared` as well, so maybe a while from now, we would need to CJS bundle that module, etc. At the end of the day, the right solution is probably to build CJS for each and every module.

Now, in the particular setup we have, if you need to include some proprietary extensions, then the easiest will be to replicate exactly what we do here in `vscode-extension-dependencies` module, let's say, create a module called `proprietary-vscode-extension-dependencies` that bundles everything like in `vscode-extension-dependencies` plus the proprietary extensions.

travisstebbins:
> @travisstebbins I understand the rationale of this change, but this is not solving the root problem, we have `instanceof` checks for things coming from `@finos/legend-shared` as well, so maybe a while from now, we would need to CJS bundle that module, etc. At the end of the day, the right solution is probably to build CJS for each and every module.
> 
> Now, in the particular setup we have, if you need to include some proprietary extensions, then the easiest will be to replicate exactly what we do here in `vscode-extension-dependencies` module, let's say, create a module called `proprietary-vscode-extension-dependencies` that bundles everything like in `vscode-extension-dependencies` plus the proprietary extensions.

Thanks for the tip! I do think in the longer term it would make sense to have CJS versions of all our packages available. However, in the meantime, I was able to get my proprietary extensions to work by creating a `proprietary-vscode-extension-dependencies` package, which has a `src/index.ts` file that contains `export * from '@finos/legend-vscode-extension-dependencies'` + exporting all the proprietary code, and then bundling that `src/index.ts` file with rollup.

PR #3903, opened 2025-02-20
datacube: enable saving and loading local file source in DataCube grid

gs-gunjan:

- [ ] Test(s) added

### DIFF
--- a/.changeset/little-crabs-exist.md
+++ b/.changeset/little-crabs-exist.md
@@ -1,0 +1,6 @@
+---
+'@finos/legend-application-data-cube': patch
+'@finos/legend-data-cube': patch
+---
+
+datacube: enable saving and loading local file source in DataCube grid
--- a/.changeset/little-crabs-exist.md
+++ b/.changeset/little-crabs-exist.md
@@ -1,6 +1,5 @@
 ---
 '@finos/legend-application-data-cube': patch
-'@finos/legend-data-cube': patch
 ---
 
-datacube: enable saving and loading local file source in DataCube grid
+Enabled saving and loading local file source.
--- a/packages/legend-application-data-cube/src/stores/LegendDataCubeDuckDBEngine.ts
+++ b/packages/legend-application-data-cube/src/stores/LegendDataCubeDuckDBEngine.ts
@@ -46,6 +46,7 @@ export class LegendDataCubeDuckDBEngine {
   // https://duckdb.org/docs/guides/meta/describe.html
   private static readonly COLUMN_NAME = 'column_name';
   private static readonly COLUMN_TYPE = 'column_type';
+  private static readonly TABLE_NAME = 'table_name';
   // Options for creating csv using papa parser: https://www.papaparse.com/docs#config
   private static readonly ESCAPE_CHAR = `'`;
   private static readonly QUOTE_CHAR = `'`;
@@ -190,6 +191,27 @@ export class LegendDataCubeDuckDBEngine {
       dbReference: ref,
       columnNames: tableSpec.map((spec) => spec[0] as string),
     };
+  }
+
+  async clearLocalFileDataIngest() {
+    const connection = await this.database.connect();
+    const tablesResult = await connection.query(`
+      SELECT table_name
+      FROM information_schema.tables
+      WHERE table_schema = '${LegendDataCubeDuckDBEngine.DUCKDB_DEFAULT_SCHEMA_NAME}'
+      AND table_name LIKE '${LegendDataCubeDuckDBEngine.INGEST_TABLE_NAME_PREFIX}%'`); // Filter tables starting with the prefix
+
+    const tableNames = tablesResult
+      .toArray()
+      .map((row) => row[LegendDataCubeDuckDBEngine.TABLE_NAME] as string);
+
+    await Promise.all(
+      tableNames.map((table) =>
+        connection.query(`
+        DROP TABLE IF EXISTS "${LegendDataCubeDuckDBEngine.DUCKDB_DEFAULT_SCHEMA_NAME}.${table}";
+      `),
+      ),
+    );
   }
 
   async runSQLQuery(sql: string) {
--- a/packages/legend-application-data-cube/src/stores/LegendDataCubeDuckDBEngine.ts
+++ b/packages/legend-application-data-cube/src/stores/LegendDataCubeDuckDBEngine.ts
@@ -29,6 +29,7 @@ import {
   assertNonNullable,
   csvStringify,
   guaranteeNonNullable,
+  isNullable,
   UnsupportedOperationError,
   uuid,
 } from '@finos/legend-shared';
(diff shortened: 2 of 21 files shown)


gs-gunjan:
Screenshot 2025-02-21 at 2 21 29 PM
Screenshot 2025-02-21 at 2 21 07 PM
Screenshot 2025-02-21 at 2 19 37 PM

packages/legend-application-data-cube/src/stores/model/LocalFileDataCubeSource.ts
```diff
@@ -42,26 +41,20 @@ export class LocalFileDataCubeSource extends DataCubeSource {
...
-  model!: PlainObject<V1_PureModelContextData>;
-  runtime!: string;
-  db!: string;
-  schema!: string;
-  table!: string;
-  count!: number;
   fileName!: string;
   fileFormat!: LocalFileDataCubeSourceFormat;
+  dbReference!: string;
+  count!: number;
```
akphi, line 47: let's discuss about where you use this `count`

packages/legend-application-data-cube/src/stores/model/LocalFileDataCubeSource.ts
```diff
@@ -42,26 +41,20 @@ export class LocalFileDataCubeSource extends DataCubeSource {
...
 export class RawLocalFileQueryDataCubeSource {
-  model!: PlainObject<V1_PureModelContextData>;
-  runtime!: string;
-  db!: string;
-  schema!: string;
-  table!: string;
-  count!: number;
   fileName!: string;
   fileFormat!: LocalFileDataCubeSourceFormat;
+  dbReference!: string;
```
akphi, line 46: rename this to `_ref`, it's an internal field really.

packages/legend-application-data-cube/src/stores/LegendDataCubeDuckDBEngine.ts
```diff
@@ -45,9 +46,14 @@ export class LegendDataCubeDuckDBEngine {
   // https://duckdb.org/docs/guides/meta/describe.html
   private static readonly COLUMN_NAME = 'column_name';
   private static readonly COLUMN_TYPE = 'column_type';
+  private static readonly TABLE_NAME = 'table_name';
   // Options for creating csv using papa parser: https://www.papaparse.com/docs#config
   private static readonly ESCAPE_CHAR = `'`;
   private static readonly QUOTE_CHAR = `'`;
+  private static dbReferenceMap: Map<
```
akphi, line 53: Don't think we need to make this static. 

Also let's rename.


```suggestion
  private static catalog: Map<
```

packages/legend-application-data-cube/src/stores/builder/source/loader/LegendDataCubeSourceLoaderBuilderState.ts
```diff
@@ -0,0 +1,38 @@
...
+  ) {
+    this._application = application;
+    this._engine = engine;
+  }
+
+  abstract get isValid(): boolean;
+
+  abstract generateSourceData(): Promise<PlainObject>;
+
+  abstract validateSourceData(source: PlainObject | undefined): boolean;
```
akphi, line 37: as commented in `LocalFile...SourceLoader`, the spirit is wrong here, we don't need this method

packages/legend-application-data-cube/src/stores/builder/LegendDataCubeBuilderStore.tsx
```diff
@@ -295,6 +299,27 @@ export class LegendDataCubeBuilderStore {
         const specification = DataCubeSpecification.serialization.fromJson(
           persistentDataCube.content,
         );
+
+        if (
+          !this.saveState.hasSucceeded &&
```
akphi, line 304: why we use `saveState` here??

packages/legend-application-data-cube/src/stores/builder/LegendDataCubeBuilderStore.tsx
```diff
@@ -295,6 +299,27 @@ export class LegendDataCubeBuilderStore {
...
+          this.sourceLoader.isPartialSouce(specification.source._type as string)
+        ) {
+          this.sourceLoader.changeSourceBuilder(
+            specification.source._type as string,
+          );
+          this.sourceLoader.setSource(specification.source);
```
akphi, line 310: In terms of UX and componentization, I think this is not accurate. What we want is not yet another window on top of the current window for people to pick the source, this is more the second stage of the wizard to load query. We have to follow the following flow:

(1) When people load `/datacube/:id`: if we spot a partial source, we could redirect to `/datacube` and open the loader and fast-forward to the wizard step in the loader where people have to provide the source, we need to do some special trick to not get stuck in a loop (e.g. after loading the partial source, user redirected to `/datacube/:id` and hence infinite loop ) <-- need to think what we do here? Or we can leave people on the same page `/datacube/:id` and do something.


---

(2) And the other flow, in `loader`, when people just casually working in DataCube and decide to click `Load DataCube`, the loader can come up and people can select a DataCube with partial source, that's another flow you have to handle

---

It's best to think of a strategy to homogenize these 2 flows and just have (1) uses the logic in (2)

akphi (changes requested):

packages/legend-application-data-cube/src/stores/builder/LegendDataCubeBuilderStore.tsx
gs-gunjan: cause we loadDataCube after we save too, so for local source this would ask user to upload source which we don't technically require

packages/legend-application-data-cube/src/stores/builder/source/loader/LocalFileDataCubeSourceLoaderBuilderState.ts
```diff
@@ -0,0 +1,211 @@
...
+
+  override validateSourceData(source: PlainObject) {
+    const deserializeSource =
+      RawLocalFileQueryDataCubeSource.serialization.fromJson(source);
+    const intersectingColumns = guaranteeNonNullable(
+      this.columnNames?.filter((col) =>
+        deserializeSource.columnNames.includes(col),
+      ),
+    );
+    if (deserializeSource.fileName !== this.fileName) {
```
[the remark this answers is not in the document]
gs-gunjan: We don't load data before validation hence type checking at this point is not possible (I am looking for ways duckDB does allow us to do that

packages/legend-application-data-cube/src/stores/builder/LegendDataCubeSourceLoaderState.tsx
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  changeSourceBuilder(type: string): void {
+    this.sourceLoaderBuilder = this.createSourceLoaderBuilder(type);
+  }
+
+  private createSourceLoaderBuilder(
+    type: string,
+  ): LegendDataCubeSourceLoaderBuilderState {
+    // We can implement this as a switch when
```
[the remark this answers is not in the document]
gs-gunjan: Meaning we can have multiple sources which could be partial, so we can then convert this as switch

packages/legend-application-data-cube/src/stores/model/LocalFileDataCubeSource.ts
gs-gunjan: We don't use it anywhere as of now, not sure if this will be required in the future at all for analytics

.changeset/little-crabs-exist.md
```diff
@@ -0,0 +1,5 @@
+---
+'@finos/legend-application-data-cube': patch
+---
+
+datacube: enable saving and loading local file source in DataCube grid
```
akphi, line 5:
```suggestion
Enabled saving and loading local file source.
```

packages/legend-application-data-cube/src/stores/builder/LegendDataCubeBuilderStore.tsx
```diff
@@ -295,6 +356,15 @@ export class LegendDataCubeBuilderStore {
         const specification = DataCubeSpecification.serialization.fromJson(
           persistentDataCube.content,
         );
+
+        if (this.loader.isPartialSouce(specification.source._type as string)) {
```
akphi, line 360: As discussed, let's make this code path declarative and not dancing around various state for the blocking window state.

gs-gunjan:
Screenshot 2025-03-05 at 1 57 19 PM

PR #3916, opened 2025-03-06
Add basic DataCube UI integration tests

travisstebbins:
## Summary

This PR adds some basic DataCube UI integration tests. For now, it adds:
 - Showing load DataCube window on load
 - Showing DataCube source search window when clicking "New DataCube" button
 - Loading DataCube from a query when a DataCube ID is passed in URL

- [x] Test(s) added


packages/legend-application-data-cube/src/stores/LegendDataCubeBaseStore.ts
```diff
@@ -200,6 +206,7 @@ export class LegendDataCubeBaseStore {
           },
         },
         {
+          engine: this.remoteEngine,
```
travisstebbins, line 209: The `this.graphManager.initialize` call was overwriting the `V1_EngineServerClient` instance that was getting created on line 119 before. To fix this, we create a `V1_RemoteEngine` instance and then pass that to the initialize call here.

packages/legend-data-cube/src/components/view/grid/DataCubeGrid.tsx
```diff
@@ -389,13 +389,18 @@ const DataCubeGridClient = observer((props: { view: DataCubeViewState }) => {
...
+  const isTestEnv = process.env.NODE_ENV === 'test';
+
   return (
     <div className="relative h-[calc(100%_-_20px)] w-full">
       <AgGridReact
         theme="legacy"
         className="data-cube-grid ag-theme-quartz"
-        rowModelType="serverSide"
-        serverSideDatasource={grid.clientDataSource}
+        rowModelType={isTestEnv ? 'clientSide' : 'serverSide'}
```
travisstebbins, line 402: @akphi ag-grid outputs a bunch of error messages in the tests when enterprise modules are included, which causes the tests to fail (because of our script here: https://github.com/finos/legend-studio/blob/master/packages/legend-dev-utils/jest/disallowConsoleError.js). For QueryBuilder, we don't use enterprise modules when running tests. I did the same here for DataCube, but since DataCube has a lot of enterprise-specific configuration, I had to disable that configuration, too, when we run tests. Let me know if you have any concerns.

packages/legend-graph/src/graph-manager/AbstractPureGraphManager.ts
```diff
@@ -122,9 +122,7 @@ import type {
 export interface TEMPORARY__EngineSetupConfig {
   env: string;
   tabSize: number;
-  clientConfig: ServerClientConfig & {
-    queryBaseUrl?: string | undefined;
-  };
+  clientConfig: ServerClientConfig;
```
travisstebbins, line 125: It seemed like most places where this config was being used had
```
clientConfig: ServerConfig & {
  queryBaseUrl?: string | undefined;
}
```
so I updated `ServerConfig` to include `queryBaseUrl?: string | undefined` in its definition. Let me know if any concerns.

scripts/test/jest.config.base.js
```diff
@@ -104,7 +104,9 @@ export const getBaseJestConfig = (isGlobal) => {
   const config = {
     ...baseConfig,
     setupFiles: [
-      '@finos/legend-dev-utils/jest/disallowConsoleError',
+      ...(options?.allowConsoleError
```
travisstebbins, line 107: ag-grid enterprise outputs a bunch of messages to console.error when running without the license set, so I added this option so we can run tests and still permit console.error messages when we need it.

akphi: This is needed so `mobx` and `react` error can be caught since they don't throw, but just `console.error` out. I would say, we can put a `NODE_ENV === 'test'` to not render this grid or modify the condition of the `console.error` mutation to include `NODE_ENV === 'test'` as well, not just 'dev'.

https://github.com/finos/legend-studio/blob/4559d62a8eec7900490ad03df1fe5e399101c7ff/packages/legend-data-cube/src/components/view/grid/DataCubeGrid.tsx#L385

and 

https://github.com/finos/legend-studio/blob/4559d62a8eec7900490ad03df1fe5e399101c7ff/packages/legend-data-cube/src/components/view/grid/DataCubeGrid.tsx#L405

can be changed to `NODE_ENV !== 'production'` I think

travisstebbins: Good call! I didn't realize we were already doing this to suppress the error logs in dev. This makes more sense so I implemented your suggestion.

PR #3936, opened 2025-03-12
Add Database editor with schema, table, and column editing capabilities

MauricioUyaguari:

- [ ] Test(s) added


packages/legend-application-studio/style/index.scss
```diff
@@ -17,6 +17,7 @@
 @forward 'components/not-found-screen';
 @forward 'components/workspace-setup';
 @forward 'components/editor';
+@forward 'components/editor/database-editor';
```
MauricioUyaguari, line 20: this shouldnt be needed. 
You need to add the forward in the file `style\components\_editor.scss` use the runtime-editor.scss as reference how to add a scss style

MauricioUyaguari:
image

Can we fix the styling here. The header `schemas` `tables` and `columns` are not viewable. 
Also the blue on select is not consistent with selection.

PR #3971, opened 2025-04-01
Add V1_BasicValueSpecificationEditor component

travisstebbins:
## Summary

This PR creates a `V1_BasicValueSpecificationEditor` component to allow the user to visually edit `V1_ValueSpecification` objects without having to build the whole graph, which is required for the standard `BasicValueSpecificationEditor`. In order to do this, I have made all the editor components rendered by the `BasicValueSpecificationEditor` component more generic, so that they can accept any type of value as an input as long as it conforms to the interface. This way, the `V1_BasicValueSpecificationEditor` component can make use of all the existing logic but pass in `V1_ValueSpecification` elements and handle updating those values itself.

NOTE: most of the logic within the various components in the BasicValueSpecificationEditor.tsx file has not been change. The biggest changes are:
- How the type-specific components update the value specification (logic is now in the `BasicValueSpecificationEditor` component instead of the type-specific components)
- How the type-specific components check for errors (logic also lives outside the type-specific components now)
- How enum value options are passed (a list of option objects is passed to the enum-specific component instead of the enum-specific component creating it from the Enumeration object)
- How collection editors convert between text and value specifications (logic lives outside the collection component itself now and is instead passed in through callbacks)
- The CustomDatePicker also moves all the logic for constructing Date functions from inside that component to outside the component. It is passed an updateValueSpecification callback that should handle building Date functions from the selected option.

- [x] Test(s) added

String and boolean values:
StringBoolean

String list:
StringList

Number values:
Numbers

Number lists:
NumbersList

Date values:
Dates

Date lists:
DatesList

Enum values:
Enums

Enum lists:
EnumsList

Filter panel editors:
FilterPanel


packages/legend-graph/src/graph-manager/action/changeDetection/V1_ValueSpecificationObserver.ts
```diff
@@ -0,0 +1,346 @@
+/**
```
travisstebbins, line 1: This file was autogenerated by Devin. Need to review to confirm it is correct.

packages/legend-query-builder/src/stores/shared/V1_ValueSpecificationEditorHelper.ts
```diff
@@ -0,0 +1,119 @@
...
+
+export const isValidV1_ValueSpecification = (
+  valueSpecification: V1_ValueSpecification,
+): boolean => {
+  if (valueSpecification instanceof V1_PrimitiveValueSpecification) {
+    const isRequired = valueSpecification.multiplicity.lowerBound >= 1;
+    // required and no values provided. LatestDate doesn't have any values so we skip that check for it.
+    if (
+      isRequired &&
+      (valueSpecification instanceof V1_CBoolean ||
```
travisstebbins, line 101: The V1 value specifications don't allow null values in their types, so this function doesn't do much in its current state, but it could be useful for the future if that changes, or if we want to do other validations on V1 value specifications.

packages/legend-graph/src/graph-manager/action/changeDetection/V1_ValueSpecificationObserver.ts
travisstebbins: Reviewed it myself and fixed the issues from the Devin-generated code.

MauricioUyaguari (approved):

MauricioUyaguari:
a todo here is we eventually want to move applicationStore out of the shared components. I noticed we only use it for logging on error. we should eventually not have those there so they can be more sharable.

packages/legend-query-builder/src/stores/shared/V1_ValueSpecificationEditorHelper.ts
```diff
@@ -0,0 +1,119 @@
...
+import { buildDatePickerOption } from '../../components/shared/CustomDatePickerHelper.js';
+import type {
+  ApplicationStore,
+  LegendApplicationConfig,
+  LegendApplicationPluginManager,
+  LegendApplicationPlugin,
+} from '@finos/legend-application';
+
+export const isPritiveType = (type: string): boolean =>
+  !Object.values(PRIMITIVE_TYPE)
```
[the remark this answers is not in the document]
travisstebbins: I think that function was actually not used anywhere. I added it and then realized we had an existing function doing the same thing. I've removed this instance now.

MauricioUyaguari (approved):

PR #3980, opened 2025-04-08
DataCube parameters support (LegendQueryDataCubeSource)

travisstebbins:
## Summary

This PR adds support for editing parameter values in DataCube when using LegendQueryDataCubeSource type.
- Allows setting query parameters when creating a new DataCube
- Allows editing query parameters when opening an existing DataCube based on a Legend Query that has parameters
- Query parameters get saved/persisted with the raw source
- If underlying query's parameters change, we make sure to use the parameters from the underlying query and only apply the values in the DataCube raw source if the name and type of the parameter match

- [x] Test(s) added

### DIFF
--- a/packages/legend-application-data-cube-bootstrap/style/index.scss
+++ b/packages/legend-application-data-cube-bootstrap/style/index.scss
@@ -14,6 +14,11 @@
  * limitations under the License.
  */
 
-@import url('@finos/legend-art/lib/fonts.css');
-@import url('@finos/legend-data-cube/lib/index.css');
 @import url('@finos/legend-application-data-cube/lib/index.css');
+@import url('@finos/legend-application/lib/index.css');
+@import url('@finos/legend-art/lib/fonts.css');
+@import url('@finos/legend-art/lib/index.css');
+@import url('@finos/legend-art/lib/normalize.css');
+@import url('@finos/legend-data-cube/lib/index.css');
+@import url('@finos/legend-lego/lib/index.css');
+@import url('@finos/legend-query-builder/lib/index.css');
--- a/packages/legend-application-data-cube/src/components/builder/LegendDataCubeBuilder.tsx
+++ b/packages/legend-application-data-cube/src/components/builder/LegendDataCubeBuilder.tsx
@@ -27,7 +27,7 @@ import {
   DataCubeEvent,
   DataCubeTitleBarMenuItems,
 } from '@finos/legend-data-cube';
-import {} from '@finos/legend-art';
+import { clsx } from '@finos/legend-art';
 import {
   useLegendDataCubeBuilderStore,
   withLegendDataCubeBuilderStore,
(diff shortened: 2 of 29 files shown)


packages/legend-application-data-cube-bootstrap/style/index.scss
```diff
@@ -14,6 +14,11 @@
...
 
+@import url('@finos/legend-application-data-cube/lib/index.css');
+@import url('@finos/legend-application/lib/index.css');
 @import url('@finos/legend-art/lib/fonts.css');
+@import url('@finos/legend-art/lib/index.css');
+@import url('@finos/legend-art/lib/normalize.css');
 @import url('@finos/legend-data-cube/lib/index.css');
-@import url('@finos/legend-application-data-cube/lib/index.css');
+@import url('@finos/legend-lego/lib/index.css');
+@import url('@finos/legend-query-builder/lib/index.css');
```
travisstebbins, line 24: Added these additional style imports for the styling of the V1_BasicValueSpecification editor. I think eventually we should consider moving the BasicValueSpecificationEditor components into a different package since they are now used in more than just QueryBuilder.

packages/legend-application-data-cube/src/stores/builder/source/SourceBuilderStateHelper.ts
```diff
@@ -0,0 +1,67 @@
+/**
```
gs-gunjan, line 1: I think we should name it legend query data cube source builder state helper or chuck it as private methods in the above file, cause these methods are not common across sources

packages/legend-application-data-cube/src/stores/builder/LegendDataCubeBuilderStore.tsx
```diff
@@ -109,6 +113,16 @@ export class LegendDataCubeBuilderState {
 
   setSource(val: DataCubeSource | undefined) {
     this.source = val;
+    // Set sourceViewerDisplay height based on length of parameters
```
gs-gunjan, line 116: does it really need to be set here?

packages/legend-application-data-cube/src/components/builder/LegendDataCubeBuilder.tsx
```diff
@@ -41,30 +41,66 @@ import { useEffect } from 'react';
...
-      </FormButton>
-      <FormButton
-        compact={true}
-        className="ml-1.5"
-        disabled={!store.builder?.dataCube}
-        onClick={() => store.saverDisplay.open()}
-      >
-        Save DataCube
-      </FormButton>
+    <div className="flex h-full w-full items-center justify-between">
```
gs-gunjan, line 53: does this need to be here? cause till someone does not load a new query or a query with a load, our source is not populated

packages/legend-application-data-cube/src/stores/builder/LegendDataCubeBuilderStore.tsx
travisstebbins: I'm open to suggestions if there's a place that you think would be better. I tried setting it where we create the display (https://github.com/finos/legend-studio/blob/master/packages/legend-application-data-cube/src/stores/builder/LegendDataCubeBuilderStore.tsx#L228) but since the source isn't defined there yet, we aren't able to set the height based on the length of parameters in the source. This felt like a viable solution since when we set the source, that's when we know how many parameters there are. But I may have missed a better place to put this logic.

packages/legend-application-data-cube/src/components/builder/LegendDataCubeBuilder.tsx
travisstebbins: Sorry, could you clarify what you mean here? Are you asking if the parameter display needs to be here or if the whole component needs to exist? If it's just the parameter display, is there a better place to put the logic?

packages/legend-application-data-cube/src/stores/builder/source/SourceBuilderStateHelper.ts
travisstebbins: Good point! I don't want to make these private functions since they are used in 2 different files, and I don't want to duplicate logic. But I have renamed the file as suggested.

packages/legend-application-data-cube/src/components/builder/LegendDataCubeBuilder.tsx
travisstebbins: As discussed offline, I moved all this logic into a plugin. This keeps LegendDataCubeBuilder cleaner and agnostic of the type of source being used. I added a `builderInnerHeaderRenderer` function to the `LegendDataCubeApplicationPlugin`, and added a new `LegendQueryDataCubeSourceBuilder_DataCubeApplicationPlugin` class that implements that function (https://github.com/finos/legend-studio/pull/3980/files#diff-534e5cf39c1d8116f173c3d43f2b50d346fdf48cb435d267393dcc376f43f3c4R77)

packages/legend-application-data-cube/src/stores/builder/source/LegendQueryDataCubeSourceBuilderStateHelper.ts
```diff
@@ -0,0 +1,67 @@
...
+  const systemEnumeration = systemModel.elements.find(
+    (element) =>
+      element.path === enumerationPath && element instanceof V1_Enumeration,
+  );
+  if (systemEnumeration) {
+    return systemEnumeration as V1_Enumeration;
+  }
+
+  // If not in the system model, fetch the enumeration from the depot server
+  const enumerationElement = (
```
MauricioUyaguari, line 54: One thing to check here, does this api work if the enum is in one of the dependency of the current project ?

MauricioUyaguari (changes requested):
looks good! small comments

packages/legend-query-builder/src/stores/shared/V1_ValueSpecificationEditorHelper.ts
```diff
@@ -86,27 +90,38 @@ export const getV1_ValueSpecificationStringValue = (
...
-    ) {
-      return true;
+        valueSpecification instanceof V1_CStrictDate
+      ) {
+        return true;
+      }
     }
+  } else if (valueSpecification instanceof V1_AppliedProperty && isRequired) {
+    return valueSpecification.property.length > 0;
   } else if (valueSpecification instanceof V1_Collection) {
```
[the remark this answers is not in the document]
travisstebbins: Makes sense! I've updated the code and added a test case to reflect this.

packages/legend-application-data-cube/src/stores/builder/source/LegendQueryDataCubeSourceBuilderStateHelper.ts
MauricioUyaguari: i think  probably best thing is to improve metadata to allow for searching in dependencies as well, maybe a flag on that api to search on dependencies if not found in current project. @gs-gunjan  your thoughts on this?. i think a while ago we needed this/had discussed this.  
otherwise we'd have to do work on the UI
- if not on current project -> fetch all dependencies from project -> in parallel search for that entity in all dependency projects.

MauricioUyaguari: i'll merge this for now for more testing. but most likely this will be needed to truly make enums work

MauricioUyaguari (approved):

packages/legend-application-data-cube/src/stores/builder/source/LegendQueryDataCubeSourceBuilderStateHelper.ts
travisstebbins: Makes sense! I'm actually working right now on a fix. There's an endpoint I can use to search across all entities in dependencies and filter by the classifier path, so I'll make this call that endpoint if the first call fails.

PR #3997, opened 2025-04-11
Fix how DataCube handles enum params

travisstebbins:
## Summary

This PR updates how DataCube handles Legend Query sources that have enum params. Previously, enum params were stored as `V1_AppliedProperty` instances, but now we store them as `V1_EnumValue` value specification instances. This fixes the error being thrown by engine when we executed the query, because engine expects `V1_EnumValue` type for enum params instead of `V1_AppliedProperty`.

- [x] Test(s) added


travisstebbins:
Closing in favor of https://github.com/finos/legend-studio/pull/3998

PR #4186, opened 2025-06-06
Support undefined stores for RelationalDatabaseConnection

kelly-thai:

- [ ] Test(s) added

### DIFF
--- a/packages/legend-application-studio/src/components/editor/editor-group/RuntimeEditor.tsx
+++ b/packages/legend-application-studio/src/components/editor/editor-group/RuntimeEditor.tsx
@@ -128,7 +128,9 @@ const getConnectionTooltipText = (
   } else if (connectionValue instanceof FlatDataConnection) {
     return `Flat-data connection \u2022 Flat-data store ${connectionValue.store.value.path}`;
   } else if (connectionValue instanceof RelationalDatabaseConnection) {
-    return `Relational database connection \u2020 database store ${connectionValue.store.value.path}`;
+    return connectionValue.store
+      ? `Relational database connection \u2020 database store ${connectionValue.store.value.path}`
+      : `Relational database connection`;
   } else if (connectionValue instanceof ModelChainConnection) {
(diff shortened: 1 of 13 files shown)


packages/legend-application-studio/src/components/editor/editor-group/RuntimeEditor.tsx
```diff
@@ -128,7 +128,7 @@ const getConnectionTooltipText = (
   } else if (connectionValue instanceof FlatDataConnection) {
     return `Flat-data connection \u2022 Flat-data store ${connectionValue.store.value.path}`;
   } else if (connectionValue instanceof RelationalDatabaseConnection) {
-    return `Relational database connection \u2020 database store ${connectionValue.store.value.path}`;
+    return `Relational database connection \u2020 database store ${connectionValue.store?.value.path}`;
```
MauricioUyaguari: could we do better here and not include the database store part if there is no store

packages/legend-graph/src/graph-manager/protocol/pure/v1/transformation/pureGraph/from/V1_ConnectionTransformer.ts
```diff
@@ -350,7 +350,9 @@ export const V1_transformRelationalDatabaseConnection = (
   context: V1_GraphTransformerContext,
 ): V1_RelationalDatabaseConnection => {
   const connection = new V1_RelationalDatabaseConnection();
-  connection.store = metamodel.store.valueForSerialization ?? '';
+  if (metamodel.store) {
+    connection.store = metamodel.store.valueForSerialization ?? '';
```
MauricioUyaguari: dont think we need the `?? ''` since it takes undefined now.

MauricioUyaguari (changes requested):
could we add an engine roundtrip where store is not undefined ?

MauricioUyaguari (approved):

PR #4250, opened 2025-07-03
Relation function mapping editor (form mode)

gs-gunjan:

- [ ] Test(s) added

### DIFF
--- a/packages/legend-graph/src/graph-manager/action/changeDetection/DomainObserverHelper.ts
+++ b/packages/legend-graph/src/graph-manager/action/changeDetection/DomainObserverHelper.ts
@@ -526,6 +526,30 @@ export const observe_FunctionTestSuite = skipObservedWithContext(
     return metamodel;
   },
 );
+
+export const observe_ConcreteFunctionDefinitionWithoutTests = skipObserved(
+  (metamodel: ConcreteFunctionDefinition): ConcreteFunctionDefinition => {
+    observe_Abstract_PackageableElement(metamodel);
+
+    makeObservable<ConcreteFunctionDefinition, '_elementHashCode'>(metamodel, {
+      functionName: observable,
+      returnMultiplicity: observable,
+      parameters: observable.shallow, // only observe the list structure, each object itself is not observed
+      expressionSequence: observable.ref, // only observe the reference, the object itself is not observed
+      stereotypes: observable,
+      taggedValues: observable,
+      tests: observable,
+      _elementHashCode: override,
+    });
+
+    metamodel.parameters.forEach(observe_RawVariableExpression);
+    observe_GenericTypeReference(metamodel.returnType);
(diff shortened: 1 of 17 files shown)


packages/legend-graph/src/graph-manager/action/changeDetection/DomainObserverHelper.ts
```diff
@@ -526,6 +526,30 @@ export const observe_FunctionTestSuite = skipObservedWithContext(
     return metamodel;
   },
 );
+
+export const observe_ConcreteFunctionDefinitionWithoutTests = skipObserved(
```
MauricioUyaguari, line 530: what is the reason for needing this separate observe function without tests ?

MauricioUyaguari (commented):
small comment

packages/legend-graph/src/graph-manager/action/changeDetection/DomainObserverHelper.ts
gs-gunjan: The reason I added another one, is because the place I call this in the mapping editor does not have a context, which is being used in the old function. Since I don't need tests, hence added this func

gs-gunjan:
https://github.com/user-attachments/assets/23b6c5bf-2636-4fa9-b6b8-a63b01a7683f

packages/legend-graph/src/graph-manager/action/changeDetection/DomainObserverHelper.ts
travisstebbins: It looks like the function is called in MappingEditorState's `changeClassMappingSourceDriver` function. Shouldn't we be able to access the observer context with `this.editorStore.changeDetectionState.observerContext` in that function?

MauricioUyaguari: maybe im missing something but whats the harm in using reusing the other function. if there are no tests it should still run fine.

gs-gunjan: for it the context is a required parameter, which I don't have at that point

MauricioUyaguari: ohh now i understand. 
yeah we should be able to do 
`this.editorStore.changeDetectionState.observerContext` as @travisstebbins  mentioned above.

MauricioUyaguari (approved):

packages/legend-graph/src/graph-manager/action/changeDetection/DomainObserverHelper.ts
gs-gunjan: thanks! will change this

PR #4275, opened 2025-07-21
Fetch Lakehouse DataProducts from new endpoint

travisstebbins:
## Summary

Instead of fetching DataProducts from metadata and individual ingest servers, fetch all DataProducts from new unified DataProducts endpoint.

- [ ] Test(s) added


packages/legend-application-marketplace/src/__lib__/LegendMarketplaceNavigation.ts
```diff
@@ -33,13 +32,8 @@ export enum LEGEND_MARKETPLACE_SEARCH_RESULTS_QUERY_PARAM_TOKEN {
 }
 
 export type LakehouseDataProductPathParams = {
-  [LEGEND_MARKETPLACE_ROUTE_PATTERN_TOKEN.GAV]: string;
```
MauricioUyaguari, line 36: we should still support this i think. as some data products won't require deployments such as modeled ones ?

MauricioUyaguari (changes requested):

packages/legend-application-marketplace/src/__lib__/LegendMarketplaceNavigation.ts
travisstebbins: Wouldn't modeled data just be a special AP/APG in a DataProduct?

MauricioUyaguari: yeah but there won't be any deployed version of this. so i suspect they would want to open this data product. so i think we should keep the gav option there.

travisstebbins: I've updated the code to support the old style URL still. For now, it will get the DID from artifact generation and then redirect to the new URL, which it generates from the packageable element name + the DID.

PR #4330, opened 2025-08-05
Redesign Marketplace Lakehouse home page

travisstebbins:
## Summary

Redesign Marketplace Lakehouse home page
- Home page is a large landing area with search box and a few highlighted data products
- What was previously the home page has now become the search results page
- Update header styling to add blur effect
- Update header to underline selected tab instead of using background color

- [ ] Test(s) added


packages/legend-application/src/stores/navigation/BrowserNavigator.ts
```diff
@@ -218,28 +218,31 @@ export class BrowserNavigator implements ApplicationNavigator {
...
     T extends Record<string, NavigationLocationParameterValue>,
-  >(replaceUrlSafeBase64Characters: boolean = false): T {
+  >(options?: {
+    replaceUrlSafeBase64Characters?: boolean | undefined;
+    sanitizeParametersInsteadOfUrl?: boolean | undefined;
+  }): T {
     const result: Record<string, NavigationLocationParameterValue> = {};
+    const address = options?.sanitizeParametersInsteadOfUrl
+      ? this.getCurrentAddress()
+      : sanitizeURL(this.getCurrentAddress());
```
travisstebbins, line 228: I had to make this change because a url like `/results?query=AB%26C` would get sanitized to `/results?query=AB&C`, which would then get interpreted as 2 different query parameters, `query='AB'` and `C=undefined`, rather than the correct behavior, which should be 1 query parameter of `query='AB&C'`. Please let me know if there's a better way to go about fixing this behavior.

MauricioUyaguari (approved):

PR #4424, opened 2025-09-16
Initial support on iceberg flow for lakehouse producers

gs-gunjan:

- [ ] Test(s) added

### DIFF
--- a/packages/legend-application-data-cube/src/components/builder/source/LakehouseProducerDataCubeSourceBuilder.tsx
+++ b/packages/legend-application-data-cube/src/components/builder/source/LakehouseProducerDataCubeSourceBuilder.tsx
@@ -31,11 +31,11 @@ export const LakehouseProducerDataCubeSourceBuilder: React.FC<{
 }> = observer(({ sourceBuilder: state }) => {
   const auth = useAuth();
   const store = useLegendDataCubeBuilderStore();
-  const [isIcebergEnabled, setIsIcebergEnabled] = useState(true);
+  const [isIcebergFlowSelected, setIsIcebergFlowSelected] = useState(true);
 
   const toggleSetisIcebergEnabled = () => {
-    setIsIcebergEnabled(!isIcebergEnabled);
-    state.setIcebergEnabled(!isIcebergEnabled);
+    setIsIcebergFlowSelected(!isIcebergFlowSelected);
+    state.setEnableIceberg(!isIcebergFlowSelected);
   };
 
   useEffect(() => {
@@ -89,7 +89,7 @@ export const LakehouseProducerDataCubeSourceBuilder: React.FC<{
             <div className="flex h-5 w-[calc(100%_-_40px)] overflow-x-auto">
               <FormCheckbox
                 label="Use Iceberg"
-                checked={isIcebergEnabled}
+                checked={isIcebergFlowSelected}
                 onChange={toggleSetisIcebergEnabled}
               />
             </div>
@@ -156,7 +156,7 @@ export const LakehouseProducerDataCubeSourceBuilder: React.FC<{
           </div>
         )}
         {state.selectedTable &&
-          (!isIcebergEnabled || !state.icebergEnabled) && (
+          (!isIcebergFlowSelected || !state.icebergEnabled) && (
             <div className="query-setup__wizard__group mt-2">
               <div className="query-setup__wizard__group__title">Warehouse</div>
               <FormTextInput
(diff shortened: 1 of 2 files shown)


packages/legend-application-data-cube/src/stores/model/LakehouseProducerDataCubeSource.ts
```diff
@@ -43,6 +53,8 @@ export class RawLakehouseProducerDataCubeSource {
       warehouse: primitive(),
       ingestServerUrl: primitive(),
       paths: list(primitive()),
+      icebergRef: optional(primitive()),
```
MauricioUyaguari: could do maybe something like 
````

Class IceBergConfig
{
  catalogUrl: string;
  reference: string;
}


class RawLakehouseProducerDataCubeSource {
  ingestDefinitionUrn!: string;
  warehouse!: string;
  ingestServerUrl!: string;
  paths!: string[];
  icebergConfig?: IceBergConfig | undefined;

}

````

File: packages/legend-application-data-cube/src/stores/builder/source/LakehouseProducerDataCubeSourceBuilderState.ts [hunk 2]
```diff
@@ -43,6 +54,11 @@ export class LakehouseProducerDataCubeSourceBuilderState extends LegendDataCubeS
   ingestUrns: string[] = [];
   tables: string[] = [];
   datasetGroup: string | undefined;
+  icebergEnabled: boolean | undefined;
```
MauricioUyaguari, line 64: think we should move all the iceberg configs as well to its own class.

packages/legend-application-data-cube/src/components/builder/source/LakehouseProducerDataCubeSourceBuilder.tsx
```diff
@@ -15,23 +15,43 @@
...
 import { guaranteeNonNullable } from '@finos/legend-shared';
-import { useEffect } from 'react';
+import { useEffect, useState } from 'react';
 
 export const LakehouseProducerDataCubeSourceBuilder: React.FC<{
   sourceBuilder: LakehouseProducerDataCubeSourceBuilderState;
 }> = observer(({ sourceBuilder: state }) => {
   const auth = useAuth();
   const store = useLegendDataCubeBuilderStore();
+  const [isIcebergEnabled, setIsIcebergEnabled] = useState(true);
```
MauricioUyaguari: Wouldnt we only show this option if the ingest env has iceberg enabled. Can we assume that all of the envs have it enabled?

.changeset/beige-aliens-train.md
```diff
@@ -0,0 +1,7 @@
+---
+'@finos/legend-application-data-cube': minor
+'@finos/legend-server-lakehouse': minor
+'@finos/legend-graph': minor
+---
+
+datacube: support iceberg querying flow for lakehouse producers
```
MauricioUyaguari, line 7: Could we add more detail here. 
So this flow i believe is we cache locally after initial query right ? 

This is so we can also add the other scalale long term flow once duckdb fixes their issue.

packages/legend-application-data-cube/src/components/builder/source/LakehouseProducerDataCubeSourceBuilder.tsx
gs-gunjan: we check for two things before making something icebergEnabled, so if an env is icebergEnabled, only then would this flag be used in the front end, otherwise it is going to be defaulted to false

packages/legend-application-data-cube/src/stores/builder/source/LakehouseProducerDataCubeSourceBuilderState.ts
gs-gunjan: this is coming from the API, so gauging from the ingest env

packages/legend-application-data-cube/src/stores/model/LakehouseProducerDataCubeSource.ts
gs-gunjan: done

.changeset/wet-lions-send.md
```diff
@@ -0,0 +1,7 @@
+---
```
MauricioUyaguari: dont think this belongs in this PR. can we remove

packages/legend-application-data-cube/src/components/builder/source/LakehouseProducerDataCubeSourceBuilder.tsx
MauricioUyaguari: but do we still show the option even if `icebergEnabled` is false? 
point being we shouldnt show this check box if the env doesnt support it right ?

gs-gunjan: We don't show it if the environment does not support it , and we don't use the iceberg flow at all for them, this is superficial for the form and for user's to choose, but the other portion is derived from the underlying producer environment

.changeset/wet-lions-send.md
gs-gunjan: rebased

MauricioUyaguari (approved):

PR #4453, opened 2025-09-22
Add Coming Soon pages to Marketplace

kelly-thai:

- [ ] Test(s) added

### DIFF
--- a/packages/legend-application-marketplace/src/pages/DataAPIs/LegendMarketplaceDataAPIs.tsx
+++ b/packages/legend-application-marketplace/src/pages/DataAPIs/LegendMarketplaceDataAPIs.tsx
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2020-present, Goldman Sachs
+ * Copyright (c) 2025-present, Goldman Sachs
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


packages/legend-application-marketplace/src/pages/DataAPIs/LegendMarketplaceDataAPIs.tsx
```diff
@@ -0,0 +1,78 @@
...
+    {
+      icon: <InsightsIcon />,
+      title: 'AI Insights',
+    },
+  ];
+
+  return (
+    <LegendMarketplacePage className="data-api-coming-soon">
+      <Container className="marketplace-lakehouse-search-results__search-container">
+        <LegendMarketplaceSearchBar
```
travisstebbins, line 62: Could we get confirmation that this is intentional? Seems like the other coming soon pages don't have a search box so I'm not sure why this one would need it.

packages/legend-application-marketplace/src/pages/VendorDetails/LegendMarketplaceVendorDetails.tsx
```diff
@@ -48,3 +59,49 @@ export const LegendMarketplaceVendorDetails =
       );
     }),
   );
+
+export const LegendMarketplaceVendorDataComingSoon = observer(() => {
```
travisstebbins, line 63: Do we need this page still? I don't think we have a vendor details page in our header anymore.

kelly-thai: This will show up when you click on the Terminals and Add Ons tab in the header, will rename to make that more apparent

PR #4507, opened 2025-09-30
Filter data products by UI env instead of explicit filters

travisstebbins:
## Summary

Filter data products by UI env instead of explicit filters
- Prod UI only shows prod data products
- Prod-parallel UI only shows prod-parallel data products (with a filter to enable showing dev data products)
- Lower environments show all data products

- [ ] Test(s) added


packages/legend-application-marketplace/src/stores/lakehouse/LegendMarketplaceSearchResultsStore.ts
```diff
@@ -182,40 +93,24 @@ export class LegendMarketplaceSearchResultsStore {
 
   get filterSortProducts(): BaseProductCardState[] | undefined {
```
MauricioUyaguari, line 123: For now, 

I think we still need to have the filter for Modeled Data Product until search is good to go (back and front end). 

In the Prod Relam I would have one filter to include Legacy Data Products or not (defaulted to off). 
In Prod-Parallel Relam I would have one filter to include dev data products (also default to off as more for our dev team to test stuff). 


To make it clear on the state, it may make sense to have another property on base store LegendMarketplaceBaseStore, maybe called MarktplaceBaseEnvState. 


Maybe something like 



```
// could 
class abtract MarktplaceBaseEnvState 
{
  // 'prod' 'prod-parallel' 
  key : string;

  abstract supportsLegacyDataProducts(): boolean;

  abstract supportsDevDataProducts(): boolean;

  astract filterDataProduct(classification: string): boolean;
} 

class ProdMarketplaceEnvState extends MarktplaceBaseEnvState
{

}

class ProdParllelMarketplaceEnvState extends MarktplaceBaseEnvState
{

}

```

Reason is i could see us adding mmore custom logic for prod parell and production in terms of UX and other filtering.

packages/legend-application-marketplace/src/stores/LegendMarketplaceBaseStore.ts
```diff
@@ -278,6 +281,14 @@ export class LegendMarketplaceBaseStore {
     this.showDemoModal = val;
   }
 
+  get isProdEnv(): boolean {
+    return this.applicationStore.config.env === 'prod';
```
MauricioUyaguari: should we use the `env` for this. 

woudl it be worth it to add a new property on the config. 

because if we want to deploy lets say `qa` with prod relam setting for testing, we would still want to do this right? 

this would be something like dataProductRelam?

travisstebbins: Good point. Made this a new config property.

PR #4517, opened 2025-10-07
Support querying ad hoc data products

gs-gunjan:

- [ ] Test(s) added

### DIFF
--- a/packages/legend-application-data-cube/src/stores/model/LakehouseConsumerDataCubeSource.ts
+++ b/packages/legend-application-data-cube/src/stores/model/LakehouseConsumerDataCubeSource.ts
@@ -129,12 +129,10 @@ export class RawLakehouseConsumerDataCubeSource {
           beforeDeserialize: (callback, jsonValue, jsonParentValue) => {
             /** backward compatibility for dpCoordinates */
             if (!jsonValue && jsonParentValue.dpCoordinates) {
-              const dpCoordinates = deserialize(
-                VersionedProjectData.serialization.schema,
-                jsonParentValue.dpCoordinates,
-              );
-              const origin = new RawLakehouseSdlcOrigin();
-              origin.dpCoordinates = dpCoordinates;
+              const origin = {
+                _type: V1_DataProductOriginType.SDLC_DEPLOYMENT,
+                dpCoordinates: jsonParentValue.dpCoordinates,
+              } as PlainObject;
               callback(null, origin);
             } else {
               callback(null, jsonValue);


packages/legend-application-data-cube/src/stores/model/LakehouseConsumerDataCubeSource.ts
```diff
@@ -18,39 +18,41 @@ import { DataCubeSource } from '@finos/legend-data-cube';
...
   paths!: string[];
 }
 
 export class RawLakehouseConsumerDataCubeSource {
-  dpCoordinates!: VersionedProjectData;
+  dpCoordinates?: VersionedProjectData;
   warehouse!: string;
   environment!: string;
   paths!: string[];
+  origin?: string;
```
MauricioUyaguari: mhm im being a pain here but i'd rather model this properly. 


could we instead have origin by abstract and have `adhoc` and  `sdlc` as subtypes with sdlc having required coordinates. 

we would need custom deserialization for existing queries with this raw source but the `LakehouseConsumerDataCubeSource` shouldnt need that

packages/legend-application-data-cube/src/stores/builder/source/LakehouseConsumerDataCubeSourceLoaderState.ts
```diff
@@ -0,0 +1,155 @@
...
+  V1_EntitlementsLakehouseEnvironmentType,
+  type PersistentDataCube,
+} from '@finos/legend-graph';
+import type { LakehouseContractServerClient } from '@finos/legend-server-lakehouse';
+import { action, makeObservable, observable } from 'mobx';
+import { LegendDataCubeSourceBuilderType } from './LegendDataCubeSourceBuilderState.js';
+import { RawLakehouseConsumerDataCubeSource } from '../../model/LakehouseConsumerDataCubeSource.js';
+
+export class LakehouseConsumerDataCubeSourceLoaderState extends LegendDataCubeSourceLoaderState {
+  adhocDPDefinition: string | undefined;
```
MauricioUyaguari: `fullGraphGrammar`

packages/legend-application-data-cube/src/stores/builder/source/LakehouseConsumerDataCubeSourceLoaderState.ts
```diff
@@ -0,0 +1,155 @@
...
+    if (dataProducts.length > 0) {
+      const dataProduct =
+        dataProducts.length === 1
+          ? dataProducts.at(0)
+          : dataProducts
+              .filter(
+                (dp) =>
+                  Boolean(dp.lakehouseEnvironment) &&
+                  dp.lakehouseEnvironment?.type ===
+                    V1_EntitlementsLakehouseEnvironmentType.PRODUCTION,
```
MauricioUyaguari: why are we selecting production here ? should we make sure the right env makes pp or producation in this case ?

MauricioUyaguari (changes requested):

packages/legend-application-data-cube/src/stores/model/LakehouseConsumerDataCubeSource.ts
travisstebbins: We could probably reuse the same `V1_EntitlementsDataProductOrigin` object, which we already have custom logic to serialize/deserialize.

packages/legend-application-data-cube/src/stores/builder/source/LakehouseConsumerDataCubeSourceLoaderState.ts
gs-gunjan: so given we are not giving an options to select production or production parallel, we pick production if either of those 2 are present, if only one env is present we pick just that one

packages/legend-application-data-cube/src/stores/model/LakehouseConsumerDataCubeSource.ts
gs-gunjan: done

packages/legend-application-data-cube/src/components/builder/LegendDataCubeSourceViewer.tsx
```diff
@@ -469,17 +469,18 @@ const LakehouseConsumerSourceViewer = observer(
...
-          generateGAVCoordinates(
-            source.dpCoordinates.groupId,
-            source.dpCoordinates.artifactId,
-            source.dpCoordinates.versionId,
-          ),
-          dataSpacePath,
-        )
-      : undefined;
+    const link =
+      application.config.legendLakehouseUrl && source.dpCoordinates
```
MauricioUyaguari, line 474: if its adhoc, we should be able to use the `DID` and data product path to open marketplace. 
this could be in follow up PRs

packages/legend-application-data-cube/src/stores/model/LakehouseConsumerDataCubeSource.ts
```diff
@@ -15,42 +15,117 @@
...
+      _type: usingConstantValueSchema(V1_DataProductOriginType.SDLC_DEPLOYMENT),
+      dpCoordinates: usingModelSchema(
+        VersionedProjectData.serialization.schema,
+      ),
+    }),
+  );
+}
+
+export class RawLakehouseConsumerDataCubeSource {
+  dpCoordinates?: VersionedProjectData;
```
MauricioUyaguari: we shouldnt need `dpCoordinates` here on the model layer. 
we should handle it with Serialization. 
if that property is defined in the json we should create the origin of type `RawLakehouseSdlcOrigin`

packages/legend-application-data-cube/src/stores/builder/source/LakehouseConsumerDataCubeSourceLoaderState.ts
MauricioUyaguari: But shouldnt we choose the env based on what env the user has selected in the runtime. 
so if its a prod env we choose prod, if its a prod-parallel we chose prod-parallel.

packages/legend-application-data-cube/src/components/builder/LegendDataCubeSourceViewer.tsx
MauricioUyaguari: NOTE: we can leave this as a follow up. doesnt have to be addressed in this PR.

packages/legend-application-data-cube/src/stores/model/LakehouseConsumerDataCubeSource.ts
gs-gunjan: done

packages/legend-application-data-cube/src/components/builder/LegendDataCubeSourceViewer.tsx
gs-gunjan: done

packages/legend-application-data-cube/src/stores/builder/source/LakehouseConsumerDataCubeSourceLoaderState.ts
gs-gunjan: done

packages/legend-application-data-cube/src/components/builder/LegendDataCubeSourceViewer.tsx
MauricioUyaguari: lets support this for adhoc data products in follow up PR please

packages/legend-application-data-cube/src/stores/builder/source/LakehouseConsumerDataCubeSourceBuilderState.ts
```diff
@@ -41,19 +41,32 @@ import {
...
+  RawLakehouseConsumerDataCubeSource,
+  RawLakehouseSdlcOrigin,
+} from '../../model/LakehouseConsumerDataCubeSource.js';
 
 export class LakehouseConsumerDataCubeSourceBuilderState extends LegendDataCubeSourceBuilderState {
   warehouse: string | undefined;
   selectedDataProduct: string | undefined;
   selectedAccessPoint: string | undefined;
   paths: string[] = [];
+  allEnvironments: string[] = [];
```
MauricioUyaguari, line 60: lets keep just `environments` and do a computed attribute for `filteredEnvironments` to filter out things we don't show depending on whether the selected data product has certain attributes (we should only supported realms based on the data products). Additionally lets fetch the full environmentSummaries and not just the url. We should use the `environmentClassification` part of the env to do this filtering as we should try and avoid parsing the url as much as possible. 



its kinda of weird of circular filtering but the way i see it is:

1. environments fetched initially (done). lets change type to `IngestDeploymentServerConfig[]| undefined` undefined to track that we havent finished fetching it yet
2. users selected data product which populates `dataProductDetails`
  - should trigger `filteredEnvironments` computed function to filter out non supported realms
3. users selects env 
   - possibly add `filteredDataProductDetails` which ideally would filter out realms based on the selected env and leave you with one data product
4. acces point dropdown should be enabled when selectedEnv is defined and one data product detail is possible.

gs-gunjan: The logic does this with this change, just need to make it as a computed attribute

PR #5177, opened 2026-05-19
add OIDC token renewal to prevent session expiry errors in using lege…

bojja-gs:
…nd AI LLM service and orchestrator

- [ ] Test(s) added

### DIFF
--- a/packages/legend-application-marketplace/src/application/LegendMarketplaceWebApplication.tsx
+++ b/packages/legend-application-marketplace/src/application/LegendMarketplaceWebApplication.tsx
@@ -21,7 +21,10 @@ import {
   CubesLoadingIndicatorIcon,
   GhostIcon,
 } from '@finos/legend-art';
-import { useApplicationStore } from '@finos/legend-application';
+import {
+  useApplicationStore,
+  LegendTokenSync,
+} from '@finos/legend-application';
 import {
   BrowserEnvironmentProvider,
   Outlet,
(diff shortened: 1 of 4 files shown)


packages/legend-application-marketplace/src/application/LegendMarketplaceWebApplication.tsx
```diff
@@ -463,11 +466,13 @@ export const LegendMarketplaceWebApplication = observer(
...
     return (
       <AuthProvider {...mergedOIDCConfig}>
-        <BrowserEnvironmentProvider baseUrl={baseUrl}>
-          <LegendMarketplaceFrameworkProvider>
-            <LegendMarketplaceWebApplicationRouter />
-          </LegendMarketplaceFrameworkProvider>
-        </BrowserEnvironmentProvider>
+        <LegendTokenSync>
+          <BrowserEnvironmentProvider baseUrl={baseUrl}>
+            <LegendMarketplaceFrameworkProvider>
```
MauricioUyaguari, line 471: If I'm reading the diff correctly, <LegendTokenSync> is being placed between <AuthProvider> and <BrowserEnvironmentProvider>. That will throw at runtime because LegendTokenSync calls useApplicationStore() internally, which requires ApplicationStoreProvider to be an ancestor — and that provider lives inside LegendMarketplaceFrameworkProvider, which is nested below BrowserEnvironmentProvider.

should it be moved?

MauricioUyaguari (changes requested):

packages/legend-application-marketplace/src/application/LegendMarketplaceWebApplication.tsx
bojja-gs: actuallly the placement is correct. ApplicationStoreProvider wraps the entire app at root level in LgendMarketplace.tsx, so when we do LegendTokenSync is already ancestor of that so. 

and to my understanding, legendmarketplaceframeworkprovider is consuming the store via applicationframeworkprovider, so guess here it wont do some sort of provider. 

no need to remove i think. there are some 8 tests which we are calling so all passed.

MauricioUyaguari (approved):

PR #5255, opened 2026-06-11
Add SDLC auth pop up

AFine-gs:
Create a mechanism to recover from lost SDLC auth without resetting local state.  When SDLC gets a 401 - trigger a pop up to generate a new auth and also add a manual SDLC Re-auth button to the task bar.


packages/legend-application-studio/src/stores/LegendStudioBaseStore.ts
```diff
@@ -206,6 +255,314 @@ export class LegendStudioBaseStore {
...
+        if (closedPoll !== undefined) {
+          clearInterval(closedPoll);
+          closedPoll = undefined;
+        }
+        try {
+          popup.close();
+        } catch {
+          // ignored — popup may already be closed or cross-origin
+        }
+        if (!signalled) {
```
MauricioUyaguari: If the SDLC server or IdP sends Cross-Origin-Opener-Policy: same-origin anywhere in the redirect chain, the browser severs window.opener and the callback page's postMessage never reaches us. The user completes auth successfully, the popup closes itself, and the closedPoll lands here — failing the flow and showing an error even though the session cookie actually landed.

Suggestion: when we get here via popup closure (as opposed to popup-blocked), run verifySDLCSessionAfterReAuth() before failing instead of resolving false outright. It costs one GET and makes the flow robust against opener severing, and against any callback page that fails to signal. The signalled flag could become a tri-state ('signalled' | 'closed') so the popup-blocked path can still fail fast.

packages/legend-application-studio/src/stores/LegendStudioBaseStore.ts
```diff
@@ -206,6 +255,314 @@ export class LegendStudioBaseStore {
...
+    if (this.inFlightAutoReAuth) {
+      return this.inFlightAutoReAuth;
+    }
+    // Loop guard: we've already auto-attempted once this episode. Surface
+    // the manual fallback and stop auto-retrying.
+    if (this.autoPopupReAuthAttempted) {
+      this.notifyManualReAuthAvailable();
+      return Promise.resolve(false);
+    }
+    this.autoPopupReAuthAttempted = true;
```
MauricioUyaguari, line 332: If a 401 arrives while a manual popup is already open, this line consumes the episode's one auto-attempt, then reAuthorizeSDLCInPopup() immediately short-circuits on popupReAuthState.isInProgress and returns false — so the user gets the "Automatic SDLC re-authentication did not complete, click the shield icon" warning while the popup is literally on screen.

Suggestion: check this.popupReAuthState.isInProgress before setting the flag, and in that case return (or await) the in-flight manual attempt's outcome instead of consuming the episode and warning.

packages/legend-application-studio/src/stores/LegendStudioBaseStore.ts
```diff
@@ -206,6 +255,314 @@ export class LegendStudioBaseStore {
...
+    // because cookies didn't actually land) MUST NOT recurse into the popup
+    // flow — that would return the in-flight outer promise and deadlock.
+    if (this.suppressAutoReAuth) {
+      return Promise.resolve(false);
+    }
+    // Coalesce concurrent callers.
+    if (this.inFlightAutoReAuth) {
+      return this.inFlightAutoReAuth;
+    }
+    // Loop guard: we've already auto-attempted once this episode. Surface
```
MauricioUyaguari, line 326: Once the single auto-attempt fails (most commonly because window.open from a 401 response handler lacks transient user activation and gets popup-blocked), the episode never resets until a re-auth succeeds. From then on, every failing SDLC request lands in this branch and fires notifyManualReAuthAvailable() — a burst of editor requests after session expiry will stack several identical warning toasts.

Consider deduping this notification (e.g. only notify if the previous warning was more than N seconds ago, or track a manualReAuthNotified flag that resets with the episode).

MauricioUyaguari (changes requested):

MauricioUyaguari (approved):

PR #5312, opened 2026-07-08
adding telemetry for ai suggestion buttons

gs-gunjan:
## Summary
Adding metrics for usage of ai suggestion buttons

- [ ] Test(s) added
- [X] No testing (please provide an explanation)


MauricioUyaguari:
i think we are missing 2 key actions, we should log when they apply the suggestion or ignore it. 
kinda of like the thumbs up thumbs down users can do on chats. 
we can use those actions as suggest or not.

MauricioUyaguari (approved):

PR #5383, opened 2026-07-28
Review token refresh

TharunRajeev:

- [ ] Test(s) added

### DIFF
--- a/.changeset/flat-moons-float.md
+++ b/.changeset/flat-moons-float.md
@@ -1,0 +1,7 @@
+---
+'@finos/legend-extension-dsl-data-product': patch
+'@finos/legend-application-marketplace': patch
+'@finos/legend-application': patch
+---
+
+Fixed marketplace users getting unauthorized errors after their ~5 minute OIDC access token expired. Removed a duplicate silent token renewal in `LegendTokenSync` that raced with the OIDC library's own automatic renewal and could wipe out a valid token, and switched approve/deny, refresh, and request-access action handlers to read the latest token via a ref instead of one captured at render time.
--- a/.changeset/polite-hounds-notice.md
+++ b/.changeset/polite-hounds-notice.md
@@ -1,0 +1,3 @@
+---
+'@finos/legend-application-studio': patch
+---
--- a/packages/legend-application-marketplace/src/application/LegendMarketplaceWebApplication.tsx
+++ b/packages/legend-application-marketplace/src/application/LegendMarketplaceWebApplication.tsx
@@ -43,7 +43,7 @@ import {
   AuthProvider,
   withAuthenticationRequired,
 } from 'react-oidc-context';
-import type { User } from 'oidc-client-ts';
+import { WebStorageStateStore, type User } from 'oidc-client-ts';
 import type { LegendMarketplaceOidcConfig } from './LegendMarketplaceApplicationConfig.js';
 import { MarketplaceLakehouseHeader } from '../components/Header/LegendMarketplaceHeader.js';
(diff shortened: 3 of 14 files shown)


MauricioUyaguari:
The PR really contains three separate fixes, and they're not equally well-chosen:

1. Removing the duplicate renewal in LegendTokenSync — yes, this is the right fix. Two independent code paths calling signinSilent() for the same expiry event is a straightforward bug, and deleting the redundant one in favor of the library's own SilentRenewService (plus addSilentRenewError for cleanup) is exactly what the library authors intend. No reasonable alternative here.

2. The tokenRef pattern for stale closures — right idea, acceptable execution. The alternatives would be keeping the token in useEffect deps (causes refetch storms on every renewal — worse) or plumbing a getToken() callback through the stores (cleaner long-term, but a much bigger refactor). The ref is the pragmatic middle ground; my only quarrel is the 10× duplication instead of one shared hook.

3. The cross-tab layer (localStorage userStore + automaticSilentRenew: false + Web Locks) — this is the questionable part, for two reasons:

Its premise may not hold. The doc comment justifies it by saying tabs "race to redeem the same single-use refresh token" — but that's internally inconsistent with the setup it describes. With the previous config (sessionStorage, per-tab), each tab completed its own silent signin and holds its own refresh token grant; there is no shared token to race over. Tabs only share a refresh token after this PR moves the store to localStorage — the change partly creates the condition it then needs Web Locks to solve. There are IdPs whose refresh-token-rotation reuse detection revokes a whole session family when tabs renew concurrently, which would justify this — but that should be established with evidence (actual invalid_grant logs from multi-tab sessions) before accepting this much machinery.
It's where all the risk lives. Replicating oidc-client-ts's storage key scheme, a hand-rolled 3-second grace period, the loser-tab stale auth.user hole from my review — every defect I found is in this layer, and none of it is covered by tests.
What I'd consider the best route:

Land fixes 1 and 2 essentially as-is (with a shared hook). There's a decent chance these two alone fix the reported bug — the dual-renewal race could plausibly explain the token getting wiped.
Hold the cross-tab layer until the multi-tab race is demonstrated to exist. If it is real, the cleaner design is leader election plus BroadcastChannel (or a storage event) with auth.events.load(user) in the follower tabs — that pushes the renewed user into react-oidc-context properly, so auth.user updates everywhere and the loser-tab staleness problem never exists. Note that MSAL and auth0-spa-js solve this exact problem internally; hand-rolling it against a library that doesn't is inherently going against the grain.
Cheap insurance regardless: a one-shot 401 → signinSilent() → retry interceptor in the server clients. It's timer-independent, covers laptop-sleep and clock-skew cases that expiry events miss, and would have masked this whole class of bug.
So: two-thirds of the PR is the right route; the third layer is the most complex, least-tested part, solves a problem that hasn't been shown to exist under the old config, and partially manufactures that problem via the storage change. I'd ask the author for the evidence behind the multi-tab race before accepting it.

MauricioUyaguari:
1. Loser tabs keep a stale auth.user — the tokenRef pattern doesn't help them (main concern).
oidc-client-ts has no storage-event listener (I checked the dist — no cross-tab user sync), so when a losing tab runs adoptRenewalFromWinner, the fresh token lands only in ApplicationStore and the cookie. auth.user?.access_token in that tab stays expired forever, and this PR's own call sites all read tokenRef.current = auth.user?.access_token. Result: in a multi-tab session, every approve/deny/refresh action in a non-winner tab still sends the expired token — the exact bug being fixed, just confined to secondary tabs. Options: have the loser call auth.events.load(user) with the persisted user to push it into the auth context, or make call sites fall back to applicationStore.getAccessToken() (which the adopt path does update)

PR #5387, opened 2026-07-29
Support MAPGs with RLE in Studio

Vithesh-Reddy:
Support MAPGs with RLE in Studio

### DIFF
--- a/.changeset/fair-rice-press.md
+++ b/.changeset/fair-rice-press.md
@@ -1,0 +1,5 @@
+---
+'@finos/legend-extension-dsl-data-product': patch
+---
+
+Refactor missing ingest flow in data product viewer to remove duplicate api calls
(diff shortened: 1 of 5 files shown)


packages/legend-graph/src/graph-manager/action/changeDetection/DataProductObserveHelper.ts
```diff
@@ -192,6 +193,17 @@ export const observe_AccessPointGroup = skipObserved(
   },
 );
 
+export const observe_SecureView = skipObserved(
+  (metamodel: SecureView): SecureView => {
+    makeObservable(metamodel, {
+      ingestPath: observable,
+      datasetName: observable,
+      func: observable.ref,
+    });
```
gs-gunjan: do you need to implement the observable.ref here, can it work with func: observable and then you have something like this: observe_RawLambda(metamodel.func);

Vithesh-Reddy: changed it to:
`func: observable,
    });
    observe_RawLambda(metamodel.func);`

PR #5391, opened 2026-07-29
Allow optional date parameters to default to 'No Value' in query execution

FeruzGS:
## Problem

When a query has Date parameters with multiplicity `[0..1]` (optional), the parameter value dialog

defaults to "Today" / "Now" with no way to leave the parameter unset. This forces users to provide

a date value even when the parameter is optional.

## Solution

- Optional Date/StrictDate/DateTime parameters (`[0..1]`) now default to "No Value" instead of a

  concrete date.

- Parameters set to "No Value" are omitted from the execution payload sent to the Engine.

### DIFF
--- a/packages/legend-query-builder/src/components/__tests__/QueryBuilderParametersPanel.test.tsx
+++ b/packages/legend-query-builder/src/components/__tests__/QueryBuilderParametersPanel.test.tsx
@@ -887,12 +887,14 @@ test(
     const paramLambda = TEST_DATA__simpeDateParameters(
       PRIMITIVE_TYPE.STRICTDATE,
     );
-    // Override multiplicity from [1..1] to [0..1]
-    (
-      paramLambda.parameters as {
-        multiplicity: { lowerBound: number; upperBound: number };
-      }[]
-    )[0]!.multiplicity = { lowerBound: 0, upperBound: 1 };
+    const firstParam = guaranteeNonNullable(
+      (
+        paramLambda.parameters as {
+          multiplicity: { lowerBound: number; upperBound: number };
+        }[]
+      )[0],
+    );
+    firstParam.multiplicity = { lowerBound: 0, upperBound: 1 };
 
     await act(async () => {
       queryBuilderState.initializeWithQuery(
--- a/packages/legend-query-builder/src/components/__tests__/QueryBuilderParametersPanel.test.tsx
+++ b/packages/legend-query-builder/src/components/__tests__/QueryBuilderParametersPanel.test.tsx
@@ -870,3 +870,52 @@ test(
     expect(getByText(executeDialog, 'var_1')).toStrictEqual(parameterValue);
   },
 );
+test(
+  integrationTest(
+    'Query builder optional date parameter shows No Value by default',
+  ),
+  async () => {
+    const { renderResult, queryBuilderState } = await TEST__setUpQueryBuilder(
+      TEST_DATA__ComplexRelationalModel,
+      stub_RawLambda(),
+      'model::relational::tests::simpleRelationalMapping',
+      'model::MyRuntime',
+      TEST_DATA__ModelCoverageAnalysisResult_ComplexRelational,
+    );
+
+    // StrictDate with [0..1] multiplicity (optional)
+    const paramLambda = TEST_DATA__simpeDateParameters(
+      PRIMITIVE_TYPE.STRICTDATE,
+    );
+    // Override multiplicity from [1..1] to [0..1]
+    (
+      paramLambda.parameters as {
+        multiplicity: { lowerBound: number; upperBound: number };
+      }[]
+    )[0]!.multiplicity = { lowerBound: 0, upperBound: 1 };
+
+    await act(async () => {
+      queryBuilderState.initializeWithQuery(
+        create_RawLambda(paramLambda.parameters, paramLambda.body),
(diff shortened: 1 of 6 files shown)


MauricioUyaguari (approved):

MauricioUyaguari (changes requested):
.

packages/legend-query-builder/src/components/__tests__/QueryBuilderParametersPanel.test.tsx
```diff
@@ -870,3 +870,54 @@ test(
...
+      PRIMITIVE_TYPE.STRICTDATE,
+    );
+    const firstParam = guaranteeNonNullable(
+      (
+        paramLambda.parameters as {
+          multiplicity: { lowerBound: number; upperBound: number };
+        }[]
+      )[0],
+    );
+    firstParam.multiplicity = { lowerBound: 0, upperBound: 1 };
```
MauricioUyaguari, line 897: can we instead enhance the `TEST_DATA__simpeDateParameters` to take a required boolean and if false, we can return the 0..1 multiplicity in the test data

packages/legend-query-builder/src/components/shared/CustomDatePickerHelper.ts
```diff
@@ -945,21 +946,31 @@ export const buildDatePickerOption = (
...
     }
   } else if (valueSpecification instanceof PrimitiveInstanceValue) {
-    return valueSpecification.genericType.value.rawType.path ===
-      PRIMITIVE_TYPE.LATESTDATE
+    const _path = valueSpecification.genericType.value.rawType.path;
+    const isDateType =
+      _path === PRIMITIVE_TYPE.STRICTDATE ||
+      _path === PRIMITIVE_TYPE.DATE ||
+      _path === PRIMITIVE_TYPE.DATETIME;
+    const isEmpty =
```
MauricioUyaguari, line 954: Inconsistent "emptiness" definitions. buildExecutionParameterValues uses lodash isEmpty(values[0]) (true for null, undefined, and ''), while buildDatePickerOption checks only !length || null || undefined. So a date the user cleared in the absolute-date input (value '') is silently omitted from the payload but does not display as "No Value". Also, lodash isEmpty is a blunt instrument here (it's true for any number and for Date objects). An explicit shared check (v == null || v === '') used in both places would be both clearer and consistent.

packages/legend-query-builder/src/components/shared/CustomDatePickerHelper.ts
```diff
@@ -945,21 +946,31 @@ export const buildDatePickerOption = (
...
         )
-      : new DatePickerOption(
-          (valueSpecification.values[0] ?? '') as string,
-          valueSpecification.values[0] === null
-            ? ''
-            : valueSpecification.genericType.value.rawType.path ===
-                PRIMITIVE_TYPE.DATETIME
+      : isDateType && isEmpty
+        ? new DatePickerOption(
+            CUSTOM_DATE_PICKER_OPTION.NO_VALUE,
```
MauricioUyaguari, line 965: Display regression for required empty dates. buildDatePickerOption has no knowledge of multiplicity — it now returns "No Value" for any empty date PrimitiveInstanceValue. That includes required [1] date values (e.g., a filter condition or a value cleared via reset, where buildDefaultInstanceValue also produces values: [null]). Previously those rendered an empty label, so the trigger button showed the "Select value" prompt; now they'll read "No Value", which implies a deliberate, valid choice for a value that is actually missing and invalid. Consider passing an isOptional hint into buildDatePickerOption (or deriving the label at the call sites that know the multiplicity) so "No Value" only appears for genuinely optional values.

packages/legend-query-builder/src/stores/shared/LambdaParameterState.ts
```diff
@@ -127,6 +131,20 @@ export const buildExecutionParameterValues = (
 ): ParameterValue[] =>
   paramStates
     .filter((ps) => !doesLambdaParameterStateContainFunctionValues(ps))
+    .filter(
+      (ps) =>
+        !(
```
MauricioUyaguari, line 136: can you confirm latest, today() is not filtered out. i think they are part of the let statements so should work right ?

packages/legend-query-builder/src/stores/shared/LambdaParameterState.ts
```diff
@@ -244,6 +262,27 @@ export class LambdaParameterState implements Hashable {
...
+        Multiplicity.ZERO_ONE,
+      );
+    if (isOptionalDate) {
+      const nullDateValue = buildDefaultInstanceValue(
+        this.graph,
+        t,
+        this.observerContext,
+        false,
+      );
+      nullDateValue.multiplicity = this.parameter.multiplicity;
```
MauricioUyaguari, line 281: we shoudlnt be editing multiplicity like that and pointing to the parameter one. 
does this return an empty instance value ?

MauricioUyaguari (changes requested):
V1 protocol branch not covered. buildDatePickerOption has a parallel branch for V1 value specs (V1_CLatestDate etc., used by the V1/DataCube editors) that was not given equivalent empty-value handling, so behavior now differs between the two flows. If that's intentionally out of scope, fine — but worth stating in the PR description.

PR #5395, opened 2026-07-29
remove inventory from marketplace tabs

MauricioUyaguari:

- [ ] Test(s) added


bojja-gs:
Why do we have to remove entirely?? We can simply hide in header, then reuse if we want to add any tab or rename later ?

bojja-gs (approved):

PR #5400, opened 2026-08-03
Support enhanced Relation Function property mappings in Studio

Vithesh-Reddy:
Support lambda-based relation function class mapping — ~src (sourceLambda) on the class mapping and expression-based [valueFn] on property mappings, as alternatives to the named [~func] pointer and plain column references.


packages/legend-graph/src/graph/metamodel/pure/packageableElements/mapping/relationFunction/RelationFunctionPropertyMapping.ts
```diff
@@ -56,7 +61,7 @@ export class RelationFunctionPropertyMapping
     property: PropertyReference,
     source: SetImplementationReference,
     target: SetImplementationReference | undefined,
-    column: RelationColumn,
+    column: RelationColumn | undefined,
```
gs-gunjan, line 64: does valueFn need to be part of constructor?

Vithesh-Reddy: not needed, it's set after construction

gs-gunjan (approved):

PR #5401, opened 2026-08-03
feat: add support for viewing and loading query version history in Legend Query

gs-gunjan:
## Summary
Adds the ability to view a saved query's revision history and load any past revision. History comes from the /query/{queryId}/history endpoint (returning V1_Query[]); each entry is identified by its version field, and a specific revision is loaded via a new revisionId route parameter.

- [X] Test(s) added

### DIFF
--- a/packages/legend-application-query/src/stores/QueryEditorStore.ts
+++ b/packages/legend-application-query/src/stores/QueryEditorStore.ts
@@ -1912,6 +1912,9 @@ export class ExistingQueryEditorStore extends QueryEditorStore {
   showQueryVersionHistory(): void {
     if (this.query) {
       this.queryLoaderState.setQueryLoaderDialogOpen(true);
+      // opened directly into history (not drilled in from the query list), so
+      // there is no list to go back to
+      this.queryLoaderState.setHistoryViewerStandalone(true);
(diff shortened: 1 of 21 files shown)


packages/legend-application-query/src/stores/QueryEditorStore.ts
```diff
@@ -1837,6 +1867,9 @@ const resolveExecutionContext = (
 
 export class ExistingQueryEditorStore extends QueryEditorStore {
   private queryId: string;
+  // A specific revision from the query's version history to load, keyed by the
+  // revision's `version` identifier. When undefined, the latest is loaded.
+  readonly revisionId: string | undefined;
```
MauricioUyaguari, line 1872: can we ensure we disable things that shouldnt work with this view. so i assume we can't save if user makes changes ? we should disable the save button with revisionId is present right ?

MauricioUyaguari (approved):

packages/legend-application-query/src/stores/QueryEditorStore.ts
gs-gunjan: They can save -> and that becomes the latest query

